PR #8, opened 2018-07-05
Add test cases for ARIA in HTML features

Bugvi-BM:
Added HTML feature tests. These can be considered blackbox tests as they are purely based on the documentation and not on the source code.


packages/alfa-aria/src/features/select.ts
```diff
@@ -21,7 +21,7 @@ function role(select: Element, context: Node): Role {
   const size = getAttribute(select, "size");
   if (
     !hasAttribute(select, "multiple") &&
-    (size === null || parseInt(size) > 1)
+    (size === null || !(parseInt(size) > 1))
```
kasperisager: I'd do `parseInt(size) <= 1` instead.

PR #12, opened 2018-07-09
Closes #9: Implement HTML script data tokenisation steps

### LINKED ISSUE #9: Implement HTML script data tokenisation steps
Implement sections [8.2.4.15](https://www.w3.org/TR/html/syntax.html#script-data-less-than-sign-state) through [8.2.4.31](https://www.w3.org/TR/html/syntax.html#script-data-double-escape-end-state) of the HTML spec in the [HTML alphabet](https://github.com/Siteimprove/alfa/blob/master/packages/alfa-html/src/alphabet.ts).

NiclasHedam:
This implements script tokenisation as requested in issue #9.


packages/alfa-html/src/alphabet.ts
```diff
@@ -126,6 +130,27 @@ const data: Pattern = (stream, emit, state) => {
...
+ * @see https://www.w3.org/TR/html/syntax.html#script-data-state
+ */
+const scriptData: Pattern = (stream, emit, state) => {
+  const char = stream.peek(0);
+  stream.advance(1);
+  switch (char) {
+    case Char.LessThanSign:
+      return scriptDataLessThanSign;
+    case Char.Null:
+      emit({ type: TokenType.Character, data: fromCharCode(65533) });
```
kasperisager: JavaScript has Unicode escape sequences:

```ts
emit({ type: TokenType.Character, data: "\ufffd" });
```

packages/alfa-html/src/alphabet.ts
```diff
@@ -126,6 +130,27 @@ const data: Pattern = (stream, emit, state) => {
...
+      break;
+    case null:
+      return Command.End;
+    default:
+      emit({ type: TokenType.Character, data: fromCharCode(char) });
+  }
+};
```
kasperisager: Make sure to leave a little room between expressions and statements; the above looks very cramped, making it harder to get an overview of what's going on. I'd do:

```ts
const scriptData: Pattern = (stream, emit, state) => {
  const char = stream.peek(0);
  stream.advance(1);

  switch (char) {
    case Char.LessThanSign:
      return scriptDataLessThanSign;

    case Char.Null:
      emit({ type: TokenType.Character, data: fromCharCode(65533) });
      break;

    case null:
      return Command.End;

    default:
      emit({ type: TokenType.Character, data: fromCharCode(char) });
  }
};
```

Whitespace is your friend!

packages/alfa-html/src/alphabet.ts
```diff
@@ -126,6 +130,27 @@ const data: Pattern = (stream, emit, state) => {
 };
 
 /**
+ * 8.2.4.4
+ * @see https://www.w3.org/TR/html/syntax.html#script-data-state
+ */
+const scriptData: Pattern = (stream, emit, state) => {
+  const char = stream.peek(0);
+  stream.advance(1);
```
kasperisager: Do:

```ts
const char = stream.next();
```

This will peek and then advance.

packages/alfa-html/src/alphabet.ts
kasperisager: This applies to everything below as well.

packages/alfa-html/src/alphabet.ts
```diff
@@ -237,6 +264,522 @@ const tagName: Pattern = (stream, emit, state) => {
...
+          state.tag.name += fromCharCode(char);
+        } else {
+          state.tag = {
+            type: TokenType.EndTag,
+            name: fromCharCode(char)
+          };
+        }
+        state.temporaryBuffer += fromCharCode(char);
+        break;
+      }
```
kasperisager: The above can be simplified quite a bit:

```ts
if (isAlpha(char)) {
  stream.advance(1);

  let str = fromCharCode(char);

  state.tag!.name += str.toLowerCase();
  state.temporaryBuffer += str;
  break;
}
```

`state.tag` will always be different from `null` when this pattern is executed, which is why I make use of a non-null assertion (`!`). The casing of the character also doesn't really matter; if it's alphabetic then we simply lowercase it.

packages/alfa-html/src/alphabet.ts
```diff
@@ -237,6 +264,522 @@ const tagName: Pattern = (stream, emit, state) => {
...
+ */
+const scriptDataEndTagName: Pattern = (stream, emit, state) => {
+  function isAppropriateEndTagToken(): boolean {
+    return state.tag !== null && state.tag.name === "script";
+  }
+
+  function anythingElse() {
+    emit({ type: TokenType.Character, data: "<" });
+    emit({ type: TokenType.Character, data: "/" });
+    for (let i = 0; i < state.temporaryBuffer.length; i++) {
```
kasperisager: Avoid accessing the length more than once:

```ts
for (let i = 0, n = state.temporaryBuffer.length; i < n; i++) { ...
```

packages/alfa-html/src/alphabet.ts
```diff
@@ -237,6 +264,522 @@ const tagName: Pattern = (stream, emit, state) => {
...
+            type: TokenType.EndTag,
+            name: fromCharCode(char)
+          };
+        }
+        state.temporaryBuffer += fromCharCode(char);
+        break;
+      }
+      return anythingElse;
+  }
+};
```
kasperisager: I'd restructure the above to this:

```ts
const scriptDataEndTagName: Pattern = (stream, emit, state) => {
  const char = stream.peek(0);

  switch (char) {
    case Char.CharacterTabulation:
    case Char.LineFeed:
    case Char.FormFeed:
    case Char.Space:
      stream.advance(1);
      if (isAppropriateEndTagToken(state.start!, state.end!)) {
        return beforeAttributeName;
      }
      break;
    ...
    default:
      if (char !== null && isAlpha(char)) {
        ...
        return;
      }
  }

  emit({ type: TokenType.Character, data: "<" });
  emit({ type: TokenType.Character, data: "/" });

  for (let i = 0; i < state.temporaryBuffer.length; i++) {
    emit({ type: TokenType.Character, data: state.temporaryBuffer[i] });
  }

  return scriptData;
};
```

This way, the flow through the function is kept linear.

packages/alfa-html/src/alphabet.ts
```diff
@@ -237,6 +264,522 @@ const tagName: Pattern = (stream, emit, state) => {
...
+
+/**
+ * 8.2.4.17
+ * @see https://www.w3.org/TR/html/syntax.html#tokenizer-script-data-end-tag-name-state
+ */
+const scriptDataEndTagName: Pattern = (stream, emit, state) => {
+  function isAppropriateEndTagToken(): boolean {
+    return state.tag !== null && state.tag.name === "script";
+  }
```
kasperisager: Move this function to the top scope and parameterise it:

```ts
/**
 * @see https://www.w3.org/TR/html/syntax.html#appropriate-end-tag-token
 */
function isAppropriateEndTagToken(start: StartTag, end: EndTag): boolean {
  return start.name === end.name;
}
```

packages/alfa-html/src/alphabet.ts
```diff
@@ -237,6 +264,522 @@ const tagName: Pattern = (stream, emit, state) => {
...
+ * 8.2.4.23
+ * @see https://www.w3.org/TR/html/syntax.html#tokenizer-script-data-escaped-less-than-sign-state
+ */
+const scriptDataEscapedLessThanSign: Pattern = (stream, emit, state) => {
+  const char = stream.peek(0);
+  if (char === Char.Solidus) {
+    stream.advance(1);
+    state.temporaryBuffer = "";
+    return scriptDataEscapedEndTagOpen;
+  } else if (char !== null && isAlpha(char)) {
```
kasperisager: No need for an `else` after a `return`:

```ts
if (char === Char.Solidus) {
  ...
  return scriptDataEscapedEndTagOpen;
}

if (char !== null && isAlpha(char)) {
  ...
  return scriptDataDoubleEscapeStart;
}

emit({ type: TokenType.Character, data: "<" });

return scriptDataEscaped;
```

packages/alfa-html/src/alphabet.ts
```diff
@@ -237,6 +264,522 @@ const tagName: Pattern = (stream, emit, state) => {
...
+            type: TokenType.EndTag,
+            name: fromCharCode(char)
+          };
+        }
+        state.temporaryBuffer += fromCharCode(char);
+        break;
+      }
+      return anythingElse;
+  }
+};
```
kasperisager: Same comments for this one as for the similar one above.

packages/alfa-html/src/alphabet.ts
```diff
@@ -237,6 +264,522 @@ const tagName: Pattern = (stream, emit, state) => {
...
+        state.temporaryBuffer += fromCharCode(char).toLowerCase();
+        emit({ type: TokenType.Character, data: fromCharCode(char) });
+        break;
+      }
+      if (isLowerCase(char)) {
+        stream.advance(1);
+        state.temporaryBuffer += fromCharCode(char);
+        emit({ type: TokenType.Character, data: fromCharCode(char) });
+        break;
+      }
```
kasperisager: Same as previously; just check for an alphabetic `char` and lowercase indiscriminately.

packages/alfa-html/src/alphabet.ts
```diff
@@ -1573,3 +2160,15 @@ function startsWith(
 
   return true;
 }
+
+function findAppropriateState(state: State) {
```
kasperisager: I'd better qualify the name of this function as it's very specific to script patterns.

packages/alfa-html/src/alphabet.ts
NiclasHedam: Duly noted. I've changed the let statement to const, as the value is never reassigned.

NiclasHedam:
I've implemented all requested changes but findAppropriateState (see https://github.com/Siteimprove/alfa/pull/12#discussion_r201241601). All tests passes and everything compiles.

packages/alfa-html/src/alphabet.ts
```diff
@@ -237,6 +279,522 @@ const tagName: Pattern = (stream, emit, state) => {
...
+        emit({ type: TokenType.Character, data: fromCharCode(char) });
+        break;
+      }
+
+      if (isLowerCase(char)) {
+        stream.advance(1);
+        state.temporaryBuffer += fromCharCode(char);
+        emit({ type: TokenType.Character, data: fromCharCode(char) });
+        break;
+      }
```
kasperisager: Same as previously; just check for an alphabetic char and lowercase indiscriminately.

packages/alfa-html/src/alphabet.ts
```diff
@@ -237,6 +277,500 @@ const tagName: Pattern = (stream, emit, state) => {
...
+
+      emit({ type: TokenType.Character, data: fromCharCode(char) });
+      return scriptDataDoubleEscaped;
+
+    default:
+      if (char !== null && isAlpha(char)) {
+        stream.advance(1);
+        state.temporaryBuffer += fromCharCode(char).toLowerCase();
+        emit({ type: TokenType.Character, data: fromCharCode(char) });
```
kasperisager: Store `fromCharCode(char)` in variable to avoid computing it twice.

packages/alfa-html/src/alphabet.ts
```diff
@@ -237,6 +277,500 @@ const tagName: Pattern = (stream, emit, state) => {
...
+      if (state.temporaryBuffer === "script") {
+        return scriptDataDoubleEscaped;
+      }
+      return scriptDataEscaped;
+
+    default:
+      if (char !== null && isAlpha(char)) {
+        stream.advance(1);
+        state.temporaryBuffer += fromCharCode(char).toLowerCase();
+        emit({ type: TokenType.Character, data: fromCharCode(char) });
```
kasperisager: Store fromCharCode(char) in variable to avoid computing it twice.

PR #26, opened 2018-08-06
Closes #13: Figure out way to present coverage data

### LINKED ISSUE #13: Figure out way to present coverage data
The `@siteimprove/alfa-test` package currently collects coverage data using V8's builtin coverage profiler. This proved to be the most performant and unobtrusive way of collecting coverage data compared to e.g. [nyc](https://github.com/istanbuljs/nyc). What we need now is a way of presenting this coverage data in a way that ideally integrates well with our existing workflow. Let's start out small and see if we can find a way of visualising what is _missing_ similar to how we only report on failing tests. Creative solutions are welcome!

NiclasHedam:
The build system has been extended to now calculate a coverage percentage and print it in the terminal whenever a file is compiled. When you run `npm start` or `yarn start`, the build system will also print out a coverage suggestion. This suggestion is currently based on three heuristics; the amount of logical operators, the amount of arithmetic operators and lastly, the byte length.

After playing around a bit, I think the system prints some good and educated guesses, and I think it is therefore ready for being on master.

There are still two things to be improved (either later or during the code review process):
 - The coverage suggestion tool is slow on large files (fx. `alphabet.ts` in `alfa-html`).
 - The tool could probably be more verbose in case of unexpected situations (fx. if a test file has no matching source file, and thereby breaking convention).

If there is any suggestions on how to archieve the above, let me know.


build/helpers/coverage-heuristics/arithmetic.js
```diff
@@ -0,0 +1,152 @@
...
+  let total = 0;
+  if (script.sources.length > 1) {
+    // typescript
+    for (let i = 1, n = script.sources.length; i < n; i++) {
+      total += visit(
+        TypeScript.createSourceFile(
+          "anon.ts",
+          script.sources[i].content,
+          TypeScript.ScriptTarget.ES2015
+        )
```
kasperisager: The above is repeated quite often and should be abstracted away.

build/helpers/coverage.js
```diff
@@ -41,7 +73,34 @@ process.on("exit", () => {
...
+        if (a.points < b.points) {
+          return 1;
+        }
+
+        if (a.points > b.points) {
+          return -1;
+        }
+
+        return 0;
+      });
```
kasperisager: `Array.sort()` modifies the array in-place so there's no need to assign to another variable. The comparator can also be simplified:

```ts
script.coverage.sort((a, b) => a.points - b.points);
```

build/helpers/coverage.js
```diff
@@ -212,11 +273,19 @@ function parseFunctionCoverage(script, map, range, name) {
...
+    points: 0,
     name
   };
+
+  if (block.count !== 0) {
+    return block;
+  }
+
+  block.points = calculateBlockCoverage(script, block);
+  return block;
```
kasperisager: Do:

```ts
return {
  range: parsed,
  count: range.count:
  points: block.count === 0 ? 0 : calculateBlockCoverage(script, block),
  name
};
```

build/helpers/coverage.js
```diff
@@ -1,6 +1,20 @@
...
+  byteLengthBlockCoverage
+} from "./coverage-heuristics/byte";
+import {
+  logicalTotalCoverage,
+  logicalBlockCoverage
+} from "./coverage-heuristics/logical";
+import {
+  arithmeticTotalCoverage,
+  arithmeticBlockCoverage
+} from "./coverage-heuristics/arithmetic";
```
kasperisager: Rather than make each heuristic two separate function exports, just export a single object that describes the heuristic:

```ts
export const Arithmetic = {
  total(...)  {
    // ...
  },

  block(...) {
    // ...
  }
}
```

To use these in `coverage.js`, do:

```ts
import { Arithmetic } from "...";

const heuristics = [
  {
    ...Arithmetic,
    weight: 1 / 3
  },
  // ...
];

File: build/helpers/coverage.js [hunk 5]
```diff
@@ -387,3 +464,90 @@ function isBlockBorder(input) {
...
+
+  const above = `${source.lines[start.line - 1].value}`;
+  const below = `${source.lines[end.line + 1].value}`;
+
+  const before = source.lines[start.line].value.slice(0, start.column);
+  const after = source.lines[end.line].value.slice(end.column);
+
+  let output = `${chalk.bold("Suggested Block to Cover")}`;
+
+  output += `\n${chalk.dim(`${filePath}:${start.line + 1}-${end.line + 1}`)}`;
```
kasperisager: You can leave out the end line number and just display the start.

build/helpers/coverage.js
```diff
@@ -387,3 +464,90 @@ function isBlockBorder(input) {
...
+  let file = "";
+  if (script.sources.length === 1) {
+    file += " in " + script.sources[0].path;
+  } else if (script.sources.length === 2) {
+    file += " in " + script.sources[1].path;
+  }
+  const total = calculateTotalCoverage(script);
+  if (total < 90) {
+    notify.warn(`Low coverage (${total.toFixed(2)}%)` + file);
+  }
```
kasperisager: To ensure consistency with the rest of the output:

```ts
let source;

switch (script.sources.length) {
  case 1:
    source = script.sources[0];
    break;
  case 2:
    source = script.sources[1];
    break;
  default:
    return;
}

const total = calculateTotalCoverage(script);

const filePath = path.relative(
  process.cwd(),
  path.resolve(script.base, source.path)
);

if (total < 90) {
  notify.warn(`${chalk.dim(filePath)} Low coverage (${total.toFixed(2)}%)`);
}
```

build/helpers/coverage.js
NiclasHedam: I actually think it makes better sense to show the end-line. We don't want them to cover line A, we want them to cover line A to B.

kasperisager: The line number is just an indication of where to start looking for the code snippet, similar to how linting and compile errors are shown.

build/helpers/coverage-heuristics/arithmetic.js
```diff
@@ -0,0 +1,156 @@
...
+
+        uncovered += visit(
+          TypeScript.createSourceFile(
+            "anon.ts",
+            file.content.substring(
+              block.range.start.offset,
+              block.range.end.offset
+            ),
+            TypeScript.ScriptTarget.ES2015
+          )
```
kasperisager: Should make use of `createSource()`.

build/helpers/coverage-heuristics/byte.js
```diff
@@ -0,0 +1,58 @@
+import { BlockCoverage, FunctionCoverage, Script } from "../coverage";
+
+/**
+ * @param {Script} script
+ * @return {Number}
```
kasperisager: Should be `@return {number}`.

build/helpers/coverage-heuristics/byte.js
```diff
@@ -0,0 +1,58 @@
...
+  return total;
+}
+
+export const Byte = {
+  /**
+   * @param {Script} script
+   */
+  total(script) {
+    let uncovered = 0;
+    const total = /** @type {Number} */ (totalLength(script));
```
kasperisager: Is there a reason why a type assertion was added here?

build/helpers/coverage-heuristics/arithmetic.js
```diff
@@ -0,0 +1,156 @@
...
+
+    let uncovered = visit(
+      TypeScript.createSourceFile(
+        "anon.ts",
+        file.content.substring(
+          block.range.start.offset,
+          block.range.end.offset
+        ),
+        TypeScript.ScriptTarget.ES2015
+      )
```
kasperisager: Should make use of `createSource()`.

build/helpers/coverage-heuristics/logical.js
```diff
@@ -0,0 +1,132 @@
...
+
+        uncovered += visit(
+          TypeScript.createSourceFile(
+            "anon.ts",
+            file.content.substring(
+              block.range.start.offset,
+              block.range.end.offset
+            ),
+            TypeScript.ScriptTarget.ES2015
+          )
```
kasperisager: Should make use of `createSource()`.

build/helpers/coverage-heuristics/logical.js
```diff
@@ -0,0 +1,132 @@
...
+
+    let uncovered = visit(
+      TypeScript.createSourceFile(
+        "anon.ts",
+        file.content.substring(
+          block.range.start.offset,
+          block.range.end.offset
+        ),
+        TypeScript.ScriptTarget.ES2015
+      )
```
kasperisager: Should make use of `createSource()`.

build/helpers/coverage.js
```diff
@@ -41,7 +61,34 @@ process.on("exit", () => {
...
+      }
+      const script = parseScript(scriptCoverage);
+
+      if (script === null) {
+        continue;
+      }
+
+      let computedPoints = computePoints(script);
+      let sortedBlocks = script.coverage
+        .filter(block => block.count == 0)
```
kasperisager: Should be `===`; I need to find a way of enabling the linter rules for JavaScript files as well.

build/helpers/coverage.js
```diff
@@ -387,3 +434,118 @@ function isBlockBorder(input) {
...
+
+/**
+ * @param {Script} script
+ * @returns {Map<BlockCoverage | FunctionCoverage, number>}
+ */
+function computePoints(script) {
+  const total = calculateTotalCoverage(script);
+  const map = new Map();
+
+  script.coverage.forEach(block => {
```
kasperisager: A `for ... of` would be good here.

NiclasHedam:
I will look into it.

NiclasHedam:
There we go. Apparently, the results of the bye coverage heuristic was reversed, meaning that a file with high byte coverage was reported as low coverage. This is the reason why many results were near 66.67%.

PR #28, opened 2018-08-06
Closes #24: Implement `getTabSequence()`

### LINKED ISSUE #24: Implement `getTabSequence()`
This function should compute a best-guess of the sequential focus navigation order (tab sequence) of a given context node. The sequence should be computed based solely on [`getTabIndex()`](https://github.com/Siteimprove/alfa/blob/master/packages/alfa-dom/src/get-tab-index.ts), hence "best-guess"; we cannot currently account for any JavaScript trickery used for hijacking the focus order.

Example:

```jsx
getTabSequence(<div><a href="#">...</a>...</div>);
// => [<a href="#">...</a>, ...]
```

See: https://www.w3.org/TR/html/editing.html#sequential-focus-navigation

NiclasHedam:
The `getTabSequence()` casts all elements to an TabIndexedElement, which contains its index weight. This is done, due to the fact that the order of which the elements are put into the array could return a bad order.

### DIFF
--- a/packages/alfa-dom/src/get-tab-sequence.ts
+++ b/packages/alfa-dom/src/get-tab-sequence.ts
@@ -1,31 +1,40 @@
 import { getTabIndex } from "./get-tab-index";
 import { Element, Node, NodeType } from "./types";
 
+export interface TabIndexedElement extends Element {
+  index: number;
+}
+
 /**
  * @see https://www.w3.org/TR/html/editing.html#the-tabindex-attribute
  */
 export function getTabSequence(
   element: Element,
-  tabSequence: Array<Element> = []
-): Array<Element> {
+  tabSequence: Array<TabIndexedElement> = []
+): Array<TabIndexedElement> {
   (<Array<Node>>element.childNodes).forEach(element => {
     if (element.nodeType === NodeType.Element) {
-      getTabSequence(<Element>element, tabSequence);
+      getTabSequence(<TabIndexedElement>element, tabSequence);
     }
   });
 
   const index = getTabIndex(element);
+
   if (index === null || index === -1) {
     return tabSequence;
   }
 
+  const weightedElement = <TabIndexedElement>element;
+
   if (index > 0) {
-    tabSequence.splice(index - 1, 0, element);
+    weightedElement.index = index - 1;
   }
 
   if (index === 0) {
-    tabSequence.splice(999999, 0, element);
+    weightedElement.index = Number.MAX_SAFE_INTEGER;
   }
 
-  return tabSequence;
+  tabSequence.push(weightedElement);
+
+  return tabSequence.sort((a, b) => a.index - b.index);
 }
--- a/packages/alfa-dom/src/get-tab-sequence.ts
+++ b/packages/alfa-dom/src/get-tab-sequence.ts
@@ -1,5 +1,5 @@
 import { getTabIndex } from "./get-tab-index";
-import { Element, Node, NodeType } from "./types";
+import { Element, NodeType } from "./types";
 
 export interface TabIndexedElement extends Element {
   index: number;
@@ -12,11 +12,13 @@ export function getTabSequence(
   element: Element,
   tabSequence: Array<TabIndexedElement> = []
 ): Array<TabIndexedElement> {
-  (<Array<Node>>element.childNodes).forEach(element => {
-    if (element.nodeType === NodeType.Element) {
-      getTabSequence(<TabIndexedElement>element, tabSequence);
+  element.childNodes.length;
+  for (let i = 0, n = element.childNodes.length; i < n; i++) {
+    const child = element.childNodes[i];
+    if (child.nodeType === NodeType.Element) {
+      getTabSequence(<TabIndexedElement>child, tabSequence);
     }
-  });
+  }
 
   const index = getTabIndex(element);
 
--- a/packages/alfa-dom/src/get-tab-sequence.ts
+++ b/packages/alfa-dom/src/get-tab-sequence.ts
@@ -1,0 +1,31 @@
+import { getTabIndex } from "./get-tab-index";
+import { Element, Node, NodeType } from "./types";
+
+/**
+ * @see https://www.w3.org/TR/html/editing.html#the-tabindex-attribute
+ */
+export function getTabSequence(
+  element: Element,
+  tabSequence: Array<Element> = []
+): Array<Element> {
+  (<Array<Node>>element.childNodes).forEach(element => {
+    if (element.nodeType === NodeType.Element) {
+      getTabSequence(<Element>element, tabSequence);
+    }
+  });
+
+  const index = getTabIndex(element);
+  if (index === null || index === -1) {
+    return tabSequence;
+  }
+
+  if (index > 0) {
+    tabSequence.splice(index - 1, 0, element);
+  }
+
+  if (index === 0) {
+    tabSequence.splice(999999, 0, element);
+  }
+
+  return tabSequence;
+}
--- a/packages/alfa-dom/src/get-tab-sequence.ts
+++ b/packages/alfa-dom/src/get-tab-sequence.ts
@@ -30,7 +30,7 @@ function indexWithin(array: Array<Element>, element: Element) {
 export function getTabSequence(element: Element): Array<Element> {
   const result: Array<Element> = [];
 
-  traverseNode(element, {
+  traverseNode(element, element, {
     enter(node, parent) {
       if (!isElement(node)) {
         return;
--- a/packages/alfa-dom/src/get-tab-sequence.ts
+++ b/packages/alfa-dom/src/get-tab-sequence.ts
@@ -1,42 +1,58 @@
 import { getTabIndex } from "./get-tab-index";
+import { traverseNode } from "./traverse-node";
 import { Element, NodeType } from "./types";
 
-export interface TabIndexedElement extends Element {
-  index: number;
+function binaryInsert(
+  arr: Array<Element>,
+  element: Element,
+  index: number,
+  i: number,
+  depth = 0
+) {
+  if (arr.length === 0) {
+    arr.push(element);
+  }
+
+  const competitorIndex = getTabIndex(arr[i]);
+
+  if (competitorIndex === null) {
+    return;
+  }
+
+  if (competitorIndex < index) {
+    binaryInsert(arr, element, index, i + i / 2);
+  }
+
+  if (competitorIndex > index) {
+    binaryInsert(arr, element, index, i - i / 2);
+  }
+
+  arr.splice(i, 0, element);
 }
 
 /**
  * @see https://www.w3.org/TR/html/editing.html#the-tabindex-attribute
  */
-export function getTabSequence(
-  element: Element,
-  tabSequence: Array<TabIndexedElement> = []
-): Array<TabIndexedElement> {
-  element.childNodes.length;
-  for (let i = 0, n = element.childNodes.length; i < n; i++) {
-    const child = element.childNodes[i];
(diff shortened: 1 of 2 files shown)


packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  if (index === 0) {
+    weightedElement.index = Number.MAX_SAFE_INTEGER;
+  }
+
+  tabSequence.push(weightedElement);
+
+  return tabSequence.sort((a, b) => a.index - b.index);
+}
```
kasperisager: While this implementation is correct–kudos for that!–it is not optimal. The time complexity is currently in the order of `O(n (n log n))` due to sorting the entire tab sequence on each recursive invocation of `getTabSequence()`. This can be brought down to `O(n log n)` in the following way:

1. Initialise an empty array `result` that will contain the tab sequence.
2. Use `traverseNode()` to traverse the given node and its children. For every node:
    1. Compute its tab index using `getTabIndex()`.
    2. If the tab index is different from `null` and greater than or equal to `0`, add the node to `result` using binary search.
3. Return `result`.

NiclasHedam:
I have completely rewritten the function to your requirements.

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,46 @@
+import { getTabIndex } from "./get-tab-index";
+import { traverseNode } from "./traverse-node";
+import { Element, NodeType } from "./types";
+
+function binaryIndexSearch(array: Array<Element>, index: number) {
```
kasperisager: I'd call this function `indexWithin()` (in line with `indexOf()`) and make the second parameter of type `Element`. That would make it read as "Given this array of elements, get the index that this element would have within that array".

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,46 @@
...
+  let low = 0;
+  let high = array.length;
```
kasperisager: I tend to call these `lower` and `upper` for the simple reason that they align 😄

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  if (index === 0) {
+    return high;
+  }
+
+  while (low < high) {
+    const mid = (low + high) >>> 1;
```
kasperisager: Another JavaScript gotcha: All bitwise operations operate on 32 bit numbers and will simply cut off bits when numbers exceed 32 bits. The safe, i.e. non-overflowing, way of computing the middle between two array indices is therefore:

```ts
const middle = (lower + (upper - lower) / 2) | 0;
```

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,46 @@
...
+    const other = <number>getTabIndex(array[mid]);
```
kasperisager: To avoid clashes with TSX syntax, type assertions should be made using the `as` keyword:

```ts
getTabIndex(array[middle]) as number
```

I'd also add a comment stating why the assertion is safe.

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,46 @@
...
+    if (other <= index && other !== 0) {
+      low = mid + 1;
+    } else {
+      high = mid;
+    }
+  }
+  return low;
+}
```
kasperisager: Thumbs up for a succinct and iterative binary search! 👍

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,46 @@
...
+ * @see https://www.w3.org/TR/html/editing.html#the-tabindex-attribute
+ */
+export function getTabSequence(element: Element): Array<Element> {
+  const result = <Array<Element>>[];
+
+  traverseNode(element, {
+    enter(node, parent) {
+      if (node.nodeType !== NodeType.Element) {
+        return;
+      }
```
kasperisager: You should use the `isElement()` guard.

packages/alfa-dom/src/get-tab-sequence.ts
NiclasHedam: Gotta love JavaScript gotchas.

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+  return lower;
+}
+
+/**
+ * @see https://www.w3.org/TR/html/editing.html#the-tabindex-attribute
+ */
+export function getTabSequence(element: Element): Array<Element> {
+  const result = <Array<Element>>[];
```
kasperisager: Should be:

```ts
const result: Array<Element> = [];
```

kasperisager:
Once merged with `master`, the call to `traverseNode()` will need to be passed a context which should be a second parameter to `getTabSequence()`.

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+  return lower;
+}
+
+/**
+ * @see https://www.w3.org/TR/html/editing.html#the-tabindex-attribute
+ */
+export function getTabSequence(
+  element: Element,
+  context: Element
```
kasperisager: `context` is always a `Node`.

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,51 @@
...
+      lower = middle + 1;
+    } else {
+      upper = middle;
+    }
+  }
+  return lower;
+}
+
+/**
+ * @see https://www.w3.org/TR/html/editing.html#the-tabindex-attribute
```
kasperisager: Should be: https://www.w3.org/TR/html/editing.html#sequential-focus-navigation

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,51 @@
...
+ */
+export function getTabSequence(
+  element: Element,
+  context: Element
+): Array<Element> {
+  const result: Array<Element> = [];
+
+  traverseNode(element, context, {
+    enter(node, parent) {
```
kasperisager: `parent` is unused.

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -1,7 +1,7 @@
 import { getTabIndex } from "./get-tab-index";
 import { isElement } from "./guards";
 import { traverseNode } from "./traverse-node";
-import { Element } from "./types";
+import { Element, Node } from "./types";
 
 function indexWithin(array: Array<Element>, element: Element) {
```
kasperisager: Should be moved to the bottom; public APIs first, then private.

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,51 @@
...
+  return lower;
+}
+
+/**
+ * @see https://www.w3.org/TR/html/editing.html#sequential-focus-navigation
+ */
+export function getTabSequence(
+  element: Element,
+  context: Node
+): Array<Element> {
```
kasperisager: Let's make the return type `Readonly<Array<Element>>`. This makes it possible for us to implement memoization at a later stage without having to copy the cached array. If the caller needs to modify the returned array, they're free to make a copy themselves.

packages/alfa-dom/test/get-tab-sequence.spec.tsx
```diff
@@ -0,0 +1,109 @@
+import { jsx } from "@siteimprove/alfa-jsx";
+import { test } from "@siteimprove/alfa-test";
+import { getTabSequence } from "../src/get-tab-sequence";
+
+test("Computes the tab sequence of three non-prioritized elements", t => {
+  const a1 = <a href="#">Foo</a>;
+  const a2 = <a href="#">Bar</a>;
+  const input = <input type="text" value="Hello" />;
```
kasperisager: I'd recommend the following fixtures as they cover both interactive and static elements:

```tsx
const div = <div />;
const a = <a href="#" />;
const input = <input type="text" />;
```

It's also a good idea to include only the bare minimum of attributes when creating fixture elements. Text (e.g. `Foo`) is sometimes useful to distinguish otherwise identical elements when debugging but can generally be left out.

packages/alfa-dom/test/get-tab-sequence.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+  t.deepEqual(
+    getTabSequence(
+      <div id="root">
+        {a1}
+        {a2}
+        {input}
+      </div>,
+      <div id="context" />
```
kasperisager: The context is supposed to be the top-most node of a given document:

```jsx
const context = (
  <div>
    {div}
    {a}
    {input}
  </div>
);

getTabSequence(context, context);
```

In test cases, the context is often the node being operated on as is the case above.

packages/alfa-dom/test/get-tab-sequence.spec.tsx
```diff
@@ -0,0 +1,109 @@
+import { jsx } from "@siteimprove/alfa-jsx";
+import { test } from "@siteimprove/alfa-test";
+import { getTabSequence } from "../src/get-tab-sequence";
+
+test("Computes the tab sequence of three non-prioritized elements", t => {
```
kasperisager: I try to make the first test case the "base case" of the function under test which also serves as an example of how the function works, e.g. "Gets the tab sequence of a node and its children".

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }
+  return lower;
+}
+
+/**
+ * @see https://www.w3.org/TR/html/editing.html#sequential-focus-navigation
+ */
+export function getTabSequence(
+  element: Element,
```
kasperisager: The first argument should be a `Node` which will allow any type of node to be used.

packages/alfa-dom/test/get-tab-sequence.spec.tsx
NiclasHedam: I think this already is the base case.

packages/alfa-dom/src/get-tab-sequence.ts
```diff
@@ -7,7 +7,7 @@ import { Element, Node } from "./types";
  * @see https://www.w3.org/TR/html/editing.html#sequential-focus-navigation
  */
 export function getTabSequence(
-  element: Element,
+  element: Node,
```
kasperisager: Should be: `node: Node`

NiclasHedam: Good point.

packages/alfa-dom/test/get-tab-sequence.spec.tsx
kasperisager: Aye, the test itself is fine; it's just a matter of the title. The vagueness is fine as the test itself _is_ vague; it just tests that given an element with some seemingly random children, the correct tab sequence is returned. That is perfectly reflected in a rather vague title such as "Gets the tab sequence of a node and its children".

As for the remaining tests, the titles will also need to be more concrete. Off the top of my head, these cases (titles) should be covered:

- Sorts element based on their tab index
- Leaves out elements with negative tab indices
- Positions elements with a tab index of zero after elements with a non-zero tab index

PR #29, opened 2018-08-07
Closes #23: Implement equivalent of `CSS.escape()`

### LINKED ISSUE #23: Implement equivalent of `CSS.escape()`
Just what it says on the tin. See: https://www.w3.org/TR/cssom/#dom-css-escape

NiclasHedam:


NiclasHedam:
Like this?

packages/alfa-css/test/escape.spec.ts
```diff
@@ -0,0 +1,37 @@
+import { Char } from "@siteimprove/alfa-lang";
+import { test } from "@siteimprove/alfa-test";
+import { escape } from "../src/escape";
+
+const { fromCharCode } = String;
+
+test("Can escape null character correctly", t => {
+  t.equal(escape(fromCharCode(Char.Null)), "\ufffd");
```
kasperisager: Do:

```ts
t.equal(escape("\u0000"), "\ufffd");
```

PR #31, opened 2018-08-10
Increasing coverage and fixing a bug

NiclasHedam:
This PR increases the coverage of various parts of the code-base, as well as fixing a bug, where aria-hidden wasn't inherited down to child elements.

I've implemented a check for this inheritance using a while-loop that climbs the tree of nodes instead of recursively calling isVisible. I chose this solution because isVisible does more than just checking for aria-hidden, which would cause an unnecessary performance penalty.

### DIFF
--- a/packages/alfa-aria/src/is-visible.ts
+++ b/packages/alfa-aria/src/is-visible.ts
@@ -1,6 +1,7 @@
 import {
   Element,
   getAttribute,
+  getClosest,
   getComputedStyle,
   getParentElement,
   isElement,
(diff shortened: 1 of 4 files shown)


packages/alfa-aria/src/is-visible.ts
```diff
@@ -15,6 +15,15 @@ export function isVisible(node: Element | Text, context: Node): boolean {
...
 
+    let parent = getParentElement(node, context);
+    while (parent !== null) {
+      if (getAttribute(parent, "aria-hidden") === "true") {
+        return false;
+      }
+
+      parent = getParentElement(parent, context);
+    }
+
```
kasperisager: The `getClosest()` function should be used in combination with the `flattened: true` option to ensure the function behaves correctly in Shadow DOM:

```ts
if (
  getClosest(
    node,
    context,
    node => isElement(node) && getAttribute(node, "aria-hidden") === "true",
    { flattened: true }
  ) !== null
) {
  return false;
}
```

PR #32, opened 2018-08-13
Adding support for hex colors in alfa-css

NiclasHedam:
I chose to try and challenge myself with a little exercise. I've therefore implemented support for hex colours in the `alfa-css` color grammar.

### DIFF
--- a/packages/alfa-css/src/alphabet.ts
+++ b/packages/alfa-css/src/alphabet.ts
@@ -18,6 +18,7 @@ export const enum TokenType {
   Ident,
   FunctionName,
   AtKeyword,
+  Hash,
   String,
   Url,
   Delim,
@@ -44,6 +45,12 @@ export type FunctionName = Readonly<{
 }>;
 
 export type AtKeyword = Readonly<{ type: TokenType.AtKeyword; value: string }>;
+
+export type Hash = Readonly<{
+  type: TokenType.Hash;
+  typeFlag: "id" | "unrestricted";
+  value: string;
+}>;
 
 export type String = Readonly<{
   type: TokenType.String;
@@ -102,6 +109,7 @@ export type Token =
   | Ident
   | FunctionName
   | AtKeyword
+  | Hash
   | String
   | Url
   | Delim
@@ -654,6 +662,35 @@ function consumeToken(stream: Stream<number>): Token | null {
       if (startsValidEscape(char, stream)) {
         return consumeIdentLike(char, stream);
       }
+      break;
+
+    case Char.NumberSign:
+      stream.advance(1);
+      const next = stream.peek(0);
+      const next2 = stream.peek(1);
+      if (
+        (next !== null && isName(next)) ||
+        (next !== null && next2 !== null && isValidEscape(next, next2))
+      ) {
+        if (startsIdentifier(next, stream)) {
+          return {
+            type: TokenType.Hash,
+            value: consumeName(char, stream),
+            typeFlag: "id"
+          };
+        }
+
+        return {
+          type: TokenType.Hash,
+          value: consumeName(char, stream),
+          typeFlag: "unrestricted"
+        };
+      }
+
+      return {
+        type: TokenType.Delim,
+        value: char
+      };
   }
 
   if (isNameStart(char)) {
--- a/packages/alfa-css/src/alphabet.ts
+++ b/packages/alfa-css/src/alphabet.ts
@@ -127,11 +127,14 @@ export type Token =
 
 export const Alphabet: Lang.Alphabet<Token> = new Lang.Alphabet(
   (stream, emit) => {
-    let token = consumeToken(stream);
+    while (true) {
+      const token = consumeToken(stream);
 
-    while (token !== null) {
+      if (token === null) {
+        break;
+      }
+
       emit(token);
-      token = consumeToken(stream);
     }
 
     return Command.End;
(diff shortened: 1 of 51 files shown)


NiclasHedam:
Hah, great! I seemingly broke a test in alphabet. Hold on.

NiclasHedam:
Here we go - it should be ready for review now :-)

packages/alfa-css/src/alphabet.ts
```diff
@@ -45,6 +46,12 @@ export type FunctionName = Readonly<{
 
 export type AtKeyword = Readonly<{ type: TokenType.AtKeyword; value: string }>;
 
+export type Hash = Readonly<{
+  type: TokenType.Hash;
+  typeFlag: "id" | "unrestricted";
```
kasperisager: Given that these are the only options I'd do the following to make the type a little less awkward:

```ts
export type Hash = Readonly<{
  type: TokenType.Hash;
  unrestricted: boolean;
  value: string;
}>;
```

packages/alfa-css/src/alphabet.ts
```diff
@@ -657,6 +665,35 @@ function consumeToken(stream: Stream<number>): Token | null {
       if (startsValidEscape(char, stream)) {
         return consumeIdentLike(char, stream);
       }
+      break;
+
+    case Char.NumberSign:
+      stream.advance(1);
+      const next = stream.peek(0);
+      const next2 = stream.peek(1);
```
kasperisager: Whenever more than one lookahead is made, I tend to name the variables `fst`, `snd`, `thd`, `fth`. etc.:

```ts
const fst = stream.peek(0);
const snd = stream.peek(1);
```

packages/alfa-css/src/alphabet.ts
```diff
@@ -657,6 +665,35 @@ function consumeToken(stream: Stream<number>): Token | null {
...
+            typeFlag: "id"
+          };
+        }
+
+        return {
+          type: TokenType.Hash,
+          value: consumeName(char, stream),
+          typeFlag: "unrestricted"
+        };
+      }
```
kasperisager: Can be simplified and optimised a tad:

```ts
const fst = stream.peek(0);

if (fst !== null) {
  const snd = stream.peek(1);

  if (isName(fst) || (snd !== null && isValidEscape(fst, snd))) {
    return {
      type: TokenType.Hash,
      unrestricted: !startsIdentifier(fst, stream),
      value: consumeName(fst, stream),
    };
  }
}
```

packages/alfa-css/src/alphabet.ts
NiclasHedam: Normally I would argue that we should follow the spec to the point so that if W3C introduces a third flag we can implement it easily. That being said, your solution is way more easy to read.

packages/alfa-css/src/alphabet.ts
NiclasHedam: Good idea.

packages/alfa-css/src/grammars/color.ts
```diff
@@ -99,13 +99,49 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+  if (hex.length !== 6) {
+    return Transparent;
+  }
+
+  const num = parseInt(hex, 16);
+  color.red = num >> 16;
+  color.green = (num >> 8) & 255;
+  color.blue = num & 255;
+
+  return color;
```
kasperisager: The parsing will need to be more robust as it currently doesn't handle things like non-hex characters and overflows. Here's a sketch of a more robust, and slightly more performant, approach:

1. Figure out the "repetition count" of the given hex value based on its length. For short hand colors (e.g. `#123`) the repetition count is 2 whereas for long hand colors (e.g. `#123456`) the repetition count is 1). If a repetition count cannot be determined, i.e. the length of the value is incorrect, return `Transparent`.
2. Figure out if an alpha channel is specified (e.g. `#1234` or `#123456ab`).
3. Run through every character and convert it to a number. An example of this is already implemented here: https://github.com/Siteimprove/alfa/blob/master/packages/alfa-css/src/alphabet.ts#L294-L300. If a non-hex character is encountered, return `Transparent`.
4. Append the number `n`-times to the channel (red, green, blue, alpha) corresponding to the character, where `n` is the repetition count. We're dealing with hex values so the code would be along the lines of:

    ```ts
    channel = (channel * 16) + value;
    ```

packages/alfa-css/src/grammars/color.ts
```diff
@@ -99,13 +99,78 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+      alpha: hasAlpha ? (bytes[6] * 16 + bytes[7]) / 255 : 1
+    };
+  }
+
+  return {
+    red: bytes[0] * 16 + bytes[0],
+    green: bytes[1] * 16 + bytes[1],
+    blue: bytes[2] * 16 + bytes[2],
+    alpha: hasAlpha ? (bytes[3] * 16 + bytes[3]) / 255 : 1
+  };
```
kasperisager: The above is close! You should however not be allocating a byte array and pushing the individual bytes into it; that's way more complex than things need to be. Allocating two variables for containing the final hex and alpha value is enough and then you need two loops: One for the hex value and one for the alpha value. For each encountered byte you then append it to the associated value (hex or alpha) as many times as specified by the repetition count.

packages/alfa-css/src/grammars/selector.ts
```diff
@@ -212,24 +213,20 @@ function isImplicitDescendant(token: Token): boolean {
...
 }
 
-function idSelector(stream: Stream<Token>): IdSelector | null {
-  const next = stream.next();
-
-  if (next === null || next.type !== TokenType.Ident) {
+function idSelector(token: Token): IdSelector | null {
+  if (token.type !== TokenType.Hash) {
     return null;
   }
```
kasperisager: You can safely declare the token type as `Hash` and lose the guard.

PR #33, opened 2018-08-13
Closes #21: Implement Shadow DOM `:host` CSS selector

### LINKED ISSUE #21: Implement Shadow DOM `:host` CSS selector
Implement the `:host` pseudo-class as described in CSS Scoping (https://drafts.csswg.org/css-scoping/#selectors). Parsing of these should already work–maybe include a test case to verify this–and so the remaining work lies in implementing these selectors in `matches()` (https://github.com/Siteimprove/alfa/blob/master/packages/alfa-dom/src/matches.ts). Most of what is needed for this should already exist in `@siteimprove/alfa-dom`.

2biazdk:


packages/alfa-css/src/types.ts
```diff
@@ -148,6 +148,8 @@ export type PseudoClass =
   | "required"
   // https://www.w3.org/TR/selectors/#user-invalid-pseudo
   | "user-invalid"
+  // https://www.w3.org/TR/css-scoping-1/#selectordef-host0
```
kasperisager: I'd simply link to https://www.w3.org/TR/css-scoping/#host-selector.

packages/alfa-dom/src/matches.ts
```diff
@@ -370,18 +420,29 @@ function matchesPseudoClass(
...
-  options: MatchesOptions
+  options: MatchesOptions,
+  root: Selector
 ): boolean {
   switch (selector.name) {
     // https://www.w3.org/TR/selectors/#scope-pseudo
     case "scope":
       return options.scope === element;
 
+    // https://www.w3.org/TR/css-scoping-1/#selectordef-host0
```
kasperisager: https://www.w3.org/TR/css-scoping/#host-selector

.gitignore
```diff
@@ -9,3 +9,5 @@ package-lock.json
 
 !**/build/**/*.js
 !**/types/**/*.d.ts
+
+.vscode
```
kasperisager: Looks like the newline at the end got removed.

packages/alfa-dom/src/matches.ts
```diff
@@ -370,18 +420,29 @@ function matchesPseudoClass(
...
-  options: MatchesOptions
+  options: MatchesOptions,
+  root: Selector
 ): boolean {
   switch (selector.name) {
     // https://www.w3.org/TR/selectors/#scope-pseudo
     case "scope":
       return options.scope === element;
 
+    // https://www.w3.org/TR/css-scoping/#host-selector
```
kasperisager: Upon closer inspection the current TR for CSS Scoping is horribly out of date so let's link to: https://drafts.csswg.org/css-scoping/#host-selector. I think there's a couple more references in other files as well.

PR #34, opened 2018-08-14
Implement functional :host pseudo-class in Matches

2biazdk:
Also add negative test cases.

### DIFF
--- a/packages/alfa-dom/src/matches.ts
+++ b/packages/alfa-dom/src/matches.ts
@@ -435,19 +435,25 @@ function matchesPseudoClass(
         return false;
       }
 
-      if (options.treeContext === undefined) {
+      if (
+        options.treeContext === undefined ||
+        !isShadowRoot(options.treeContext)
+      ) {
         return false;
       }
 
       const host = getParentElement(options.treeContext, context, {
         composed: true
       });
-      const isHostMatched =
-        selector.value === null ||
-        matches(element, context, selector.value, options, root);
 
+      if (host !== element) {
+        return false;
+      }
+
+      // Match host with possible selector argument (e.g. ":host(.foo)")
       return (
-        isShadowRoot(options.treeContext) && host === element && isHostMatched
+        selector.value === null ||
+        matches(element, context, selector.value, options, root)
       );
 
     // https://www.w3.org/TR/selectors/#negation-pseudo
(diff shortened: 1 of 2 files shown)


packages/alfa-dom/src/matches.ts
```diff
@@ -430,12 +430,24 @@ function matchesPseudoClass(
...
+        matches(element, context, selector.value, options, root);
+
       return (
-        root === selector &&
-        options.treeContext !== undefined &&
         isShadowRoot(options.treeContext) &&
         getParentElement(options.treeContext, context, { composed: true }) ===
-          element
+          element &&
+        isHostMatched
```
NiclasHedam: There is something funky about these indents, isn't there?

2biazdk: It. Is. Impossible to fix. 😭

kasperisager: You can pull out `getParentElement(options.treeContext, context, { composed: true })` to a variable and reference that instead. [Prettier](https://github.com/prettier/prettier) is pretty particular on indents 🙈

packages/alfa-dom/src/matches.ts
```diff
@@ -430,12 +430,24 @@ function matchesPseudoClass(
 
     // https://drafts.csswg.org/css-scoping/#host-selector
     case "host":
+      // Do not allow prefix (e.g. "div:host")
+      if (root !== selector) {
+        return false;
+      }
+
+      if (options.treeContext === undefined) {
```
kasperisager: This should also check for `isShadowRoot(options.treeContext)` so as to bail out early if the tree context isn't a shadow root.

packages/alfa-dom/src/matches.ts
```diff
@@ -430,12 +430,24 @@ function matchesPseudoClass(
...
+        return false;
+      }
+
+      const host = getParentElement(options.treeContext, context, {
+        composed: true
+      });
+      const isHostMatched =
+        selector.value === null ||
+        matches(element, context, selector.value, options, root);
+
```
kasperisager: Doing this...

```ts
if (host !== element) {
  return false;
}

return selector.value === null || matches(element, context, selector.value, options, root);
```

...will ensure that we don't attempt to do selector matching unless we absolutely have to.

PR #35, opened 2018-08-14
Adding support for HSLA colors in alfa-css color grammar

NiclasHedam:
This is pure black magic. If you have any comments, let me know.

### DIFF
--- a/packages/alfa-css/src/grammars/color.ts
+++ b/packages/alfa-css/src/grammars/color.ts
@@ -165,38 +165,57 @@ function hslaColor(stream: Stream<Token>): Color {
     }
   }
 
-  const { r, g, b } = hslToRgb(h / 60, s / 100, l / 100);
-
-  return { red: r, green: g, blue: b, alpha: a };
-}
-
-function hslToRgb(h: number, s: number, l: number) {
-  const snd = l <= 0.5 ? l * (s + 1) : l + s - l * s;
-  const fst = l * 2 - snd;
+  const rgb = hslToRgb(h / 60, s / 100, l / 100);
   const { round } = Math;
 
   return {
-    r: round(hueToRgb(fst, snd, h + 2) * 255),
-    g: round(hueToRgb(fst, snd, h) * 255),
-    b: round(hueToRgb(fst, snd, h - 2) * 255)
+    red: round(rgb[0] * 255),
+    green: round(rgb[1] * 255),
+    blue: round(rgb[2] * 255),
+    alpha: a
   };
 }
 
-function hueToRgb(fst: number, snd: number, hue: number) {
+/**
+ * @copyright  Copyright © 2016 W3C® (MIT, ERCIM, Keio, Beihang). W3C liability,
+ *             trademark and document use rules apply.
+ * @license    https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
+ */
+function hslToRgb(hue: number, sat: number, light: number) {
+  let t2;
+  if (light <= 0.5) {
+    t2 = light * (sat + 1);
+  } else {
+    t2 = light + sat - light * sat;
+  }
+  const t1 = light * 2 - t2;
+  const r = hueToRgb(t1, t2, hue + 2);
+  const g = hueToRgb(t1, t2, hue);
+  const b = hueToRgb(t1, t2, hue - 2);
+  return [r, g, b];
+}
+
+/**
+ * @copyright  Copyright © 2016 W3C® (MIT, ERCIM, Keio, Beihang). W3C liability,
+ *             trademark and document use rules apply.
+ * @license    https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document
+ */
+function hueToRgb(t1: number, t2: number, hue: number) {
+  if (hue < 0) {
+    hue += 6;
+  }
   if (hue >= 6) {
     hue -= 6;
-  } else if (hue < 0) {
-    hue += 6;
   }
 
   if (hue < 1) {
-    return (snd - fst) * hue + fst;
+    return (t2 - t1) * hue + t1;
   } else if (hue < 3) {
-    return snd;
+    return t2;
   } else if (hue < 4) {
-    return (snd - fst) * (4 - hue) + fst;
+    return (t2 - t1) * (4 - hue) + t1;
   } else {
-    return fst;
+    return t1;
   }
 }
 
--- a/packages/alfa-css/src/grammars/color.ts
+++ b/packages/alfa-css/src/grammars/color.ts
@@ -1,9 +1,11 @@
 import * as Lang from "@siteimprove/alfa-lang";
-import { Grammar, Stream } from "@siteimprove/alfa-lang";
+import { getNumericValue, Grammar, Stream } from "@siteimprove/alfa-lang";
 import { clamp, Mutable } from "@siteimprove/alfa-util";
-import { FunctionName, Ident, Token, TokenType } from "../alphabet";
+import { FunctionName, Hash, Ident, Token, TokenType } from "../alphabet";
 import { whitespace } from "../grammar";
 import { Color } from "../properties/color";
+
+const { min } = Math;
 
 const enum Component {
   Red = 0,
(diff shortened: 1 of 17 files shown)


packages/alfa-css/src/grammars/color.ts
```diff
@@ -101,6 +109,121 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+  return { red: r, green: g, blue: b, alpha: a };
+}
+
+/**
+ * Converts an HSL color value to RGB. Conversion formula
+ * adapted from http://en.wikipedia.org/wiki/HSL_color_space.
+ * Assumes h, s, and l are contained in the set [0, 1] and
+ * returns r, g, and b in the set [0, 255].
+ *
+ * @see https://stackoverflow.com/questions/2353211/hsl-to-rgb-color-conversion
```
kasperisager: The function will have to be reimplemented; the code below is not licensed. Use: https://www.w3.org/TR/css-color/#hsl-to-rgb

kasperisager: ...and add a copyright notice and a license link to https://www.w3.org/Consortium/Legal/2015/copyright-software-and-document.

packages/alfa-css/src/grammars/color.ts
```diff
@@ -101,6 +109,97 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+        a = value;
+    }
+  }
+
+  const { r, g, b } = hslToRgb(h / 60, s / 100, l / 100);
+
+  return { red: r, green: g, blue: b, alpha: a };
+}
+
+function hslToRgb(h: number, s: number, l: number) {
```
kasperisager: Use full names for the variables to make things easier to read. Applies to some of the above declarations as well.

kasperisager: Start off by clamping the hue, saturation, and lightness according https://www.w3.org/TR/css-color/#hsl-to-rgb rather than round later on.

packages/alfa-css/src/grammars/color.ts
```diff
@@ -101,6 +109,97 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+
+  if (hue < 1) {
+    return (snd - fst) * hue + fst;
+  } else if (hue < 3) {
+    return snd;
+  } else if (hue < 4) {
+    return (snd - fst) * (4 - hue) + fst;
+  } else {
+    return fst;
+  }
```
kasperisager: No need for `else` after a block that `return`s.

packages/alfa-css/src/grammars/color.ts
```diff
@@ -101,6 +109,116 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+  let l = 0;
+  let a = 1;
+
+  for (let i = 0, n = args.length; i < n; i++) {
+    const component = args[i];
+
+    switch (i) {
+      case 0:
+      case 3:
+        if (args[i].type !== TokenType.Number) {
```
kasperisager: `args[i]` => `component`

packages/alfa-css/src/grammars/color.ts
```diff
@@ -101,6 +109,116 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+        }
+        break;
+      case 1:
+      case 2:
+        if (args[i].type !== TokenType.Percentage) {
+          return Transparent;
+        }
+    }
+
+    let { value } = <Number | Percentage>component;
```
kasperisager: Not needed with the above change.

packages/alfa-css/src/grammars/color.ts
```diff
@@ -101,6 +109,116 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+        h = value;
+        break;
+      case Component.Green:
+        s = value;
+        break;
+      case Component.Blue:
+        l = value;
+        break;
+      case Component.Alpha:
+        a = value;
```
kasperisager: The `Component` enum doesn't make sense for HSL(A) so just use numeric constants instead, that's fine.

packages/alfa-css/src/grammars/color.ts
```diff
@@ -101,6 +109,116 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+    }
+  }
+
+  const rgb = hslToRgb(h / 60, s / 100, l / 100);
```
kasperisager: This can be destructured:

```ts
const [red, green, blue] = hslToRgb(h / 60, s / 100, l / 100);
```

packages/alfa-css/src/grammars/color.ts
```diff
@@ -101,6 +109,116 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+        }
+    }
+
+    let { value } = <Number | Percentage>component;
+
+    if (i === Component.Alpha) {
+      value = clamp(value, 0, 1);
+    } else {
+      if (component.type === TokenType.Percentage) {
+        value *= 0x64;
```
kasperisager: `0x64` => `100`

packages/alfa-css/src/grammars/color.ts
NiclasHedam: I've added the HSL constants to the components enum, as they share the same fundamental structure.

packages/alfa-css/src/grammars/color.ts
```diff
@@ -101,6 +104,119 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+
+    switch (i) {
+      case 0:
+      case 3:
+        if (component.type !== TokenType.Number) {
+          return Transparent;
+        }
+        break;
+      case 1:
+      case 2:
```
kasperisager: Use the associated component enum.

packages/alfa-css/src/grammars/color.ts
```diff
@@ -101,6 +104,119 @@ function rgbaColor(stream: Stream<Token>): Color {
...
+      case Component.Light:
+        light = value;
+        break;
+      case Component.Alpha:
+        alpha = value;
+    }
+  }
+
+  const [red, green, blue] = hslToRgb(hue / 60, saturation / 100, light / 100);
+  const { round } = Math;
```
kasperisager: Move this to just below the imports at the top of the module.

packages/alfa-css/src/grammars/color.ts
```diff
@@ -101,6 +104,119 @@ function rgbaColor(stream: Stream<Token>): Color {
   return color;
 }
 
+function hslaColor(stream: Stream<Token>): Color {
+  const args = functionArguments(stream);
+
+  if (args.length < 3) {
```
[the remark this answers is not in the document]
NiclasHedam: True, but then we'll have a dead block in the default case of the switch, which for me is a bit infuriating (Especially with the new coverage tool).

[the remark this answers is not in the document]
kasperisager: You don't need that case as you've already validated the arguments length.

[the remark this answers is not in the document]
NiclasHedam: That is not correct. The for/switch validates the indexes from inclusive 0 to inclusive 3. Typescript cannot infer that the loop will never exceed `i = 3;` and therefore requires me to typecast `let { value } = component;` as only `i <= 3` are validated.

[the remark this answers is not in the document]
kasperisager: TypeScript does no inference on array accesses that may be out of bounds so the above is not true. Validating the array length up front is therefore sufficient.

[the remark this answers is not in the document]
NiclasHedam: As discussed before, I'll put this pull request on hold, so that you can find a "viable" solution to the for/switch problem.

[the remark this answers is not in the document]
kasperisager: So the issue turned out to be TypeScript not considering switch cases over the component enum as exhaustive. I need to do some general refactoring of both the `rgbaColor()` and `hslaColor()` functions in order to accommodate this.

kasperisager:
@NiclasHedam I totally forgot the fully read the spec before reviewing this 🙈 The grammar is not yet complete as the hue is defined as:

```
<hue> = <number> | <angle>
```

...and the alpha value defined as:

```
<alpha-value> = <number> | <percentage>
```

The implemented grammar only supports `<number>` for the hue and `<percentage>` for the alpha value.

NiclasHedam:
@kasperisager Haha, interesting. I just read the spec and I did not come to the same conclusion; what goes for the hue, the following is stated in the spec: 

> This angle is so typically measured in degrees that the unit is implicit in CSS; syntactically, only a number is given.

The number you can give in the HSL grammar is already going from 0-360.

The spec also states the following about the alpha-value: 

> Syntactically a number. The uniform opacity setting to be applied across an entire object. Any values outside the range 0.0 (fully transparent) to 1.0 (fully opaque) will be clamped to this range.

I think you read the CSS Color Module Level 4, which is in draft. The currently active recommendations is [Level 3](https://www.w3.org/TR/2018/REC-css-color-3-20180619/)

PR #36, opened 2018-08-15
Color browser support

NiclasHedam:


packages/alfa-compatibility/src/supported-browsers.ts
```diff
@@ -20,7 +20,7 @@ export function getSupportedBrowsers(): ReadonlyArray<string> {
  * browser scope and resetting to the previous browser scope when done.
  */
 export function setSupportedBrowsers(
-  browsers: string | ReadonlyArray<string>,
```
kasperisager: Both of the above arrays must be readonly.

NiclasHedam: Sorry, I had some issues earlier, of which I cannot reproduce. I've put the type back as it was.

PR #37, opened 2018-08-15
Font-weight property in alfa-css

NiclasHedam:


packages/alfa-css/src/grammars/font-weight.ts
```diff
@@ -0,0 +1,48 @@
...
+      case 300:
+      case 400:
+      case 500:
+      case 600:
+      case 700:
+      case 800:
+      case 900:
+        return { type: "absolute", value: token.value };
+    }
+    return { type: "absolute", value: 400 };
```
kasperisager: This is not quite according to spec. The grammar looks as such:

```
<font-weight-absolute> = [normal | bold | <number>]
```

For `<number>`, it is stated that:

> Only values greater than or equal to 1, and less than or equal to 1000, are valid, and all other values are treated as parse errors.

packages/alfa-css/src/properties/font-weight.ts
```diff
@@ -0,0 +1,31 @@
...
+  // https://www.w3.org/TR/css-fonts/#font-weight-absolute-values
+  | {
+      type: "absolute";
+      value: 100 | 200 | 300 | 400 | 500 | 600 | 700 | 800 | 900;
+    }
+  | {
+      type: "relative";
+      value: "bolder" | "lighter";
+    }
+>;
```
kasperisager: Can be simplified to:

```ts
export type FontWeight = number | "bolder" | "lighter";
```

packages/alfa-css/src/properties/font-weight.ts
```diff
@@ -0,0 +1,31 @@
...
+export const FontWeightProperty: Property<FontWeight> = {
+  inherits: true,
+  parse(input) {
+    return parse(input, FontWeightGrammar);
+  },
+  initial() {
+    return { type: "absolute", value: 400 };
+  },
+  computed(getProperty) {
+    return getProperty("fontWeight");
```
kasperisager: Per the spec, the computed value of the `font-weight` property is a numeric weight value. The above will simply return the specified or inherited value, even if it's `bolder` or `lighter`, neither of which are numeric values. You will need to implement the `font-weight` step of the [font matching algorithm](https://www.w3.org/TR/css-fonts/#font-matching-algorithm) in order to determine to computed `font-weight` value. For now, we can assume that all weights are available regardless of `font-family`; we have yet to determine how to handle `font-family`, see https://github.com/auto-wcag/auto-wcag/issues/135.

packages/alfa-css/src/grammars/font-weight.ts
NiclasHedam: Not sure I agree about the definition. The spec (see below) states that the number can be any hundred from inclusive 100 to inclusive 900. In the spec linked below nothing is written about parse errors, so currently it just falls back to the "normal". Maybe it should return null?

image
Source: [https://www.w3.org/TR/css-fonts-3/#font-weight-prop](CSS Fonts Module Level 3).

packages/alfa-css/src/properties/font-weight.ts
NiclasHedam: Not if there is a limited about of numbers to pick from.

NiclasHedam: And do you want to discard the relativity attribute? I suppose we need it to infer whether or not to look at a parent element.

packages/alfa-css/src/grammars/font-weight.ts
kasperisager: Well ain't this peachy; there's nothing to disagree about, we're simply looking at different levels of the same spec 🤦‍♂️ I'm looking at https://www.w3.org/TR/css-fonts/ which currently contains a working draft of the level 4 fonts module instead of the latest level 3 fonts module. To make matters worse, https://drafts.csswg.org/css-fonts/ actually contains the latest level 3 fonts module... I give up 🙄

Irrespective of the confusion brought on by the beautifully inconsistent publishing scheme of the W3C, we have to support both grammars though. This is however going to be interesting as Can I Use does _not_ contain any data on the level 4 fonts module. MDN has _some_ data (https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight#Browser_compatibility), but it's incorrect and incomplete.

packages/alfa-css/src/properties/font-weight.ts
kasperisager: See https://github.com/Siteimprove/alfa/pull/37#discussion_r210486158. We only need to know whether or not `font-weight` is numeric and then we look up the tree and apply the rules for `bolder` and `lighter` if it isn't.

packages/alfa-css/src/properties/font-weight.ts
```diff
@@ -0,0 +1,46 @@
...
+      case 400:
+      case 500:
+        return isLighter ? 100 : 700;
+      case 600:
+      case 700:
+        return isLighter ? 400 : 900;
+      case 800:
+      case 900:
+        return isLighter ? 700 : 900;
+    }
```
kasperisager: The above should check for ranges instead of fixed values. If, for example, the inherited value is 480 then the computed value in case of `bolder` should be 700.

NiclasHedam: Sorry, I followed the "wrong" spec. Looking at you again, Level 3.

packages/alfa-css/src/properties/font-weight.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    if (parentValue >= 750 && parentValue <= 899) {
+      return isBolder ? 900 : 700;
+    }
+
+    if (parentValue >= 900 && parentValue <= 999) {
+      return isBolder ? parentValue : 700;
+    }
+
+    return 400;
```
kasperisager: As you know that `1 <= parentValue <= 1000` you can leave out the last case and simply do:

```ts
return isBolder ? parentValue : 700;
```

packages/alfa-css/src/properties/font-weight.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    const parentValue = getParentProperty("fontWeight");
+    const isBolder = value === "bolder";
+
+    switch (parentValue) {
+      case undefined:
+      case "lighter":
+      case "bolder":
+        return 400;
+    }
```
kasperisager: You can get rid of the `switch` by simply doing:

```ts
if (typeof parentValue !== "number") {
  return undefined;
}
```

If `computed()` cannot determine a computed value, it should return `undefined` in which case `initial()` will kick in.

NiclasHedam: Why not null?

PR #39, opened 2018-08-16
Implement host-context pseudo-class in matches

2biazdk:


packages/alfa-dom/test/matches.spec.tsx
```diff
@@ -183,3 +183,21 @@ test("Matches an element against a functional host selector", t => {
...
+        <span class="child" />
+      </shadow>
+    </div>
+  );
+  const context = <div class="foo">{host}</div>;
+
+  const root = host.shadowRoot!;
+
+  t(matches(host, context, ":host-context(.foo)", { treeContext: root }));
+  t(matches(host, host, ":host-context(.bar)", { treeContext: root }));
```
kasperisager: I'm guessing you mean for this to be `matches(host, context, ...)`?

packages/alfa-dom/src/matches.ts
```diff
@@ -429,22 +430,19 @@ function matchesPseudoClass(
...
       // Do not allow prefix (e.g. "div:host")
       if (root !== selector) {
         return false;
       }
 
-      if (
-        options.treeContext === undefined ||
-        !isShadowRoot(options.treeContext)
-      ) {
+      const treeCtx = options.treeContext;
```
kasperisager: I'm reminded a bit of the whole Golang `ctx context.Context` "stutter" when reading this 🙈 This should stutter a little less:

```ts
const { treeContext } = options;
```

2biazdk: Sure. 😁 I was trying to avoid the Prettier formatting kicking in by keeping it short.

I guess I could also do:
```const { treeContext: treeCtx } = options;```
but I'll go for the object destructuring to avoid the redundancy.

packages/alfa-dom/test/matches.spec.tsx
2biazdk: Oh, yeah. That wasn't supposed to slip through.

PR #42, opened 2018-08-16
Tests/test output configuration

Bugvi-BM:
The branch name is unfortunately misleading.

I have added some tests for the get-closest function in Alfa-dom. 

Furthermore, I have tweaked the test output a bit. It should now filter out blocks that do not make much sense. Also, I made it look a bit nicer with the statement "Suggested Block to Cover" only printed once instead of once per suggested block.


build/helpers/coverage.js
```diff
@@ -92,7 +92,8 @@ process.on("exit", code => {
 
       printCoverageStatistics(script, total);
 
-      if (process.env.npm_lifecycle_event === "start") {
+      if (process.env.npm_lifecycle_event === "start" && uncovered.length > 0) {
+        process.stdout.write(chalk.bold("\nSuggested Blocks to Cover:\n"));
```
kasperisager: Normal casing would be lovely 👍

build/helpers/coverage.js
```diff
@@ -281,13 +282,7 @@ function parseFunctionCoverage(script, map, range, name) {
...
  */
 function parseBlockCoverage(script, map, range) {
-  let parsed = parseRange(script, map, range, {
-    trim: true
-  });
-
-  if (parsed === null) {
-    parsed = parseRange(script, map, range);
-  }
+  let parsed = parseRange(script, map, range);
```
kasperisager: The non-trimming retry is needed in cases where trimming causes source maps to refuse mapping the given range.

build/helpers/coverage.js
```diff
@@ -317,6 +318,11 @@ function parseRange(script, map, range, options = {}) {
   startOffset = max(first.start, startOffset - header.length);
   endOffset = min(last.end, endOffset - header.length);
 
+  const uncovered = content.substring(startOffset, endOffset).trim();
+  if (isBlockBorder(uncovered) || uncovered === "") {
```
kasperisager: `isWhitespace(uncovered)`?

NiclasHedam:
@kasperisager Búgvi made the last changes, so I don't think there is anything to look at.

kasperisager:
I prefer to hold off on merging code that the authors haven't had a chance to look at just prior to the merge. I all too often have realisations of "shoot, I could have done this differently, but now I need to open a new PR!" 🙈

Bugvi-BM:
I have cleaned up the test output. 
- Added separators between blocks.
- Made blocks aligned so that it looks nicer.
- Handled several cases related to the alignment of the blocks. 

The code is probably way to complex and can be simplified and optimized. Tips and suggestions would be welcome :)

kasperisager:
I've adjusted the output slightly so it's more in line with the test output:

image

image

Might a similar legend be good to add?

NiclasHedam:
:+1: Looks good!

PR #43, opened 2018-08-17
Feature adjusting test metric weights

Bugvi-BM:
I figured a heuristic/metric shouldn't have an impact on the calculated coverage if it wasn't applicable. So I just went ahead and tried some things. Please let me know your thoughts and if I am totally off here :)


build/helpers/coverage.js
```diff
@@ -24,19 +24,18 @@ const session = new Session();
...
+    ...Arithmetic
   },
   {
-    ...Byte,
-    weight: 1 / 3
+    ...Byte
   },
   {
-    ...Logical,
-    weight: 1 / 3
```
NiclasHedam: Since you have removed the second element of the metric-object, this can be reduced to something like `const metrics = [Arithmetic, Byte, Logical];`

build/helpers/coverage.js
```diff
@@ -24,19 +24,18 @@ const session = new Session();
...
+    ...Logical
   }
 ];
 
+let weight = metrics.length;
+
```
kasperisager: There's quite a bit of global state introduced in this PR. It's nothing that a bit of refactoring can't fix though; be smart about passing state around as function parameters and you should be able to get rid of the global state.

PR #46, opened 2018-08-24
Closes #20: Implement support for CSS namespaces in `matches()`

### LINKED ISSUE #20: Implement support for CSS namespaces in `matches()`
Let's start this one out by implementing namespace support in `matches()` according to: https://www.w3.org/TR/selectors/#type-nmsp.

2biazdk:


packages/alfa-dom/src/types.ts
```diff
@@ -432,3 +432,10 @@ export const enum Namespace {
    */
   XMLNS = "http://www.w3.org/2000/xmlns/"
 }
+
+/**
+ * Map of namespace alias (key) and namespace URI (value).
+ *
+ * @see https://www.w3.org/TR/selectors/#type-nmsp
+ */
+export type NamespaceDeclarations = Map<string | null, Namespace>;
```
kasperisager: The namespace map is only ever used internally in the package so you can leave out the type alias.

packages/alfa-dom/src/matches.ts
```diff
@@ -178,15 +200,62 @@ function matchesClass(element: Element, selector: ClassSelector): boolean {
...
+    return true;
+  }
+
+  const elementNamespace = getElementNamespace(element, context);
+
+  // As all elements in HTML5 will use the XHTML namespace,
+  // unless another namespace is explicitly specified, the
+  // only way to currently test the no-namespace selector
+  // (i.e. "|div") is to provide a disconnected context.
+  // https://www.w3.org/TR/selectors/#type-nmsp
```
kasperisager: This comment should be moved to the spec file as it doesn't really relate to the implementation.

packages/alfa-dom/src/matches.ts
```diff
@@ -178,15 +200,62 @@ function matchesClass(element: Element, selector: ClassSelector): boolean {
...
+
+  if (declaration === undefined) {
+    return false;
+  }
+
+  if (elementNamespace !== declaration) {
+    return false;
+  }
+
+  return true;
```
kasperisager: Can be reduced to:

```ts
return elementNamespace === declaration;
```

I do like the explicitness of pulling out the declared namespace and checking if it's `undefined`. Maybe rename `declaration` to `declaredNamespace` though?

packages/alfa-dom/src/matches.ts
```diff
@@ -178,15 +200,62 @@ function matchesClass(element: Element, selector: ClassSelector): boolean {
...
+ */
+function matchesNamespace(
+  element: Element,
+  context: Node,
+  selector: TypeSelector,
+  options: MatchesOptions
+): boolean {
+  if (selector.namespace === "*" || options.namespaces === undefined) {
+    return true;
+  }
```
kasperisager: In the event that `options.namespaces` is `undefined`, `matchesNamespace()` should return `true` _only if_ `selector.namespace` is also `null`.

PR #47, opened 2018-08-28
Style-checker

NiclasHedam:
Currently, prettier formats all staged files as part of the pre-commit hook, but it is possible to circumvent this both intentionally and accidentally. I suggest that a style-check should be part of the continuous integration tool. The style-check being a part of the CI forces all users to commit formatted code no matter their environment.

The next step could be to create yet another helper to run the prettier on all files or a single file so that you do not have to stage files and commit to format them.


NiclasHedam:
I would have preferred to have it being stand-alone so that I would be able to invoke it without running `CI=true yarn prepare`, which also unintentionally stages files. However, I accept your suggestion, as the code is much better and more maintainable.

build/tasks/format.js
```diff
@@ -25,6 +27,10 @@ function format(file) {
     return true;
   }
 
+  if (process.env.CI !== "true") {
+    notify.skip(chalk.gray(file));
+  }
+
```
kasperisager: This can be removed as `precommit.js` takes care of notifying about skipped files.

NiclasHedam: got it!

PR #48, opened 2018-09-06
Increasing tests of alfa-util

NiclasHedam:
Alfa-util might be simple and easy to understand, but since it can be referenced all over the Alfa project, I believe it deserves some tests. The tests task are mainly to protect from unexpected regressions.


NiclasHedam:
Why did semaphore die mid-build?
https://semaphoreci.com/siteimprove/alfa/branches/feature-increasing-tests-in-util/builds/1

packages/alfa-util/test/array/map.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { test } from "@siteimprove/alfa-test";
+import { map } from "../../src/array/map";
+
+test("Can group by element in array", t => {
```
kasperisager: Copy/pasta gone wrong?

Issue #49, opened 2018-09-06
Coverage warnings are missing for completely untested files


NiclasHedam:
When running `yarn test`, the build system notifies the builder about any files with a coverage lower than 90%. Unfortunately, it completely ignores files that are entirely untested, as it never invokes the coverage suggestion tool. The aforementioned gives an incorrect overview of the coverage and should be corrected.

NiclasHedam:
An issue could be to detect which files can't be tested such as `/packages/alfa-crypto/src/types.ts`.

kasperisager:
The simple approach is to add en empty spec file for all the files that need tests. No magic, no complicated code, just conventions 👍

NiclasHedam:
Do you mean manually adding the spec files? I think that will be forgotten and overlooked in the long run - that is why I want to create an automatic solution.

kasperisager:
Combining the two approaches is where the benefit lies; have the build system complain about missing spec files and then leave it at that. Every implementation file then needs an associated spec file, but the spec files can be empty if needed.

kasperisager:
@NiclasHedam That's not always the case. Consider this:

```ts
enum Whatever {
  Foo = 1,
  Bar = 2,
  Baz = 3,
  Qux = Flags.Foo + Flags.Bar
}
```

We'd never want to test the above even though it includes arithmetic.

PR #51, opened 2018-09-11
Detection of files with missing spec file

NiclasHedam:
This tries to address #49 as simple as possible. I think this solution seems rather horrifying, so I would like some advice on how to refactor it.


NiclasHedam:
All good now!

kasperisager:
Looking at the output, this really causes way too many false positives to be useful currently, so I've reverted the commit ☹️

NiclasHedam:
What do you recommend doing then?

kasperisager:
Given that automatically determining testable code does not really seem like an option, my previous recommendation still stands; add empty spec files for the files that _do_ need tests.

kasperisager:
We have hundreds of files that can't and shouldn't be tested (e.g. files containing nothing but data), so a whitelist approach also seems suck 😢

NiclasHedam:
Absolutely, but the status quo is not acceptable either. We cannot just ignore untested files as if there are no issues with them.

kasperisager:
Unless we can reliably call them out, there's little else we can do but ignore them though. I'm happy to consider a revised approach, but reliability is a requirement.

PR #54, opened 2018-09-18
Closes #49: Detect untested files

### LINKED ISSUE #49: Coverage warnings are missing for completely untested files

NiclasHedam:
As discussed in person, precise heuristics to detect untested files are required. These heuristics must not create false positives.

This initial version extends the prepare method to look for source files without an associated test file. The unassociated files will then be compiled to search for exported functions.

### DIFF
--- a/scripts/prepare.js
+++ b/scripts/prepare.js
@@ -44,6 +44,10 @@ for (const pkg of packages) {
   );
 
   for (const file of files) {
+    if (file.indexOf("/src/") === -1) {
+      continue; // File is not in source folder
+    }
+
     const dir = path
       .dirname(file)
       .split(path.sep)
--- a/scripts/prepare.js
+++ b/scripts/prepare.js
@@ -12,11 +12,6 @@ const notify = require("./helpers/notify");
 const { build } = require("./tasks/build");
 const { clean } = require("./tasks/clean");
 
-const { Logical } = require("./helpers/metrics/logical");
-const { Arithmetic } = require("./helpers/metrics/arithmetic");
-
-const metrics = [Arithmetic, Logical];
-
 /**
  * @param {Array<string>} files
  */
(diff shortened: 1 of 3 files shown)


NiclasHedam:
By the way; this solution is modular. It is easy to add cases to which testing is required.

scripts/prepare.js
```diff
@@ -35,11 +40,71 @@ for (const pkg of packages) {
...
+      path.join(dir, `${path.basename(file, ".ts")}.spec.tsx`)
+    ];
+
+    if (potentialTestFiles.some(file => isFile(file))) {
+      continue; // An associated test file was found
+    }
+
+    const compiled = TypeScript.createSourceFile(
+      "anon.ts",
+      fs.readFileSync(file).toString(),
```
kasperisager: Use `readFile()` from `file-system.js`.

scripts/prepare.js
```diff
@@ -35,11 +40,71 @@ for (const pkg of packages) {
...
+      path.join(dir, `${path.basename(file, ".ts")}.spec.ts`),
+      // TSX source with TS test file
+      path.join(dir, `${path.basename(file, ".tsx")}.spec.ts`),
+      // TSX source with TSX test file
+      path.join(dir, `${path.basename(file, ".tsx")}.spec.tsx`),
+      // TS source with TSX test file
+      path.join(dir, `${path.basename(file, ".ts")}.spec.tsx`)
+    ];
+
+    if (potentialTestFiles.some(file => isFile(file))) {
```
kasperisager: No need to wrap in another lambda:

```ts
potentialTestFiles.some(isFile)
```

PR #55, opened 2018-09-18
Alfa Dom Tests

Bugvi-BM:
Just a bunch of tests for the dom package. 

I have many more tests to add but I figured I should just get these tests merged into master.


packages/alfa-dom/test/get-input-type.spec.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const input = <input type="radio" />;
+  t.equal(getInputType(input), InputType.Radio);
+});
+
+test("getInputType returns submit for input element with submit type", t => {
+  const input = <input type="submit" />;
+  t.equal(getInputType(input), InputType.Submit);
+});
+
+test("getInputType returns text for input element with undefined type", t => {
```
NiclasHedam: Wouldn't a better title på `getInputType defaults to text when not specified`

packages/alfa-dom/test/contains.spec.tsx
```diff
@@ -0,0 +1,23 @@
...
+import { contains } from "../src/contains";
+
+test("Contains returns true when a node contains the specifed node using string query", t => {
+  const strong = <strong />;
+  const div = <div>{strong}</div>;
+
+  t.equal(contains(div, <body>{div}</body>, "strong"), true);
+});
+
+test("Contains returns true when a node contains the specifed node using object query", t => {
```
NiclasHedam: The titles in this test file is a bit confusing - especially the last part of the titles. Alternatives could be:
`Contains returns true when element is contained in parent element.`
`Contains returns true when tag name is contained in parent element.`

PR #56, opened 2018-09-18
Optmizing code output

### LINKED ISSUE #53: Truncation of coverage suggestion output
The test output should be formatted such that large blocks are truncated. In addition, if one of the suggested uncovered blocks is really large, the number of blocks outputted should be cut down.

NiclasHedam:
This changes the output of the coverage helper a bit:
- It will now print as many blocks as there is room for (not three), but minimum 1.
- ~It was sorting based on line number, which I removed (We don't know how many is going to be printed, so we can't slice it).~

### DIFF
--- a/scripts/helpers/coverage.js
+++ b/scripts/helpers/coverage.js
@@ -108,38 +108,43 @@ process.on("beforeExit", code => {
 
         process.stdout.write(chalk.underline(`${filePath}\n`));
 
-        const widths = {
-          // Make the gutter width as wide as the line number of the last line.
-          // Since line counts are 1-indexed and an additional line may be added
-          // to the output, we bump the line count twice to make sure the gutter
-          // is wide enough.
-          gutter: `${uncovered[uncovered.length - 1].range.end.line + 2}`.length
-        };
-
+        // We can fill out the whole height of the terminal
+        // If we don't know the height, fall back to 24.
+        const space = process.stdout.rows ? process.stdout.rows : 24;
         let lineSum = 5;
-        let i = 0;
+        let blockCount = 0;
 
-        while (true) {
-          if (uncovered[i] === undefined) {
-            break;
+        for (const block of uncovered) {
+          // Buffer added for dots and spacing
+          lineSum += block.range.end.line - block.range.start.line + 5;
+
+          // At least one block has to be printed
+          if (lineSum > space && blockCount > 0) {
+            break; // If the next block is going to fill more than the height
           }
 
+          blockCount++;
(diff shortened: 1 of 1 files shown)


NiclasHedam:
Right, I'll walk you through it. 

We cannot slice the array anymore, as we have no way of knowing how many blocks will be printed. It can be more than three if the blocks are small.
By not slicing, we now have an array of blocks sorted by complexity from the coverage metrics. By now re-sorting by line numbers, we lose the initial complexity sorting. The resorting means that only line numbers determine the shown blocks.

We can probably get around this by calculating how many elements there is space for and then slicing, but I honestly think that is overengineering it.

NiclasHedam:
I chose to overengineer it, as I am waiting for some other tasks to complete.
It now runs over the array of blocks once to determine how many blocks it can print, and then it will slice it, and then lastly print out the blocks in question.

kasperisager:
That's just the right amount of engineering; don't throw the baby out with the bathwater 😉

scripts/helpers/coverage.js
```diff
@@ -112,20 +108,43 @@ process.on("beforeExit", code => {
...
 
-          printBlockCoverage(script, uncovered[i], widths);
+          const widths = {
+            // Make the gutter width as wide as the line number of the last line.
+            // Since line counts are 1-indexed and an additional line may be added
+            // to the output, we bump the line count twice to make sure the gutter
+            // is wide enough.
+            gutter: `${selectedBlocks[selectedBlocks.length - 1].range.end
+              .line + 2}`.length
+          };
```
kasperisager: This can be moved outside the loop.

scripts/helpers/coverage.js
```diff
@@ -112,20 +108,43 @@ process.on("beforeExit", code => {
...
+          if (lineSum > space && blockCount > 0) {
+            break; // If the next block is going to fill more than the height
+          }
+
+          blockCount++;
+        }
+
+        const selectedBlocks = uncovered.slice(0, blockCount).sort((a, b) => {
+          return a.range.start.line - b.range.start.line;
+        });
```
kasperisager: Bueno 👍

scripts/helpers/coverage.js
kasperisager: This can be moved outside the loop.

NiclasHedam:
I would prefer to submit another pull-request, as I think this feature alone should be on master (So that we can revert easily)

NiclasHedam:
Thank you!

PR #62, opened 2018-10-04
\r are now replaced with the empty string. 

Bugvi-BM:
Moved an assignment out of lambda function.


scripts/helpers/coverage.js
```diff
@@ -602,6 +602,7 @@ function printBlockCoverage(script, coverage, widths) {
     );
   }
 
+  const eol = chalk.gray.dim("\u00ac");
```
kasperisager: Unused?

NiclasHedam:
Conflict resolved.

PR #64, opened 2018-10-09
Incorrect path separator in build phase

NiclasHedam:
The current `prepare.js` does not work correctly on Windows, due to the incorrect path separator.


NiclasHedam:
Due to a mistake, while debugging on Búgvis machine, Búgvis tests have been merged into this bugfix branch. For that reason, we need a new review.

kasperisager:
@NiclasHedam @Bugvi-BM Fixed it 👍

PR #65, opened 2018-10-09
Adding tests to the new option utility

NiclasHedam:


packages/alfa-util/test/option.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+test("Options runs some if parameter is something, else none", t => {
+  t.equal(
+    option(4, (num: number) => "something", () => "nothing"),
+    "something"
+  );
+  t.equal(
+    option(null, (num: number) => "something", () => "nothing"),
+    "nothing"
+  );
+});
```
kasperisager: You can get around having to annotate everything by defining these at the top:

```ts
let something: Option<number> = 4;
let nothing: Option<number> = null;
```

NiclasHedam: I think that is a great improvement. Thank you.

PR #67, opened 2018-10-11
Cyclomatic complexity metric

Bugvi-BM:
I added a new metric to be used for testing. It functions in a similar way as the other metrics.
See comment about branching from wrong branch. Should be fine though, as it was all meant to go to master anyways.

### DIFF
--- a/scripts/helpers/metrics/cyclomatic.js
+++ b/scripts/helpers/metrics/cyclomatic.js
@@ -1,0 +1,138 @@
+const TypeScript = require("typescript");
+
+/**
+ * @typedef {import("../coverage").Script} Script
+ */
+
+/**
+ * @typedef {import("../coverage").FunctionCoverage} FunctionCoverage
+ */
+
+/**
+ * @typedef {import("../coverage").BlockCoverage} BlockCoverage
+ */
+
+/**
+ * @param {String} source
+ */
+const createSource = source =>
+  TypeScript.createSourceFile(
+    "anon.ts",
+    source,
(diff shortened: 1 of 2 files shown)


Bugvi-BM:
Also, some alfa-dom tests are included in this as well. I must have branched from my testing branch instead of master.

scripts/helpers/metrics/cyclomatic.js
```diff
@@ -0,0 +1,138 @@
...
+      total += newNode.statements.length;
+      break;
+    case TypeScript.SyntaxKind.CatchClause:
+    case TypeScript.SyntaxKind.ConditionalExpression:
+    case TypeScript.SyntaxKind.DoStatement:
+    case TypeScript.SyntaxKind.ForStatement:
+    case TypeScript.SyntaxKind.ForInStatement:
+    case TypeScript.SyntaxKind.ForOfStatement:
+    case TypeScript.SyntaxKind.IfStatement:
+    case TypeScript.SyntaxKind.WhileStatement:
```
NiclasHedam: try-catches? switch? possibly functions? (especially anonymous)

NiclasHedam: probably also if else and else

Bugvi-BM: As discussed in person:
1. Try and else do not increase complexity, as they are already part of the control flow. 
2. Anonymous functions are covered
3. If else is covered, in the sense that if is covered. See 1. for else.
4. Switch is covered because each case statement is covered.

PR #69, opened 2018-10-16
Tests of alfa-tests

NiclasHedam:
This very meta pull request tries to address the coverage requirements of alfa-tests.
Alfa-test should probably be rewritten to allow dependency injection better, but for now, this should do the trick.
Unfortunately, the coverage system does not catch these tests, as they run in a separate process. They do so, as testing a failing test, would result in process termination.


packages/alfa-test/src/test.ts
```diff
@@ -2,9 +2,24 @@ import * as assert from "assert";
...
+    process.stderr.write(`${message}\n`);
+    process.exit(1);
+  }
+};
+
 export async function test(
   name: string,
-  assertion: (assert: Assertions) => void | Promise<void>
+  assertion: (assert: Assertions) => void | Promise<void>,
+  notifier = defaultNotifier
```
kasperisager: The signature that accepts a notifier will also have to marked as internal.

PR #71, opened 2018-10-16
Alphabet refactoring and tests

NiclasHedam:
I have refactored alphabet in alfa-html to use a stack to determine if a tag is an appropriate end tag. The beforementioned helper has also been split into two different functions, doing two separate things (updating the lexer state, and determining the next html state).


packages/alfa-html/src/alphabet.ts
```diff
@@ -2528,44 +2539,57 @@ function startsWith(
...
     case "plaintext":
-      if (tag.type === TokenType.StartTag) {
+      if (state.tag.type === TokenType.StartTag) {
         return plaintext;
       }
-      break;
+  }
+  return data;
+}
+function updateState(state: State) {
```
kasperisager: An empty line would the functions would be superb 👌

PR #74, opened 2018-10-23
Closes #73: Sanshikan rules are not seen as testable

### LINKED ISSUE #73: Sanshikan rules are not seen as testable
Currently, Sanshikan rules are not determined to be testable, as they export a constant as an object with multiple arrow functions. Constants and arrow functions are by themselves not seen as testable, as constants are, well, constant and arrow functions can also be constant in some contexts (see fx. [packages/alfa-aria/src/features/body.ts](https://github.com/Siteimprove/alfa/blob/master/packages/alfa-aria/src/features/body.ts)). Since Sanshikan rules should be tested, the `isTestable` heuristic needs to be extended.

NiclasHedam:
The missing-spec heuristic will now detect arrow functions, which have one or more parameters. If the arrow function takes parameters that it uses, it is prone to malicious input and should, therefore, be tested.


scripts/prepare.js
```diff
@@ -90,23 +90,34 @@ handle(findFiles("docs", endsWith(".ts", ".tsx")));
...
+  if (
+    node.modifiers &&
+    node.modifiers.some(el => el.kind === TypeScript.SyntaxKind.ExportKeyword)
+  ) {
+    exporting = true;
+  }
   switch (node.kind) {
-    case TypeScript.SyntaxKind.FunctionDeclaration:
+    case TypeScript.SyntaxKind.ArrowFunction:
+      const castedNode = /** @type {TypeScript.ArrowFunction} */ (node);
```
kasperisager: `castedNode` -> `arrowFunction`.

PR #76, opened 2018-10-25
alfa-bench test

NiclasHedam:
This tests the fundamental logic behind the alfa-bench package.


NiclasHedam:
I need to see some code on how you want to achieve this.
Currently, I import a clean benchmark.js instance and overwrite it's run-handler only to run once and return a predetermined result. This solution is the best I could come up with, while not doing a lot of nasty stuff.

packages/alfa-bench/src/benchmark.ts
```diff
@@ -8,25 +8,69 @@ interface Target {
...
+ */
+export interface Event {
+  readonly target: Target;
+}
+
+/**
+ * @internal
+ */
+export interface Notifier {
+  out: (message: string) => void;
```
kasperisager: `out` -> `print`

packages/alfa-bench/src/benchmark.ts
```diff
@@ -8,25 +8,69 @@ interface Target {
...
+ * @internal
+ */
+export interface Event {
+  readonly target: Target;
+}
+
+/**
+ * @internal
+ */
+export interface Notifier {
```
kasperisager: `Printer`?

packages/alfa-bench/src/benchmark.ts
```diff
@@ -8,25 +8,69 @@ interface Target {
...
+  out: (message: string) => void;
+}
+
+const defaultNotifier: Notifier = {
+  out: message => {
+    process.stderr.write(`${message}\n`);
```
kasperisager: `stdout`

packages/alfa-bench/test/benchmark.spec.ts
```diff
@@ -0,0 +1,115 @@
...
+      ]);
+    },
+    run: () => {},
+    add: (title: string, callback: () => void | Promise<void>) => suite
+  };
+
+  benchmark(suite, result, notifier);
+});
+
+test("Outputs stats in each cycle", t => {
```
kasperisager: I'd remove this.

NiclasHedam:
`printer.print(`${output}\n`);` is some really interesting code.

NiclasHedam:
Now this PR fails because master is broken. :joy:

PR #79, opened 2018-11-01
Coverage tools stopped working on never node versions

NiclasHedam:
Our coverage tool stopped working on some newer versions of node (10.12.0 and later). This issue is possibly related to https://github.com/nodejs/node/pull/22251.
Nonetheless, I have added a replace statement that is triggered if the protocol is prepended to the URL. This fixes the issue and should be backwards-compatible as well :-)


NiclasHedam:
See also https://github.com/nodejs/node/issues/23688

PR #80, opened 2018-11-01
Tests of alfa-css

NiclasHedam:
This does not include any tests of [packages/alfa-css/src/resolve-style.ts](https://github.com/Siteimprove/alfa/blob/master/packages/alfa-css/src/resolve-style.ts), as I think it is currently too complex for me to completely understand the intentions. This file could also possibly be refactored to be easier to maintain in the future.

### DIFF
--- a/packages/alfa-css/test/parse.spec.ts
+++ b/packages/alfa-css/test/parse.spec.ts
@@ -1,0 +1,67 @@
+import { test } from "@siteimprove/alfa-test";
+import { parseDeclaration, parseRule } from "../src/parse";
+
+test("Can parse a declaration", t => {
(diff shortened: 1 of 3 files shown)


packages/alfa-css/test/parse.spec.ts
```diff
@@ -0,0 +1,74 @@
+import { test } from "@siteimprove/alfa-test";
+import { parseDeclaration, parseRule, parseSelector } from "../src/parse";
+
+test("Can parse a declaration", t => {
+  t.deepEqual(parseDeclaration("text-color: #000 !important;"), [
+    {
+      name: "text-color",
+      value: [
+        {
+          type: 10
```
kasperisager: Should make use of `TokenType`. Same applies to the rest of the code.

PR #82, opened 2018-11-06
Tests for SIA-R1

NiclasHedam:
Shansikan rules are to be tested in Alfa. Currently, no tests exist for any Sanshikan rules, and therefore a 'standard' has to be established on how they are structured.

This PR is an initial draft of tests for SIA-R1.


packages/alfa-wcag/test/helpers/outcome.ts
```diff
@@ -0,0 +1,70 @@
...
+export function outcome<T extends Target, A extends Aspect>(
+  t: Assertions,
+  rule: Rule<A>,
+  aspects: AspectsFor<A>,
+  assert:
+    | Outcome.Inapplicable
+    | Readonly<
+        { [O in Outcome.Failed | Outcome.Passed | Outcome.CantTell]?: Array<T> }
+      >,
+  compositeRules: Array<Rule<A>> = []
```
kasperisager: `compositeRules` -> `dependencies`

packages/alfa-wcag/test/helpers/outcome.ts
```diff
@@ -0,0 +1,70 @@
...
+  Outcome,
+  Rule,
+  Target
+} from "@siteimprove/alfa-act";
+import { Assertions } from "@siteimprove/alfa-test";
+import { concat } from "@siteimprove/alfa-util";
+
+export function outcome<T extends Target, A extends Aspect>(
+  t: Assertions,
+  rule: Rule<A>,
```
kasperisager: `Rule<A>` -> `Rule<A, T>`

packages/alfa-wcag/test/helpers/outcome.ts
```diff
@@ -0,0 +1,70 @@
...
+  audit,
+  isResult,
+  Outcome,
+  Rule,
+  Target
+} from "@siteimprove/alfa-act";
+import { Assertions } from "@siteimprove/alfa-test";
+import { concat } from "@siteimprove/alfa-util";
+
+export function outcome<T extends Target, A extends Aspect>(
```
kasperisager: `<T extends Target, A extends Aspect>` -> `<A extends Aspect, T extends Target>`

packages/alfa-wcag/test/helpers/outcome.ts
```diff
@@ -0,0 +1,70 @@
...
+export function outcome<A extends Aspect, T extends Target>(
+  t: Assertions,
+  rule: Rule<A, T>,
+  aspects: AspectsFor<A>,
+  assert:
+    | Outcome.Inapplicable
+    | Readonly<
+        { [O in Outcome.Failed | Outcome.Passed | Outcome.CantTell]?: Array<T> }
+      >,
+  dependencies: Array<Rule<A, T>> = []
```
kasperisager: `Array` -> `ReadonlyArray`

packages/alfa-wcag/test/helpers/outcome.ts
```diff
@@ -0,0 +1,70 @@
...
+    const expected = outcome in assert ? assert[outcome]! : [];
+
+    t.equal(
+      actual.length,
+      expected.length,
+      `There must be ${expected.length} ${outcome} results`
+    );
+
+    for (const target of expected) {
+      if (target === null) {
```
kasperisager: This should not happen so try removing it and see if the code still types.

packages/alfa-wcag/test/helpers/outcome.ts
```diff
@@ -0,0 +1,70 @@
...
+        // Black magic will happen here
+      } else {
+        const holds = actual.some(
+          result =>
+            result.outcome === Outcome.Inapplicable || result.target === target
```
kasperisager: Inapplicable results should be handled by previous code so try removing the first condition and see if the code still types.

NiclasHedam: Unfortuntately, removing the Inapplicable part results in a compile-error.
```
Property 'target' does not exist on type 'Readonly<{ rule: string; outcome: Outcome.Inapplicable; }> | Readonly<{ rule: string; outcome: Outcome.Passed | Outcome.Failed; target: T; }>'.
  Property 'target' does not exist on type 'Readonly<{ rule: string; outcome: Outcome.Inapplicable; }>'
```

kasperisager: I'm guessing that's because this is not a proper type guard: https://github.com/Siteimprove/alfa/pull/82/files#diff-34397abfa06ec59f4dac72d55a3c25fcR47

NiclasHedam: Do you have any suggestions on a way to implement this type-guard?

kasperisager: Extracting the lambda and annotating it as a type guard should do the trick: https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

NiclasHedam: I tried exactly that, but I only managed to construct the inverse type-guard.

```ts
export function isInapplicable(
  input: Result
): input is Readonly<{ rule:string; outcome: Outcome.Inapplicable }> {
  return input.outcome === Outcome.Inapplicable;
}
``` 
works as expected, while 
```ts
export function isNotInapplicable(
  input: Result
): input is Readonly<{ rule:string; outcome: Outcome.Failed | Outcome.Passed; target: Target }> {
  return input.outcome === Outcome.Failed || input.outcome === Outcome.Passed;
}
```
does not change the end type at all.

kasperisager: Results are now generic in their outcome, which made the following type guard possible:

```ts
function hasMatchingOutcome(result: Result): result is Result<T, typeof outcome> {
  return result.outcome === outcome;
}
```

NiclasHedam: I don't think that would work, as Result only takes one argument (T).
```ts 
Result<T, typeof outcome>
```

kasperisager: f0437fb95c044686de2eeda94d9e6e9555e96cf7 😉

NiclasHedam: Right, thanks, I think it is good to go then

PR #84, opened 2018-11-08
Closes #83: Add tests for SIA-R2 through SIA-R18

### LINKED ISSUE #83: Sanshikan tests
After #73, Sanshikan rules are not considered testable and should have unit-tests.

NiclasHedam:
Please review #82 first, as this is a branch of that PR. If #82 is rejected, leave this as is until it is approved.

I also corrected a bug in SIA-R9


packages/alfa-wcag/test/sia-r11/rule.spec.tsx
```diff
@@ -7,23 +8,36 @@ import { SIA_R11 } from "../../src/sia-r11/rule";
...
+    width: 1920,
+    height: 1080,
+    orientation: Orientation.Landscape
+  },
+  display: {
+    resolution: 1,
+    scan: Scan.Progressive
+  }
+};
+
```
kasperisager: `getDefaultDevice()` was made for this purpose: https://github.com/Siteimprove/alfa/blob/master/packages/alfa-device/src/get-default-device.ts

NiclasHedam:
Since #82 is merged, this is ready for review!

packages/alfa-wcag/src/sia-r9/rule.ts
```diff
@@ -30,7 +30,12 @@ export const SIA_R9: Atomic.Rule<Document, Element> = {
     });
 
     expectations((target, expectation) => {
-      expectation(1, getRefreshTime(getAttribute(target, "content")!) === 0);
+      expectation(
+        1,
+        getRefreshTime(getAttribute(target, "content")!) === 0 ||
+          (getRefreshTime(getAttribute(target, "content")!) !== null &&
+            getRefreshTime(getAttribute(target, "content")!)! > 72000)
+      );
```
kasperisager: Assign `getRefreshTime(getAttribute(target, "content")!)` to a variable in order to avoid re-parsing `content` thrice. In this case, `getRefreshTime()` also cannot return `null` as that has already been checked in the applicability of the rule.

NiclasHedam:
This closed #83. I would not consider #83 closed yet, as there is still missing ten or so tests.

kasperisager:
That's a moving target 😉 Just create PRs for any future tests; no need for an associated issue as well.

Issue #77, opened 2018-10-25
Compilation sometimes fails during CI due to excessive memory consumption


NiclasHedam:
This issue looks like it is caused by Semaphore timing out or crashing.

NiclasHedam:
A solution could be not to compile the Unicode dataset each time; it could be something to do in a `version` hook.
Also, `packages/alfa-unicode/scripts/characters.js` seemingly has a hardcoded Unicode version number, so downloading and compiling the same dataset on each commit is just a waste of CPU cycles.

kasperisager:
The data set is only fetched when requested locally so the only time consuming task performed during CI is compilation. We however cannot naïvely skip compilation of select files even if they haven't changed as a change to any file within their import graph could affect type checking. 

The best, and safest, approach to speeding up compilation would therefore be to traverse the import graph of all files that have changed within a given commit range during CI and then compile only those files. Off the top of my head, I can't think of better ways to go about incremental compilation at least.

kasperisager:
Semaphore actually output an error message for this one build: https://semaphoreci.com/siteimprove/alfa/branches/todo-list-fix/builds/19. Not surprisingly, the cause is that the process runs out of memory.

NiclasHedam:
That's new! Good to know though.

kasperisager:
> The data set is only fetched when requested locally so the only time consuming task performed during CI is compilation. We however cannot naïvely skip compilation of select files even if they haven't changed as a change to any file within their import graph could affect type checking.
> 
> The best, and safest, approach to speeding up compilation would therefore be to traverse the import graph of all files that have changed within a given commit range during CI and then compile only those files. Off the top of my head, I can't think of better ways to go about incremental compilation at least.

kasperisager:
Fat-fingered the close button 🤦‍♀️

kasperisager:
ff80be7 should hopefully provide a solution to this but we'll see 🤞

PR #87, opened 2018-11-27
Caching Unicode and IANA subtags

NiclasHedam:
This PR is caching the Unicode and IANA database so that it does not have to be rebuilt over and over. It works by constructing the TypeScript source as usual, then compiling it and then removing the TypeScript source again. This process will yield the JavaScript version alongside the TypeScript definitions. Leaving the definitions assures that TypeScript understands the JavaScript correctly.

This brings down the build time about 45-90 seconds depending on the load of the host. Possibly also solves #77.


kasperisager:
While I do appreciate the suggestion, this is simply too much special casing for too little reward; the total CI time still approaches the 4 minute mark. https://github.com/Siteimprove/alfa/issues/77#issuecomment-435350660 outlines a generally applicable solution that would speed up compilation across the board. That, or something similar, would be worth pursuing.

kasperisager:
> For that reason, I cannot see why the other solution would improve performance at all compared to this.

If that's the case then you don't really understand what I'm suggesting. What I'm suggesting is that if we can infer that a file is in no way affected by a given change, we never touch it. That will benefit the entire code base and not just the special cases.

Build artifacts are not to be committed to version control, period 😉

PR #91, opened 2018-12-04
Check for missing spec-file when building files

Bugvi-Benjamin-M:
Added a missing spec file check when building files.


kasperisager:
Definitely needs a rebase on top of `master` 🙈

PR #95, opened 2018-12-11
Tests of SIA-R19 to SIA-R27

NiclasHedam:
Due to a bug in the type inference of R27, this pull request should not be considered complete.


NiclasHedam:
Is it possible that you can look into the type-bug before I do anything about it?

kasperisager:
Alles klar: https://github.com/Siteimprove/alfa/commit/f5471611f9749d1e16a32fd87bb3e83b00a75121

packages/alfa-aria/src/attributes.ts
```diff
@@ -90,6 +90,14 @@ export const DescribedBy: Attribute = {
...
 };
 
+/**
+ * @see https://www.w3.org/TR/wai-aria/#aria-describedat
+ */
+export const DescribedAt: Attribute = {
+  name: "aria-describedat",
+  type: "uri"
+};
+
```
kasperisager: This will need to be marked as obsolete and is not to be considered a valid ARIA attribute.

NiclasHedam:
:clap:

packages/alfa-wcag/test/sia-r27/rule.spec.tsx
```diff
@@ -0,0 +1,86 @@
...
+  );
+
+  const textAlternative = (
+    <p>
+      Not being able to use your computer because your mouse doesn't work, is
+      frustrating. Many people use only the keyboard to navigate websites.
+      Either through preference or circumstance. This is solved by keyboard
+      compatibility. Keyboard compatibility is described in WCAG. See the video
+      below to watch the same information again in video form.
+    </p>
```
kasperisager: Some bytes could be shaved off that fixture 🙈

PR #99, opened 2019-01-08
Moving format check from diagnose to prepare

NiclasHedam:
Currently, the code to check if a file has been correctly formatted resides in diagnose.js. While this suffices for code-files, it introduces a bug for other types of files. Prettier also formats JSON, CSS, HTML and other formats, and due to the setup of our pre-commit hook, any files supported by Prettier will be formatted. This leads to us formatting files such as tsconfig.json, but not validating it on the CI.

I have moved the format-check for prepare, and it checks ALL files for formatting issues. While this is a minor performance hit, I think it is necessary to ensure high-quality code.


NiclasHedam:
Now it is checking the entire source-tree with the exemption of dot-files and dot-directories. This is to not test cache files as well as git files. The check is split up and performed two different places for performance reasons.

scripts/prepare.js
```diff
@@ -2,10 +2,12 @@ const { findFiles } = require("./helpers/file-system");
 const { endsWith } = require("./helpers/predicates");
 const { Project } = require("./helpers/project");
 const { packages } = require("./helpers/meta");
-const { format, now } = require("./helpers/time");
+const { format: formatTime, now } = require("./helpers/time");
```
kasperisager:
```js
const time = require("./helpers/time");
```

scripts/prepare.js
```diff
@@ -21,19 +23,37 @@ const handle = (files, project) => {
...
+  if (process.env.CI === "true" && format(file)) {
+    notify.error(`${chalk.gray(file)} File has not been formatted`);
+    process.exit(1);
+  }
+};
+
+for (const file of findFiles(`.`, file => !file.startsWith("packages"))) {
+  checkFormat(file);
+}
+
```
kasperisager: Discussed in person.

scripts/helpers/file-system.js
```diff
@@ -70,12 +70,17 @@ exports.removeFile = removeFile;
 
 /**
  * @param {string | Iterable<string>} directories
- * @param {function(string): boolean} predicate
+ * @param {(function(string): boolean) | null} predicate
```
kasperisager:
```js
/** @param {(function(string): boolean)} [predicate]
```

NiclasHedam: That does not work:
`Type 'null' is not assignable to type '((arg0: string) => boolean) | undefined'.`

Issue #98, opened 2019-01-07
Investigate TypeScript builder program API


kasperisager:
Our current use of the TypeScript language service API in which individual source files are diagnosed and emitted has at least the following drawbacks:

- While diagnostics for source files that are _not_ our own are actually collected, we never show these anywhere. Only recently have we introduced support for traversing source file dependencies (7c94bb0add493815572c20dd5e809e1199ac0301) such that we _could_ improve upon this, but the next point will outline the problems with this as well.

- While we _can_ traverse source file dependencies by resolving imports, the TypeScript module resolver resolves imports from packages to their entry point defined in `package.json`, which is how things should work. What we're interested in, however, is which _actual_ source files a given source file depends on and not necessarily which source files it imports.

In effect, directly running the TypeScript compiler _can_ yield different results than when we compile the project using the language service API. 

By using the builder program API introduced in TypeScript 2.7, we _should_ be able to fix the drawbacks of our current approach. Specifically, the builder program API includes functionally for getting all dependencies of a file (https://github.com/Microsoft/TypeScript/blob/b7d7d5f7b39a5b9619c77590e5fe7f434ed68f1e/src/compiler/builder.ts#L651) in addition to providing incremental diagnostics for an entire program (https://github.com/Microsoft/TypeScript/blob/b7d7d5f7b39a5b9619c77590e5fe7f434ed68f1e/src/compiler/builder.ts#L660). How far these APIs will take us remains to be seen though.

kasperisager:
As the name does imply, `getAllDependencies()` (https://github.com/Microsoft/TypeScript/blob/b7d7d5f7b39a5b9619c77590e5fe7f434ed68f1e/src/compiler/builder.ts#L651) returns _all_ dependencies of a given file, direct _and_ transitive. I also haven't been able to make it behave as I would expect with regards to project references introduced in 2a3edb67ed4bd796755ad229665d402bf7381835; it resolves dependencies across packages to the implementation files, not declaration files. That is, if file `foo.ts` from package `foo` imports `bar.ts` from package `bar`, I would expect the dependency to resolve to `bar.d.ts` rather than `bar.ts`.

kasperisager:
I have some experimental code in place for this but will not pursue it further for now.

PR #100, opened 2019-01-08
Closes #97: Incremental builds

### LINKED ISSUE #97: Implement incremental builds
[`Graph`]: https://github.com/Siteimprove/alfa/blob/master/scripts/helpers/graph.js
[`Cache`]: https://github.com/Siteimprove/alfa/blob/master/scripts/helpers/cache.js
[`Project#resolveImports()`]: https://github.com/Siteimprove/alfa/blob/c4d268e81d96257e44675830405e9bb9ffd60bb5/scripts/helpers/project.js#L92-L128

In continuation of https://github.com/Siteimprove/alfa/issues/77#issuecomment-435350660, it would be great to have `yarn prepare` perform incremental builds either by default or based on a flag. We should by now have the necessary tools in place for implementing incremental builds and my plan for doing it is loosely as follows:

- Use [`Graph`][] and [`Project#resolveImports()`][] for incrementally constructing a dependency graph.
- Use [`Cache`][] to store the version number of every TypeScript source file. 
  - If the version does not change between builds and no dependencies of the file have changed, skip the file.
  - Otherwise, build the file, mark it as changed, update its dependencies in the [`Graph`][], and store its version in the [`Cache`][].

niclashedam:
I think I got incremental builds to work. When changing a file, all incoming dependencies are built as well.


NiclasHedam:
Can you give me a hint on what the next step is?

kasperisager:
Slating this until we're further along with #98.

PR #104, opened 2019-02-21
Better windows support

singingknight:
Limit when tsconfig.json is updated

Previously it was updated with windows relative path in files.
Previously the manifests wasn't formatted on update.

### DIFF
--- a/scripts/prepublish.js
+++ b/scripts/prepublish.js
@@ -42,6 +42,9 @@ for (const directory of packages) {
 }
 
 /**
+ * Update tsconfig.json in directory with
+ * - Add all .ts?(x) files to files
+ * - Add @siteimprove/ dependencies from package.json as relative path references
  * @param {string} directory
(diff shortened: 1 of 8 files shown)


NiclasHedam:
A known bug (character.js timeout) caused the build to fail. I have restarted the build.
Mind that the build will also fail due to the new "format"-check, but we can ignore that for now.

scripts/prepublish.js
```diff
@@ -62,7 +65,7 @@ function references(directory, pkg, options = {}, additional = []) {
       and(endsWith(".ts", ".tsx"), not(endsWith(".d.ts")))
     )
   ]
-    .map(file => path.relative(directory, file))
+    .map(file => path.relative(directory, file).replace(/\\/g, "/"))
```
kasperisager:
```ts
path.relative(directory, file).split(path.sep).join("/")
```

PR #105, opened 2019-02-22
Improve testability

singingknight:
Allow test for files in scripts folder

Make `yarn start`:
- test specification for changed file
- reprocess failed files on new update

### DIFF
--- a/scripts/start.js
+++ b/scripts/start.js
@@ -2,6 +2,7 @@ const { watchFiles } = require("./helpers/file-system");
 const { endsWith } = require("./helpers/predicates");
 const { format, now } = require("./helpers/time");
 const notify = require("./helpers/notify");
+const { getSpecification } = require("./helpers/typescript");
 
 const { build } = require("./tasks/build");
 const { diagnose } = require("./tasks/diagnose");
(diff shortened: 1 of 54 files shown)


scripts/start.js
```diff
@@ -2,38 +2,63 @@ const { watchFiles } = require("./helpers/file-system");
...
 
 const isSpec = endsWith(".spec.ts", ".spec.tsx");
 
+/**
+ * These are files which failed at last run
+ * Saved so we can process them again when next file changes
+ *
+ * @type {Array<string>}
+ */
+let failed = [];
```
kasperisager: While I do see the point of introducing something like this, I'm also a bit concerned of the impact it will have. I for one would very much dislike having my feedback stream littered with errors I _already know exist_ because I've seen them previously. If I make a change to file `A` and compilation fails because I also need to change something in file `B`, I don't want to be reminded that file `A` fails every time I write to file `B`. Once I'm finished with file `B` I'll head back and check for myself if file `A` still fails. In the event that I forgot to do so and it still fails, CI will remind me in the worst case.

kasperisager: To clarify, I'd actually love to be reminded about file `A`, I just don't want to see a bunch of errors unrelated to file `B` if that's the file I'm writing to.

PR #106, opened 2019-02-22
ADR 4: Model browser specific values and computations

### LINKED ISSUE #45: Write up ADR on browser specific values and computations

kasperisager:
Closes #45


docs/architecture/decisions/adr-004.md
```diff
@@ -0,0 +1,33 @@
...
+## Context
+
+Having decided on [ADR 2](adr-002.md), we will no longer be tied to any one specific browser and are left with the choice of what to do about things that browsers are not aligned on. For example, if Alfa needs to implement an API to perform a given task and a given set of browsers do not agree on how the API should be implemented, Alfa can do several things:
+
+1. Implement the API the way it is implemented in the _least_ capable browser, in terms of the API, among a given set of browsers.
+
+2. Implement the API the way it is implemented in the _most_ capable browser, in terms of the API, among a given set of browsers.
+
+3. Provide several implementations of the API that cover all the different implementations of the API among a given set of browsers.
+
```
singingknight: A fourth possibility which you could argue is the one used in the old engine is:

Implement the API the way it is implemented in the _most_ commonly used browser among a given set of browsers.

This may be the same as either 1 or 2, but might be a third way.

NiclasHedam: I think talking about _common_ in the sense of browsers is very subjective and should be avoided. When is something _common enough_?

kasperisager: That's a good observation, though it is equivalent to option 1 and 2 when the set of browsers is in fact the most commonly used browser under some measure of usage.

docs/architecture/decisions/adr-003.md
```diff
@@ -2,7 +2,7 @@
...
+Having decided on [ADR 2](adr-002.md), we foresee that the Alfa code base will be both significantly larger and more complex than the code base of our proprietary engine. This due to the fact that we will have to implement a great deal of APIs that we have previously relied on the browser implementations of. Coupled with the fact that the most common type of bug we have encountered in the past has been stray `undefined` or `null` values and APIs receiving incorrect parameters, plain JavaScript, even if covered by tests, is simply not an option moving forward. We will either need tooling that can sanity check our JavaScript or move to a language with a proper type system that can enforce API contracts.
```
Bugvi-Benjamin-M: > This due to the fact 

Would prefer "This **is** due to the fact". Flows better.

docs/architecture/decisions/adr-004.md
```diff
@@ -0,0 +1,33 @@
+[accessibility support baseline]: https://www.w3.org/TR/WCAG-EM/#step1c
+
```
Bugvi-Benjamin-M: > if Alfa needs to implement an API to perform a given task and a given

if Alfa needs to implement an API to perform a given task, and a given

docs/architecture/decisions/adr-004.md
```diff
@@ -0,0 +1,33 @@
...
+# ADR 4: Model browser specific values and computations
+
+## Context
+
+Having decided on [ADR 2](adr-002.md), we will no longer be tied to any one specific browser and are left with the choice of what to do about things that browsers are not aligned on. For example, if Alfa needs to implement an API to perform a given task and a given set of browsers do not agree on how the API should be implemented, Alfa can do several things:
+
+1. Implement the API the way it is implemented in the _least_ capable browser, in terms of the API, among a given set of browsers.
```
Bugvi-Benjamin-M: Might just be me, but is it clear enough what is meant by "least capable" and "most capable"? 

Should this perhaps be defined somewhere? Or is it?

docs/architecture/decisions/adr-004.md
```diff
@@ -0,0 +1,33 @@
...
+
+2. Implement the API the way it is implemented in the _most_ capable browser, in terms of the API, among a given set of browsers.
+
+3. Provide several implementations of the API that cover all the different implementations of the API among a given set of browsers.
+
+Our proprietary accessibility conformance testing engine at Siteimprove takes the second approach in practice as it is usually run within the latest version of the Chrome browser and relies heavily on browser APIs. Other tools that operate with the concept of an [accessibility support baseline][] will typically take a combination of the first and second approach where some browser APIs are used and then others implemented manually according to the capabilities of the [accessibility support baseline][].
```
Bugvi-Benjamin-M: @singingknight said in the comments that 

> A fourth possibility which you could argue is the one used in the old engine is:
Implement the API the way it is implemented in the _most_ commonly used browser among a given set of browsers.
This may be the same as either 1 or 2, but might be a third way.

Just wondering about inconsistency here. We should all agree on how Alfa relates to the proprietary engine.

docs/architecture/decisions/adr-004.md
```diff
@@ -0,0 +1,33 @@
...
+
+The quality shared by the first and second approach is that they both produce just a single implementation of the API, though with potentially different capabilities. However, both approaches suffer from the fact that depending on how the API is implemented, an accessibility issue could be found in either the least or most capable browser, while being missed in the other. As an example, a contrast issue might appear in an older browser that does not support a specific color format, while the contrast would be perfectly fine in a newer browser that does support the specific color format. The third approach avoids this issue entirely by implementing the API in all possible ways for a given set of browsers.
```
Bugvi-Benjamin-M: > As an example,

Would prefer "E.g.,"

Just being "på tværs"

Bugvi-Benjamin-M: > a contrast issue might appear in an older browser that does not support a specific color format, while the contrast would be perfectly fine in a newer browser

I am assuming that older and newer here refer to "least capable" and "most capable"?

kasperisager:
Will address non-critical changes separately.

PR #116, opened 2019-04-16
Closes #111: Implements overflow property

### LINKED ISSUE #111: Implement `overflow` CSS property
Among other things, this property is needed to heuristically check if a node is visible or not.

NiclasHedam:


NiclasHedam:
I have implemented the longhands as per the spec, but I am struggling with how the shorthands should work. Any input?

kasperisager:
While longhands return a single value for a CSS property, shorthands return an object will multiple values for several CSS properties. You can find an example here: https://github.com/Siteimprove/alfa/blob/master/packages/alfa-css/src/properties/font/property.ts

packages/alfa-css/src/properties/overflow/property.ts
```diff
@@ -0,0 +1,39 @@
...
+        overflowY = result;
+        offset = position;
+      }
+    }
+
+    return {
+      overflowX,
+      overflowY
+    };
+  }
```
kasperisager: Getting there! Take a quick look at this section of the CSS syntax spec though: https://www.w3.org/TR/css-syntax/#error-handling. Specifically:

> When interpreting a list of declarations, unknown syntax at any point causes the parser to throw away whatever declaration it’s currently building, and seek forward until it finds a semicolon (or the end of the block). It then starts fresh, trying to parse a declaration again.

NiclasHedam:
I think we are good now.

packages/alfa-css/src/properties/overflow/property-y.ts
```diff
@@ -0,0 +1,38 @@
...
+  },
+  computed(style) {
+    const value = getSpecifiedProperty(style, "overflowY");
+
+    switch (value) {
+      case keyword("visible"):
+        return keyword("auto");
+      case keyword("clip"):
+        return keyword("hidden");
+    }
```
kasperisager: Comparisons of objects in JavaScript are based on reference equality, not structural equality. As `keyword()` returns an object, the above `switch` effectively has no effect.

packages/alfa-css/src/properties/overflow/property-y.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    return getSpecifiedProperty(style, "overflowY");
```
kasperisager: Use `value`.

packages/alfa-css/src/properties/overflow/property-x.ts
```diff
@@ -0,0 +1,38 @@
...
+  },
+  computed(style) {
+    const value = getSpecifiedProperty(style, "overflowX");
+
+    switch (value) {
+      case keyword("visible"):
+        return keyword("auto");
+      case keyword("clip"):
+        return keyword("hidden");
+    }
```
kasperisager: Comparisons of objects in JavaScript are based on reference equality, not structural equality. As `keyword()` returns an object, the above `switch` effectively has no effect.

packages/alfa-css/src/properties/overflow/property-x.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    return getSpecifiedProperty(style, "overflowX");
```
kasperisager: Use `value`.

packages/alfa-css/test/properties/overflow/property.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+});
+
+test("Cannot parse invalid full shorthand", t => {
+  const lexer = lex("foo bar", Alphabet);
+  t.deepEqual(overflow.parse(lexer.result), {
+    overflowX: overflowX.initial(),
+    overflowY: overflowY.initial()
+  });
+});
```
kasperisager: Make a short helper for lexing and parsing things similar to this: https://github.com/Siteimprove/alfa/pull/116/files#diff-361044a82987e2ea39051466b46d8dbdR10

packages/alfa-css/src/properties/overflow/property-x.ts
```diff
@@ -0,0 +1,43 @@
...
+  },
+  initial() {
+    return keyword("visible");
+  },
+  computed(style) {
+    const value = getSpecifiedProperty(style, "overflowX");
+
+    if (
+      getSpecifiedProperty(style, "overflowY").value != "visible" &&
+      getSpecifiedProperty(style, "overflowY").value != "clip"
```
kasperisager: Store in variable.

packages/alfa-css/src/properties/overflow/property.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @see https://drafts.csswg.org/css-overflow-3/#propdef-overflow
+ */
+export const overflow: Shorthand<"overflowX" | "overflowY"> = {
+  longhands: ["overflowX", "overflowY"],
+  parse(tokens) {
+    let offset = 0;
+    let overflowX: Overflow = Longhands.overflowX.initial();
+    let overflowY: Overflow = Longhands.overflowX.initial();
```
kasperisager: There should be no need to initialise `overflowX` and `overflowY` here.

PR #129, opened 2019-06-25
Closes #112: Implement ARIA role mappings for SVG

### LINKED ISSUE #112: Implement semantic roles from WAI-ARIA Graphics Module
https://www.w3.org/TR/graphics-aria/

Bugvi-Benjamin-M:
This is the first version. There are some issues that need to be handled at a later stage.


packages/alfa-aria/src/features/svg/audio.ts
```diff
@@ -0,0 +1,10 @@
+import * as Roles from "../../roles";
+import { Feature } from "../../types";
+
+/**
+ * @see https://html.spec.whatwg.org/multipage/canvas.html#canvas
```
NiclasHedam: This shouldn't be canvas, should it?

packages/alfa-aria/src/features/svg/symbol.ts
```diff
@@ -0,0 +1,13 @@
+import * as Roles from "../../roles";
+import { Any, Feature } from "../../types";
+
+/**
+ * @see http://www.w3.org/TR/SVG2/struct.html#SymbolElement
+ */
+export const Switch: Feature = {
+  element: "switch",
```
NiclasHedam: Shouldn't this be `symbol`?

packages/alfa-aria/src/get-role.ts
```diff
@@ -18,7 +18,10 @@ for (const role of values(Roles)) {
 
 const featuresByElement = new Map<string, Feature>();
 
-for (const feature of values(Features)) {
+for (const feature of values(Features.HTML)) {
+  featuresByElement.set(feature.element, feature);
+}
+for (const feature of values(Features.SVG)) {
```
kasperisager: Features will need to be distinguished on namespace as SVG features that overlap with HTML features will otherwise overwrite these.

packages/alfa-aria/src/features/svg/animate-motion.ts
```diff
@@ -0,0 +1,10 @@
+import * as Roles from "../../roles";
+import { Feature, None } from "../../types";
+
+/**
+ * @see https://svgwg.org/specs/animations/#AnimateMotionElement
+ */
+export const AnimateMotion: Feature = {
+  element: "animatemotion",
```
kasperisager: Element names in SVG are reported as defined in the SVG spec, so `<animateMotion>` would have a `localName` of `animateMotion` rather than `animatemotion`.

packages/alfa-aria/src/features/svg/a.ts
```diff
@@ -0,0 +1,25 @@
+import { Element, getClosest, hasAttribute, Node } from "@siteimprove/alfa-dom";
+import * as Roles from "../../roles";
+import { Any, Feature, Role } from "../../types";
+
+/**
+ * @see https://www.w3.org/TR/SVG2/linking.html#AElement
```
kasperisager: We need to reference the place that defines the ARIA role mapping for each element. Unfortunately, I think the closest we can get is https://www.w3.org/TR/svg-aam/#mapping_role_table.

PR #130, opened 2019-06-25
Video flow diagram

Jym77:


Jym77:
Yes, I agree that this is not really maintainable.
I am, however, also pretty sure that automatic tools will do a very bad job out of laying out that many nodes in a graph. It’s probably possible to obtain a similar result in Mermaid but having the graph flow in many directions seems to require quite a lot of manual hacking (invisible nodes and subgraphs) that completely destroy the automation (and maintainability…)

Maybe this means that such a complete flow graph is something we only produce once in a while and which is not meant to reside withing the github repo but rather as internal doc in Confluence or such (of course, the source should also be shared, not just the pdf).

OTOH, checks of individual atomic rules tend to have very simple workflows (essentially linear with some early escapes). These are likely well suited for an automatic tool and would provide a nice graphic doc for the code. Here, the complexity of the graph comes from trying to merge identical nodes (same test), rather than having 9 parallel flows.

PR #131, opened 2019-06-26
Closes #57: Adding support for AnB Microsyntax

### LINKED ISSUE #57: Implement support for An+B microsyntax
Pseudo-class selectors currently only support values that are either selectors or lists of selectors:

https://github.com/Siteimprove/alfa/blob/27e4ccc3a9da94ab936715b12034fc23ddd74648/packages/alfa-css/src/types.ts#L109

However, some functional pseudo-class selectors make use of the An+B microsyntax (https://www.w3.org/TR/css-syntax/#anb), for example `:nth-child(2n+1)`. To support these kinds of selectors, we first need to implement support for this syntax within the selector parser. The lexer can already lex them:

https://github.com/Siteimprove/alfa/blob/27e4ccc3a9da94ab936715b12034fc23ddd74648/packages/alfa-css/test/alphabet.spec.ts#L335-L349

niclashedam:


packages/alfa-lang/src/parse.ts
```diff
@@ -11,13 +11,17 @@ export interface ParseResult<R> {
...
 ): ParseResult<R> {
-  const readToken: (i: number) => T = i => input[i];
+  const readToken: (i: number) => T = i =>
+    input instanceof Stream ? input.peek(i)! : input[i];
 
-  const stream = new Stream(input.length, readToken, offset);
+  const stream =
+    input instanceof Stream
+      ? input
+      : new Stream(input.length, readToken, offset);
```
kasperisager: Should be:

```ts
let stream: Stream<T>;

if (input instanceof Stream) {
  stream = input;
} else {
  const readToken: (i: number) => T = i => input[i];

  stream = new Stream(input.length, readToken, offset);
}
```

PR #142, opened 2019-09-17
Refactor element checks (name, input type, namespace, role)

Jym77:


packages/alfa-rules/src/helpers/element-checker.ts
```diff
@@ -0,0 +1,142 @@
...
+  public withInputType(...inputTypes: Array<InputType>): ElementChecker {
+    this.inputTypeSet = new Set(inputTypes);
+    return this;
+  }
+
+  public build(): (node: Node) => boolean | BrowserSpecific<boolean> {
+    return node => this.evaluate(node);
+  }
+
+  public evaluate(node: Node): boolean | BrowserSpecific<boolean> {
```
Jym77: 1. The topmost `evaluate` always return `boolean` but the inherited one can return a Browser Specific one. I'd like to have a way to say "this can be both because of inheritance, but if calling from this class, it's always `boolean`" but it seems not really possible. This ripples into a bunch of cast (`as boolean`) when using the toplevel method…
2. I'd also love to be able to refine the type (from `node` to `element`) here, but that does not seem possible with browser specific booleans…

packages/alfa-rules/src/helpers/element-checker.ts
```diff
@@ -0,0 +1,142 @@
...
+      device,
+      new Set(roles),
+      this.namesSet,
+      this.inputTypeSet,
+      this.namespacesSet
+    );
+  }
+
+  public evaluate(node: Node): boolean | BrowserSpecific<boolean> {
+    if (!isElement(node) || !super.evaluate(node)) {
```
Jym77: Because `super.evaluate` does not narrow type, I'm forced to recheck `isElement` here 😞

packages/alfa-rules/src/helpers/element-checker.ts
```diff
@@ -0,0 +1,142 @@
...
+      emptyOrHas(elementNamespace, this.namespacesSet)
+    );
+  }
+}
+
+class BrowserSpecificElementChecker extends ElementCheckerWithContext {
+  private readonly device: Device;
+  private readonly rolesSet: Set<Option<Role>>;
+
+  constructor(
```
Jym77: That's typically the place where having to create a new object when getting to the subclass is annoying. I don't want to carry over all those variables, I want to keep the value they already have…

packages/alfa-rules/src/sia-r1/rule.ts
```diff
@@ -51,10 +51,10 @@ function hasDocumentElement(document: Document): boolean {
...
-    return false;
-  }
-
-  return element.localName === "title";
+function isTitle(node: Node, context: Node): node is Element {
+  return new ElementChecker()
+    .withName("title")
+    .withContext(context)
+    .withNamespace(Namespace.HTML)
+    .evaluate(node) as boolean;
```
Jym77: This is typically the cast I'd like to avoid. The toplevel checker never return browser specific values…

packages/alfa-rules/test/sia-r12/rule.spec.tsx
```diff
@@ -43,3 +43,15 @@ test("Is inapplicable when element is not in the accessibility tree", t => {
     Outcome.Inapplicable
   );
 });
+
+test("Is inapplicable when element is not a button", t => {
```
Jym77: Unrelated additional test for increasing coverage.

packages/alfa-rules/src/helpers/element-checker.ts
```diff
@@ -0,0 +1,142 @@
...
+} from "@siteimprove/alfa-dom";
+import { Option } from "@siteimprove/alfa-util";
+// import { Predicate } from "@siteimprove/alfa-util";
+
+function emptyOrHas<T>(elt: T, set?: Set<T>) {
+  return set === undefined || set.has(elt);
+}
+
+export class ElementChecker {
+  constructor(namesSet?: Set<string>, inputTypeSet?: Set<InputType>) {
```
Jym77: Having the extra variables here is only used for the class inheritance, making this a bit hacky. Normal use is the `with*` methods, even if initialisation could override this…

packages/alfa-rules/src/helpers/element-checker.ts
```diff
@@ -0,0 +1,142 @@
...
+    return new ElementCheckerWithContext(
+      context,
+      this.namesSet,
+      this.inputTypeSet
+    );
+  }
+
+  protected namesSet?: Set<string>;
+  public withName(...names: Array<string>): ElementChecker {
+    this.namesSet = new Set(names);
```
Jym77: Storing the test in instance variables is annoying since it essentially require a new instance for every different test, and even for every "non-constant" test (_e.g._ using context).

packages/alfa-rules/src/helpers/predicate-builder.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  public and<V extends U = U>(
+    predicate: Predicate<U, V>
+  ): PredicateBuilder<T, V> {
+    return new PredicateBuilder(v => this.predicate(v) && predicate(v));
+  }
+
+  public browserSpecific(): BrowserSpecificPredicateBuilder<T, U> {
+    return new BrowserSpecificPredicateBuilder(this.predicate);
+  }
```
Jym77: This is ugly 😞 
`and` should be overloaded with `predicate: BrowserSpecificPredicate<U, V>` and handle this. But it does not look like I can test the type between `Predicate` and `BSPredicate` to choose what to do…

packages/alfa-rules/src/helpers/predicate-builder.ts
```diff
@@ -0,0 +1,160 @@
...
+      BrowserSpecific.map(getRole(element, context, device), elementRole =>
+        roles.some(role => role === elementRole)
+      )
+    );
+  }
+}
+
+type BrowserSpecificPredicate<T, U extends T = T> = (
+  t: T
+) => boolean | BrowserSpecific<boolean>; // t is U
```
Jym77: The second type variable is useful for typing the `BSPredicateBuilder` methods.

packages/alfa-rules/src/helpers/predicate-builder.ts
```diff
@@ -0,0 +1,160 @@
...
+
+class BrowserSpecificPredicateBuilder<T, U extends T = T> {
```
Jym77: This should somehow inherit from `PredicateBuilder` because now `withName` and such must be used before `withRole` and duplicating them would be bad. Not sure this is really possible 😞

packages/alfa-rules/src/helpers/predicate-builder.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  public constructor(predicate: BrowserSpecificPredicate<T, U> = () => true) {
+    this.predicate = predicate;
+  }
+
+  public and<V extends U = U>(
+    predicate: BrowserSpecificPredicate<U, V>
+  ): BrowserSpecificPredicateBuilder<T, V> {
+    return new BrowserSpecificPredicateBuilder((t: T) =>
+      and(this.predicate(t), predicate(t as U))
```
Jym77: And that's where I'd love to be able to define infix operator or even better to overload `&&` 😢

Jym77: The cast is needed to keep the type structure happy. It is safe-ish since `this.predicate(t)` is (browser-specifically) narrowing the type, and lazy boolean expression should ensure we don't reach the second part without it being true…

packages/alfa-rules/src/sia-r11/rule.ts
```diff
@@ -41,14 +34,13 @@ export const SIA_R11: Atomic.Rule<Device | Document, Element> = {
...
-              }
-
-              return isExposed(element, document, device);
-            });
+          filter(querySelectorAll(document, document, isElement()), element => {
+            return isElement(builder =>
+              builder
+                .withNamespace(document, Namespace.HTML)
+                .withRole(device, document, Roles.Link)
+                .and(element => isExposed(element, document, device))
```
Jym77: `isExposed` could obviously be added as a predicate builder method. Maybe another time…

packages/alfa-rules/src/sia-r11/rule.ts
```diff
@@ -41,14 +34,13 @@ export const SIA_R11: Atomic.Rule<Device | Document, Element> = {
...
+            )(element);
```
Jym77: Thus said, this whole construction is rather nice.

packages/alfa-rules/src/sia-r11/rule.ts
```diff
@@ -41,15 +34,15 @@ export const SIA_R11: Atomic.Rule<Device | Document, Element> = {
...
-            });
-          }),
+          filter(
+            querySelectorAll(document, document, isElement()),
+            isElement(builder =>
+              builder
+                .withNamespace(document, Namespace.HTML)
+                .withRole(device, document, Roles.Link)
+                .and(element => isExposed(element, document, device))
+            )
```
Jym77: Thus said, this is a rather nice way of expressing the condition 😃

packages/alfa-rules/src/helpers/predicate-builder.ts
```diff
@@ -0,0 +1,160 @@
...
+  factory?: (builder: ElementPredicateBuilder) => ElementPredicateBuilder<T>
+): Predicate<Node, T>;
+
+export function isElement(
+  factory: (
+    builder: ElementPredicateBuilder
+  ) => BrowserSpecificPredicateBuilder<Node>
+): BrowserSpecificPredicate<Node>;
+
+export function isElement(
```
Jym77: The name conflict with the guard is a bit annoying…

packages/alfa-rules/src/sia-r2/rule.ts
```diff
@@ -37,22 +34,22 @@ export const SIA_R2: Atomic.Rule<Device | Document, Element> = {
...
+            isElement(builder =>
+              builder
+                .withNamespace(document, Namespace.HTML)
+                .browserSpecific()
+                .and(
+                  isElement(builder =>
+                    builder
+                      .withRole(device, document, Roles.Img)
+                      .and(element => isExposed(element, document, device))
+                      .or(isElement(builder => builder.withName("img")))
```
Jym77: This is a bit ugly. Especially the implicit left-associativity of the boolean operations…

Jym77:
> I'm thinking the next steps would be for us to figure out a way to keep the feel of the current API, but implement it in a way that does not require all these class extensions and method overrides and whatnot. Some inspiration might be found in [RxJS](https://rxjs-dev.firebaseapp.com/), which used to implement method chaining for operations on [observables](https://rxjs-dev.firebaseapp.com/api/index/class/Observable) similar to what is implemented for predicates in your pull request. Now, it uses a single `.pipe()` method that is then used for chaining together operations instead. If one wishes to add a new operation, it's therefore just a matter of implementing a function that can be passed to `.pipe()`; no need to add class methods and overrides.

That's a bit the direction that is taken with the `.and` and `.or` methods. We could go further and have an API like `isElement(builder => builder.and(nameIs("div")).and(namespaceIs(context, Namespace.HTML)))`. That still leaves weird associativity issues which I assume the `.pipe` does not have because it mimicks actual piping (while boolean operations have different associativity rules). Obviously, going fully that way a nice API would be `isElement((nameIs("div") or inputTypeIs(IjnputType.Radio)) and namespaceIs(context, Namespace.HTML))` but we're hitting the "no infix" limit (and the same thing with prefix and/or would be not so nice).

I'm wondering whether passing tests as an option object would be a nice API: `isElement( {names: ["div"]; namespace: { context: document; namespaces: [Namespace.HTML] }} )` that would have an implicit "and" and any "or" must be explicitly made.

Jym77:
@kasperisager I tried to improve the API usage (with what is already here…) in the test file https://github.com/Siteimprove/alfa/blob/is-element-by-name/packages/alfa-rules/test/helpers/predicate-builder.spec.tsx

That should solve the "duplicate inheritance" and "extending inheritance" issue, but not the associativity issue.
Is that what we should use?

I also tried to get a "options" API (1 example, at the end of the test file). Not sure what is best.

kasperisager:
I'll take a peek 👍

kasperisager:
@Jym77 I've put the new `Predicate` types in the `@siteimprove/alfa-compatibility` package for now, but I'm actually thinking that introducing a new package might be better. How do you feel about `@siteimprove/alfa-logic`?

Jym77:
> @Jym77 I've put the new `Predicate` types in the `@siteimprove/alfa-compatibility` package for now, but I'm actually thinking that introducing a new package might be better. How do you feel about `@siteimprove/alfa-logic`?

That sounds OK.

kasperisager:
I'll leave that for some other time; let's get this on `master`.

PR #148, opened 2019-10-09
Update lang rules

Jym77:
Update lang rules as per:
- https://github.com/act-rules/act-rules.github.io/pull/822
- https://github.com/act-rules/act-rules.github.io/pull/934


packages/alfa-rules/src/sia-r5/rule.ts
```diff
@@ -24,10 +25,11 @@ export const SIA_R5: Atomic.Rule<Document, Attribute> = {
...
             document,
             document,
-            node =>
-              isElement(node) &&
-              isDocumentElement(node, document) &&
-              hasLanguageAttribute(node)
+            Predicate.from(
+              isElement
+                .and(element => isDocumentElement(element, document))
+                .and(element => hasAttribute(element, "lang"))
```
Jym77: I suggest that in some time we `grep .and(element =>` and do a review on which functions could use parameter swapping + curryfication to be used more easily in the Predicate builder.

PR #152, opened 2019-10-18
Fix SIA-R42

Jym77:
Changes authored by @singingknight 

I don't know why he cannot push directly to the repo…


kasperisager:
Gah, those global DOM types keep haunting us 🤦‍♂ Good catch! As for lack of direct push access, only @Siteimprove/alfa currently has write access; everyone else must use forks.

PR #162, opened 2019-12-05
SIA-R59: Document has headings

Jym77:
It literally took me 10 minutes to write the rule 😀 …
… because I had to dig a bit in the code to find the correct name and syntax of some helper functions 🤣


packages/alfa-rules/src/sia-r59/rule.ts
```diff
@@ -0,0 +1,50 @@
...
+              hasNamespace(equals(Namespace.HTML)),
+              hasRole(hasName(equals("heading")))
+            )
+          )
+        );
+
+        return {
+          1: isEmpty(headings)
+            ? Err.of("The document does not have a heading element")
+            : Ok.of("The document has at least one heading element")
```
kasperisager: Rather than do `isEmpty(filter(..))` you can simply do `some(...)` and then switch around `Err` and `Ok`.

Jym77: 🤔 but `some` is expecting a predicate to test on the iterable and has no default predicate.
This means I have to do `some(heading, isElement)` or something similar which is actually less readable I think (and feels a bit redundant).

=> Should `some` have a default predicate (always true) for this use case?

kasperisager: To clarify, you'd do this:

```ts
const hasHeadings = some(
  target.descendants({ flattened: true }),
  and(
    Element.isElement,
    and(
      hasNamespace(equals(Namespace.HTML)),
      hasRole(hasName(equals("heading")))
    )
  )
)
```

Jym77: 🤦‍♂ 
Of course…

Issue #92, opened 2018-12-05
Parallelize package builds across available cores


kasperisager:
With the changes introduced in  ff80be7, packages no longer use a shared workspace when performing full builds. As such, we now have the option to parallelize package builds across the available cores of the host system. By using the [cluster module](https://nodejs.org/api/cluster.html), we can create a number of process forks, each of which will handle the compilation of packages handed to them by the master process. 

Do be aware of dependencies between packages! Hint: https://cs.stackexchange.com/questions/2524/getting-parallel-items-in-dependency-resolution/2525#2525

Bugvi-BM:
If you will let me, I would love to have a look at this. If only to give a shot, fail miserably and have a few laughs.

kasperisager:
Go for it! 👍

kasperisager:
This may or may not be impacted by #98.

kasperisager:
Putting this on hold for now.

kasperisager:
In light of #170 I'll close this.

kasperisager:
This is now part of the TypeScript 3.9 iteration plan: https://github.com/Microsoft/TypeScript/issues/30235

PR #170, opened 2020-01-23
Rework the build system

### LINKED ISSUE #97: Implement incremental builds

kasperisager:
This pull request does away with most of the current build scripts in favour of instead relying on the new solution builder introduced in TypeScript 3.6 (https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-6.html#apis-to-support---build-and---incremental). The end result is that doing a `yarn build` is now no different than doing a `tsc --build`. So... why even have these build scripts, then? Time for a bit of history!

Back when we started work on Alfa, the code base was obviously fairly small. We could compile all files using `tsc` and then subsequently lint them using `tslint` fairly fast. As the code base grew, "fairly fast" soon turned into "fairly slow". Back then, the best option for regaining some speed was to rely on the TypeScript language service using a custom language service host that could additionally share a document registry between the TypeScript compiler and TSLint:

https://github.com/Siteimprove/alfa/blob/98ee8bc8da39bbfad100e10d53681394be14f534/scripts/helpers/project.js#L7-L239

This made for very efficient incremental compiles of individual files when run as a long-lived process. However, full builds were still slow and were becoming slower over time. This lead to #92 and #97, which now a year later have still not been solved.

Since their introduction, the current build scripts have been extended with quite a few features that made them increasingly complex. They still haven't really solved the issue I originally set out to solve though; to make builds _fast_.

By instead relying on the solution builder provided in TypeScript 3.6, we get fast builds for free as TypeScript does all the heavy lifting for us. The only additional functionality provided on top is a way of building and running our tests.

In summary, the new set of build scripts perform the following tasks:

- **`build.js`:** Compile TypeScript sources to JavaScript.
- **`clean.js`:** Remove compiled artifacts.
- **`test.js`:** Compile TypeScript sources to JavaScript and run all resulting `*.spec.js` files.
- **`watch.js`:** Watch TypeScript sources and compile to JavaScript on changes.

All of these can be scoped to work on individual packages, and their dependencies of course, by supplying a path to the package: `build.js packages/alfa-dom`.

Do note that the `watch.js` script has changed quite a bit compared to the previous version, which would also re-run `*.spec.js` files on changes. I have not personally missed this functionality given the ease of simply doing a `test.js [package-i-am-working-on]`. Especially with the incremental builds provided by TypeScript, which means that if I _also_ run `watch.js`, `test.js` only needs to run the tests as it can skip compilation entirely.

Other functionality that was previously available, such as syncing fields of `package.json` files and automatically adding project references and TypeScript sources to `tsconfig.json` files, has not been implemented _yet_. I'd like to figure out what our actual **needs** are and then go from there.

Closes #97.


kasperisager:
Also worth pointing out:

`rework-build-system`

```console
$ loc scripts
--------------------------------------------------------------------------------
 Language             Files        Lines        Blank      Comment         Code
--------------------------------------------------------------------------------
 JavaScript              11          115           21           12           82
--------------------------------------------------------------------------------
 Total                   11          115           21           12           82
--------------------------------------------------------------------------------
```

`master`

```console
$ loc scripts
--------------------------------------------------------------------------------
 Language             Files        Lines        Blank      Comment         Code
--------------------------------------------------------------------------------
 JavaScript              27         2098          394          582         1122
 JSON                     1           38            0            0           38
--------------------------------------------------------------------------------
 Total                   28         2136          394          582         1160
--------------------------------------------------------------------------------
```

If we can cover _most_ of our build system needs with an order of magnitude less code then that's a major win in my view.

kasperisager:
It can indeed be made a bit more verbose; simply pass along the `--verbose` flag 👌

PR #174, opened 2020-01-24
Three valued logic

Jym77:
Crude trilean logic, on one rule. To see if it's OK.


packages/alfa-rules/src/sia-r38/rule.ts
```diff
@@ -17,13 +22,45 @@ export default Rule.Composite.of<Page, Element, Question>({
...
+  predicate: PredicateTrilean<T>
+): PredicateTrilean<Iterable<T>> {
+  return iterable =>
+    Iterable.reduce(
+      iterable,
+      (pred: PredicateTrilean<void>, val) =>
+        PredicateTrilean.or(pred, () => predicate(val)),
+      () => false
+    )();
+}
```
Jym77: This should totally move but where? same place as everyTrilean.

packages/alfa-act/src/oracle.ts
```diff
@@ -4,7 +4,7 @@ import { Option } from "@siteimprove/alfa-option";
...
 import { Rule } from "./rule";
 
-export type Oracle<Q> = <I, T, A>(
-  rule: Rule<I, T, Q>,
-  question: { [K in keyof Q]: Question<K, Q[K], T, A> }[keyof Q]
-) => Future<Option<A>>;
+export type Oracle<Q> = <I, S, T>(
+  rule: Rule<I, S, Q>,
+  question: { [K in keyof Q]: Question<K, Q[K], S, T> }[keyof Q]
+) => Future<Option<T>>;
```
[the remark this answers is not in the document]
kasperisager: Now they don't match `Rule` though 🙈 `I` is for `Input`, `T` is for `Target`, and `A` is for `Answer`.

packages/alfa-rules/test/common/make-oracle.ts
```diff
@@ -0,0 +1,28 @@
...
+  return (rule, question) => {
+    const answer = answers[question.uri];
+    if (question.type === "boolean" && isBoolean(answer))
+      return Future.now(Option.of(question.answer(answer)));
+    if (question.type === "node" && isOption(answer))
+      return Future.now(Option.of(question.answer(answer)));
+    if (question.type === "node[]" && isIterable(answer))
+      return Future.now(Option.of(question.answer(answer)));
+
+    return Future.now(None);
```
kasperisager: You can wrap the entire thing in `Future.delay(() => { ... })`.

packages/alfa-predicate-trilean/src/predicate-trilean.ts
```diff
@@ -0,0 +1,106 @@
+import { Equatable } from "@siteimprove/alfa-equatable";
+import { Mapper } from "@siteimprove/alfa-mapper";
+
+export type Trilean = boolean | undefined;
+
+export type PredicateTrilean<T, A extends Array<unknown> = Array<unknown>> = (
+  value: T,
+  ...args: A
+) => Trilean;
```
kasperisager: This is already within a package named "trilean" so there's no need to repeat that in the name of the type. I'd simply make two modules within this package: `trilean.ts`, which defines the `Trilean` type, and `predicate.ts`, which defines the `Predicate` type. If consumers need to disambiguate between boolean predicates and trilean predicates, they can easily do so via import aliases.

packages/alfa-result/src/noresult.ts
```diff
@@ -0,0 +1,98 @@
+import { None } from "@siteimprove/alfa-option";
```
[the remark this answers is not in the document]
kasperisager: It however mixes concerns that should not be mixed; `Result<T, E>` and `Option<T>` have well-defined semantics and this muddies those semantics. The use-case is trivially achieved using `Option<Result<T, E>>`.

packages/alfa-predicate-trilean/package.json
```diff
@@ -0,0 +1,31 @@
+{
+  "private": true,
+  "name": "@siteimprove/alfa-predicate-trilean",
```
kasperisager: Let's just name this `@siteimprove/alfa-trilean`.

packages/alfa-act/src/oracle.ts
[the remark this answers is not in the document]
Jym77: I guess, maybe Question should also have "targets" rather than "subjects" 😄 
(I spend a lot of time going between these type definitions trying to understand them and the mismatching type variable really bothered me… Another possibility could be to use longer type variable name (Target, Subject, …) instead of single letter ones, maybe full upper case (TARGET, SUBJECT, …) if we want to distinguish them visually)

packages/alfa-rules/src/common/expectations/audio-or-video-text-alternative.ts
```diff
@@ -0,0 +1,99 @@
+import { Question as ActQuestion } from "@siteimprove/alfa-act";
```
Jym77: @kasperisager Do you have an opinion about that refactor?
At the very least, it shows that we'r doing the exact same test (but different applicability) on these cases.

(affected rules: 26, 29, 31)

I am not entirely convinced that this is good. Notably trying to keep both the audio and video tests together. Refactoring only the video code used in 2 rules makes a lot of sense, I'm less sure about adding the audio part in it (even though it **is** the same code…)

kasperisager: If we refer to the combination of "audio" and "video" as "media", then I think this might work a little better.

Jym77: Yes!

PR #181, opened 2020-02-04
Duplicate is streaming

### LINKED ISSUE #166: Separate questions for audio and video live-streaming
Currently using the same question for both audio and video. It would be more user-friendly to be specific.

Is the video live-streaming?
Is the audio live-streaming?

Jym77:
closes #166


kasperisager:
@Jym77 Is this something we're ready to merge soon?

Jym77:
Probably next sprint.
We don't want to merge this before backend is ready for it and because it is the first question update, this need some work for DB updates… Planned for next sprint…

Issue #188, opened 2020-02-12
Replace CSS `Function` value type with more specific types


kasperisager:
The CSS functional notation is slowly moving to a format that is not easily represented using a shared value type in Alfa: https://wiki.csswg.org/ideas/functional-notation. For example, the `rgb()` functional notation for colors is now `rgb(20 54 10 / 90%)` rather than `rgb(20, 54, 10, 90%)` though the latter is still supported for compatibility reasons.

PR #189, opened 2020-02-13
Provide additional CSS value types

### LINKED ISSUE #168: Implement CSS transform parser
One of the things that didn't make it through #165 was the parser for the CSS `transform` property. This will need to be reimplemented for SIA-R44 to behave as expected:

https://github.com/Siteimprove/alfa/blob/98ee8bc8da39bbfad100e10d53681394be14f534/packages/alfa-css/src/properties/transform/grammar.ts#L12-L359

kasperisager:
This pull request provides additional CSS value types that aim to solve #188. As it also introduces value types for representing colors, it's needed for #187 as well.

The goal of _all_ the CSS value types used in Alfa is to allow for a generic representation of CSS property values in a way that allows us to represent both the values specified by an author in addition to values computed by the style system. To this end, I was inspired by the way that CSS values are represented in Servo (https://github.com/servo/servo/tree/master/components/style/values), in particular its representation of generic values. As an example, consider the `translate()` value; when specified by an author, it accepts arguments that are either lengths or percentages, but when computed these arguments are resolved to absolute lengths with the canonical unit `px`. As such, the CSS value type representing the `translate()` value is generic with regards to the type of its arguments:

https://github.com/Siteimprove/alfa/blob/bb0b31045b1babc1004ad6e51187277d17c5b94e/packages/alfa-css/src/value/transform/translate.ts#L9-L13

This way, what an author specifies would result in just `Translate` with whatever defaults it has, but it would compute to `Translate<Length<"px"> | Percentage, Length<"px"> | Percentage, Length<"px">>`.

Closes #168.


packages/alfa-css/src/value/color/hsl.ts
```diff
@@ -0,0 +1,194 @@
...
+    return Number.of(0);
+  }
+
+  public get green(): Number {
+    return Number.of(0);
+  }
+
+  public get blue(): Number {
+    return Number.of(0);
+  }
```
kasperisager: I still need to implement these conversions.

packages/alfa-css/src/value/gradient.ts
```diff
@@ -0,0 +1,224 @@
...
+ */
+export type Gradient = Linear;
+
+export namespace Gradient {
+  export type JSON = Linear.JSON;
+
+  /**
+   * @see https://drafts.csswg.org/css-images/#color-stop
+   */
+  export class Stop<
```
Jym77: Should Stop and Hint be moved to separate files in a subfolder? Like it's done with colours, for example.

packages/alfa-css/src/value/transform.ts
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * @see https://drafts.csswg.org/css-transforms/#typedef-transform-function
+   */
+  export const parse = either(
+    Matrix.parse,
+    either(
+      Perspective.parse,
+      either(
+        Rotate.parse,
+        either(Scale.parse, either(Skew.parse, Translate.parse))
```
Jym77: There seems to be a lot of chaining `either` in this PR. Would it be worth to have a n-ary `either(...args)`?

kasperisager: We unfortunately cannot type those currently as TypeScript lacks variadic kinds. I'm crossing my fingers for the pipeline operator to arrive though 😄

packages/alfa-css/src/value/gradient.ts
kasperisager: The reason for them to live in the `Gradient` namespace is that they're not really useful on their own as they are only used within the definitions of the individual gradient kinds.

kasperisager:
I hope to get around to adding tests after the color contrast rule has been implemented. I'll however get this stuff merged now as I really need it 🙈

Issue #171, opened 2020-01-23
Set up ESLint configuration


kasperisager:
As part of #165, TSLint was removed as it is being deprecated in favour of ESLint, see https://medium.com/palantir/tslint-in-2019-1a144c2317a9. If we want to continue with a linter setup, which I'm very much in favour of, we therefore need to set up an ESLint configuration.

kasperisager:
I've just attempted to run ESLint against Alfa with the following configuration, which essentially does nothing:

```json
{
  "$schema": "http://json.schemastore.org/eslintrc",
  "root": true,
  "parser": "@typescript-eslint/parser",
  "parserOptions": {
    "sourceType": "module",
    "project": ["tsconfig.json", "packages/*/tsconfig.json"]
  },
  "plugins": ["@typescript-eslint"]
}
```

Even with no rules enabled, Node.js ends up running out of memory. I had to increase the maximum allowed heap space to 4 GB to make things work, which seems excessive. Additionally, removing the Unicode database and the IANA subtag registry (~200K generated LoC) had no effect on this meaning that ESLint cannot even chew its way through our actual ~26K LoC.

I'm investigating further.

kasperisager:
Turns out that we're very much affected by https://github.com/typescript-eslint/typescript-eslint/issues/1192 as we have 70+ TypeScript projects within the Alfa monorepo. The solution was a separate TypeScript configuration that combines _all_ sources into a single project:

```jsonc
{
  "$schema": "http://json.schemastore.org/tsconfig",
  "extends": "./tsconfig.json",
  // ESLint needs a single TypeScript configuration that includes _all_ files
  // to be linted.
  // https://github.com/typescript-eslint/typescript-eslint/issues/1192
  "include": ["packages/**/*.ts", "packages/**/*.tsx"]
}
```

kasperisager:
https://github.com/Siteimprove/alfa/pull/179#issuecomment-901873506

PR #194, opened 2020-03-03
make HTML attribute names case insensitive

Jym77:


packages/alfa-dom/src/node/attribute.ts
```diff
@@ -1,8 +1,10 @@
-import { None, Option } from "@siteimprove/alfa-option";
+import {None, Option} from "@siteimprove/alfa-option";
+import {Predicate} from "@siteimprove/alfa-predicate";
 
 import { Namespace } from "../namespace";
 import { Node } from "../node";
 import { Element } from "./element";
+import equals = Predicate.equals;
```
kasperisager:
```suggestion

const { equals } = Predicate;
```

packages/alfa-dom/src/node/attribute.ts
```diff
@@ -46,7 +48,15 @@ export class Attribute extends Node {
   }
 
   public get name(): string {
-    return this._name;
+    // HTML attributes are case insensitive. Other, e.g. SVG, are case sensitive.
+    return this._namespace.some(equals(Namespace.HTML)) ? this._name.toLowerCase() : this._name;
+  }
+
+  public matchName(name: string): boolean {
```
kasperisager:
```suggestion
  public hasName(name: string): boolean {
```

This should also be placed after the getters. We should enforce the ordering using https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-ordering.md once ESLint is up and running (#171).

kasperisager:
On a more important note, HTML attributes don't have namespaces so the check should actually be `this._namespace.isNone()`. A safer option could also be to look at the namespace of the owning element.

packages/alfa-dom/src/node/attribute.ts
```diff
@@ -1,8 +1,11 @@
-import { None, Option } from "@siteimprove/alfa-option";
+import {None, Option} from "@siteimprove/alfa-option";
+import {Predicate} from "@siteimprove/alfa-predicate";
 
-import { Namespace } from "../namespace";
-import { Node } from "../node";
-import { Element } from "./element";
+import {Namespace} from "../namespace";
+import {Node} from "../node";
+import {Element} from "./element";
```
kasperisager: Something odd happened to the formatting here 🙈

Jym77:
Yeah, it's anyway better to push the computation to the constructor rather than the getter…

Problem is that the constructor can't really call `this.isHTML` (especially if its using `this.owner`). Or at least this sounds like a recipe for disaster… Since `hasName` is still needed (to lowercase the match), this results in some smelly duplication of the test…

packages/alfa-dom/src/node/attribute.ts
```diff
@@ -57,6 +61,13 @@ export class Attribute extends Node {
     return this._owner;
   }
 
+  public hasName(name: string): boolean {
+    // HTML attributes are case insensitive. Other, e.g. SVG, are case sensitive.
+    return this.owner.some(element => element.namespace.some(equals(Namespace.HTML))) ?
+      this._name.toLowerCase() === name.toLowerCase() :
+      this._name === name;
```
kasperisager:
```suggestion
    // HTML attributes are case insensitive. Other, e.g. SVG, are case sensitive.
    return this._name ===
      this._owner.some(element => element.namespace.some(equals(Namespace.HTML)))
      ? name.toLowerCase()
      : name;
```

packages/alfa-dom/src/node/attribute.ts
```diff
@@ -32,7 +35,8 @@ export class Attribute extends Node {
 
     this._namespace = namespace;
     this._prefix = prefix;
-    this._name = name;
+    // HTML attributes are case insensitive. Other, e.g. SVG, are case sensitive.
```
kasperisager:
```suggestion
    // HTML attributes are case insensitive while attributes in other namespaces aren't.
```

packages/alfa-dom/src/node/attribute.ts
```diff
@@ -57,6 +61,13 @@ export class Attribute extends Node {
     return this._owner;
   }
 
+  public hasName(name: string): boolean {
+    // HTML attributes are case insensitive. Other, e.g. SVG, are case sensitive.
```
kasperisager:
```suggestion
    // HTML attributes are case insensitive while attributes in other namespaces aren't.
```

kasperisager:
@Jym77 I'm not too worried about this bit of duplication and do much prefer repeating the same stanza twice over introducing a private method 👍 It's a single line of code anyway 😄

PR #197, opened 2020-03-13
Cell header association

Jym77:
Should be OK now.


packages/alfa-dom/src/tables/structure.ts
```diff
@@ -0,0 +1,353 @@
...
+}
+
+function setSlot(table: Table, col: number, row: number, slot: Slot) {
+  if (table.slots[col] === undefined) table.slots[col] = [];
+  table.slots[col][row] = slot;
+}
+
+export function newTable(): Table {
+  return { slots: [[]], width: 0, height: 0, cells: [], rowGroups: [], colGroups: []}
+}
```
Jym77: Obviously needs to be a class.

packages/alfa-dom/src/tables/structure.ts
```diff
@@ -0,0 +1,353 @@
...
+
+// Bad global variables! Bad!
+const global = { yCurrent:0};
```
Jym77: Either move to table class or have a specific "table under construction" class?

packages/alfa-dom/src/tables/structure.ts
```diff
@@ -0,0 +1,353 @@
...
+
+// https://html.spec.whatwg.org/multipage/tables.html#concept-cell
+export type Cell = {
+  kind: "data" | "header";
+  // "top left" corner of the cell
+  anchor: { x: number; y: number };
+  // size of the cell
+  width: number;
+  height: number;
+  growing: boolean; // true if part of the growing cells list.
```
Jym77: Maybe the growing cell list would be better handled as part of the "table under construction" class?

packages/alfa-dom/src/tables/structure.ts
```diff
@@ -0,0 +1,353 @@
...
+export function parseSpan(element: Element, name: string, min: number, max: number, failed: number): number {
+  return element.attribute(name)
+    .map(parseAttribute(parseNonNegativeInteger))
+    .map(r => r.map(x => clamp(x, min, max)))
+    .getOr(Ok.of(failed))
+    .getOr(failed);
+}
+
+// https://html.spec.whatwg.org/multipage/tables.html#algorithm-for-growing-downward-growing-cells
+function growingCell(yCurrent: number, keepGrowing: boolean = false): ((cell: Cell) => void) {
```
Jym77: Needs to update cell (rather than create a new one) because the cell is shared between the table and the slots. Not super fan of it, I’d rather have this as a pure function and repalce the `forEach` with `map` where needed…

packages/alfa-dom/src/tables/structure.ts
```diff
@@ -0,0 +1,353 @@
...
+      // The col element represents column within the colgroup but is not a colgroup itself. The rest of the algorithm seems to never use that again…
+      // const colGroup: ColGroup = { anchor: {x: global.theTable.width - span}, width: span, element: currentCol}; // need better name! Technically not a "column group"…
+      // global.theTable.colGroups.push(colGroup);
+    }
+    // 4 and 7 done in main function
+    return { anchor: { x: xStart }, width: totalSpan, element: colgroup };
+  }
+}
+
+export function formingTable(element: Element): Table {
```
Jym77: Should return a `Result` for cleaner handling of errors.

packages/alfa-dom/test/tables/structure.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+    .getOr("");
+
+
+const cleanSlot = (slot: Slot) => (
+  {...slot,
+    elements: slot
+      .elements
+      .map(cleanElement)
+  }
+);
```
Jym77: Pure debug helpers to `console.dir` the table without showing the full elements…

packages/alfa-dom/src/tables/structure.ts
kasperisager: I'd encapsulate all the state in an object that is passed around between the different functions.

packages/alfa-dom/src/tables/structure.ts
kasperisager: Agreed 👍

packages/alfa-dom/src/tables/cell.ts
```diff
@@ -0,0 +1,703 @@
...
+        this._anchorY + this._height - 1 < y
+      ) // slot is below cell
+    );
+  }
+
+  /**
+   * compare cell according to their anchor
+   * in a given group of cells (row, rowgroup, table, …), no two different cells can have the same anchor, so this is good.
+   */
+  public compare(cell: Cell): number {
```
kasperisager: I've added a `Comparable<T>` interface that this class should implement: https://github.com/Siteimprove/alfa/blob/master/packages/alfa-comparable/src/comparable.ts.

packages/alfa-dom/src/tables/cell.ts
```diff
@@ -0,0 +1,703 @@
...
+    anchor: { x: number; y: number };
+    width: number;
+    height: number;
+    element: string; //Element.JSON;
+    headers: string[]; //Element.JSON[];
+  }
+
+  export enum Kind {
+    Header = "HEADER",
+    Data = "DATA",
```
kasperisager:
```suggestion
    Header = "header",
    Data = "data",
```

All string constants in Alfa are kebab cased.

packages/alfa-dom/src/tables/cell.ts
```diff
@@ -0,0 +1,703 @@
...
+  }
+}
+
+export class BuildingCell implements Equatable, Serializable {
```
kasperisager: Let's put this in the `Cell` namespace and simply call it `Builder`.

packages/alfa-dom/src/tables/cell.ts
```diff
@@ -0,0 +1,703 @@
...
+    };
+  }
+}
+
+export namespace Cell {
+  export interface JSON {
+    [key: string]: json.JSON;
+
+    kind: Kind;
+    anchor: { x: number; y: number };
```
kasperisager:
```suggestion
    x: number; 
    y: number;
```

packages/alfa-dom/src/tables/cell.ts
```diff
@@ -0,0 +1,703 @@
...
+      this._anchorX === value._anchorX &&
+      this._anchorY === value._anchorY &&
+      this._element.equals(value._element)
+    );
+  }
+
+  public toJSON(): Cell.JSON {
+    return {
+      kind: this._kind,
+      anchor: this.anchor,
```
kasperisager:
```suggestion
      x: this._x,
      y: this._y,
```

packages/alfa-dom/src/tables/cell.ts
```diff
@@ -0,0 +1,703 @@
...
+  }
+
+  public equals(value: unknown): value is this {
+    return (
+      value instanceof Cell &&
+      this._kind === value._kind &&
+      this._width === value._width &&
+      this._height === value._height &&
+      this._anchorX === value._anchorX &&
+      this._anchorY === value._anchorY &&
```
kasperisager:
```suggestion
      this._x === value._x &&
      this._y === value._y &&
```

packages/alfa-dom/src/tables/cell.ts
```diff
@@ -0,0 +1,703 @@
...
+
+  /**
+   * compare cell according to their anchor
+   * in a given group of cells (row, rowgroup, table, …), no two different cells can have the same anchor, so this is good.
+   */
+  public compare(cell: Cell): number {
+    if (this._anchorY < cell._anchorY) return -1;
+    if (this._anchorY > cell._anchorY) return 1;
+    if (this._anchorX < cell._anchorX) return -1;
+    if (this._anchorX > cell._anchorX) return 1;
```
kasperisager:
```suggestion
    if (this._y < cell._y) return -1;
    if (this._y > cell._y) return 1;
    if (this._x < cell._x) return -1;
    if (this._x > cell._x) return 1;
```

packages/alfa-dom/src/tables/cell.ts
```diff
@@ -0,0 +1,703 @@
...
+  }
+
+  public isCovering(x: number, y: number): boolean {
+    return !(
+      // cell is *not* covering if either
+      (
+        x < this._anchorX || // slot is left of cell
+        this._anchorX + this._width - 1 < x || // slot is right of cell
+        y < this._anchorY || // slot is above cell
+        this._anchorY + this._height - 1 < y
```
kasperisager:
```suggestion
        x < this._x || // slot is left of cell
        this._x + this._width - 1 < x || // slot is right of cell
        y < this._y || // slot is above cell
        this._y + this._height - 1 < y
```

packages/alfa-comparable/src/comparable.ts
```diff
@@ -1,9 +1,15 @@
+export enum Comparison {
+  Smaller = -1,
```
kasperisager:
```suggestion
  Less = -1,
```

Great thinking introducing an enum for this though, I dig it! 💯

packages/alfa-dom/src/tables/cell.ts
```diff
@@ -0,0 +1,730 @@
...
+    element: string; //Element.JSON;
+    headers: string[]; //Element.JSON[];
+  }
+
+  export enum Kind {
+    Header = "header",
+    Data = "data",
+  }
+
+  export class Building
```
kasperisager:
```suggestion
  export class Builder
```

This ensures alignment with the builder pattern (https://en.wikipedia.org/wiki/Builder_pattern), which we also employ in other packages.

packages/alfa-dom/src/tables/cell.ts
```diff
@@ -0,0 +1,730 @@
...
+
+    // debug only
+    public get name(): string {
+      return this._cell.element.attribute("id").get().value;
+    }
+
+    // end debug
+
+    public get anchor(): { x: number; y: number } {
+      return { x: this._cell.anchor.x, y: this._cell.anchor.y };
```
kasperisager:
```suggestion
      return this._cell.anchor;
```

packages/alfa-dom/src/tables/colgroup.ts
```diff
@@ -0,0 +1,175 @@
+import { Comparable, Comparison } from "@siteimprove/alfa-comparable";
+import { Equatable } from "@siteimprove/alfa-equatable";
+import * as json from "@siteimprove/alfa-json";
+import { Serializable } from "@siteimprove/alfa-json";
+import { Err, Ok, Result } from "@siteimprove/alfa-result";
```
kasperisager:
```suggestion
import { Comparable, Comparison } from "@siteimprove/alfa-comparable";
import { Equatable } from "@siteimprove/alfa-equatable";
import { Serializable } from "@siteimprove/alfa-json";
import { Err, Ok, Result } from "@siteimprove/alfa-result";

import * as json from "@siteimprove/alfa-json";
```

packages/alfa-dom/src/tables/colgroup.ts
```diff
@@ -0,0 +1,175 @@
...
+          // 5
+          // The col element represents column within the colgroup but is not a colgroup itself. The rest of the algorithm seems to never use that again…
+          // const colGroup: ColGroup = { anchor: {x: global.theTable.width - span}, width: span, element: currentCol}; // need better name! Technically not a "column group"…
+          // global.theTable.colGroups.push(colGroup);
+        }
+      }
+      // 1.4 and 1.7 done in main function
+      // 2.2 and 2.3 done in main function
+      return Ok.of(Building.of(x, totalSpan, colgroup));
+    }
```
kasperisager: This can be moved to the associated namespace instead.

packages/alfa-dom/src/tables/row.ts
```diff
@@ -0,0 +1,289 @@
...
+          // 2 is done when creating the row, default value for xCurrent is 0.
+          Building.of(yCurrent, w, 1, tr, [], growingCells)
+            // 3
+            ._growCells(yCurrent)
+        )
+      );
+
+      // return row;
+      // 4 and 16 done after the calls to avoid side effects.
+    }
```
kasperisager: This can be moved to the associated namespace instead.

packages/alfa-dom/src/tables/row.ts
```diff
@@ -0,0 +1,289 @@
...
+    public equals(value: unknown): value is this {
+      if (!(value instanceof Building)) return false;
+      const sortedThisCells = this._cells.sort((a, b) => a.compare(b));
+      const sortedValueCells = value._cells.sort((a, b) => a.compare(b));
+      const sortedThisDGCells = this._downwardGrowingCells.sort((a, b) =>
+        a.compare(b)
+      );
+      const sortedValueDGCells = value._downwardGrowingCells.sort((a, b) =>
+        a.compare(b)
+      );
```
kasperisager: `Array#sort()` sorts in place so there's a) no need to use the return value and b) no need to do it more than once during construction.

packages/alfa-dom/src/tables/rowgroup.ts
```diff
@@ -0,0 +1,254 @@
+import { Comparable, Comparison } from "@siteimprove/alfa-comparable";
+import { Equatable } from "@siteimprove/alfa-equatable";
+import * as json from "@siteimprove/alfa-json";
+import { Serializable } from "@siteimprove/alfa-json";
+import { Err, Ok, Result } from "@siteimprove/alfa-result";
```
kasperisager:
```suggestion
import { Comparable, Comparison } from "@siteimprove/alfa-comparable";
import { Equatable } from "@siteimprove/alfa-equatable";
import { Serializable } from "@siteimprove/alfa-json";
import { Err, Ok, Result } from "@siteimprove/alfa-result";

import * as json from "@siteimprove/alfa-json";
```

packages/alfa-dom/src/tables/rowgroup.ts
```diff
@@ -0,0 +1,254 @@
...
+      );
+      // ending row group 2
+      // When emptying the growing cells list, we need to finally add them to the group.
+      rowgroup = rowgroup._update({
+        cells: rowgroup._cells.concat(growingCellsList),
+      });
+      // 3, returning the row group for the table to handle
+      // we could check here if height>0 and return an option, to be closer to the algorithm but that would be less uniform.
+      return Ok.of(rowgroup);
+    }
```
kasperisager: This can be moved to the associated namespace.

Jym77:
I tried to reorg stuff. Things are clearer now. Still have 2 outstanding questions I'm not sure how to handle…

---
https://github.com/Siteimprove/alfa/blob/cell-header-association/packages/alfa-dom/src/common/predicate.ts is clearly not in its place, but I don't know where to put it. It's common predicates that are used both in `alfa-rules` and in `alfa-table`. Thus, there is no real natural place for them. 

Should we have some specific collection of shared predicates (which could make more sense than in `alfa-rules/src/common/predicates`)?
Can't really be in `alfa-predicate` as these shared predicates depends on other packages… An `alfa-shared-predicate` package seems a bit overkilled…

Ideas?

---
I'm not super happy with what I ended up doing with the microsyntaxes (now split between `alfa-dom` and `alfa-table` depending where they are used but with some export still creeping out…) The splitting (and somewhat burrowing down) means we're going to forget them and repeat them…

Would you think an `alfa-microsyntaxes` package make sense? That's currently not used in many places, but could be…

packages/alfa-dom/src/common/predicate.ts
```diff
@@ -0,0 +1,32 @@
...
+  predicate: Predicate<Namespace> = () => true
+): Predicate<Element> {
+  return (element) => element.namespace.some(predicate);
+}
+
+export function hasName<T extends { readonly name: string }>(
+  predicate: Predicate<string> = () => true
+): Predicate<T> {
+  return property("name", predicate);
+}
```
kasperisager, line 16: This predicate is not related to DOM as it's just a wrapper around `property("name", predicate)` so let's not put it in the DOM package. I'm thinking `Element#hasName(predicate: Predicate<string>): boolean` and `Element.hasName(predicate: Predicate<string>): Predicate<Element>` would make sense though. This should also remove the need for `isElementByName()`.

packages/alfa-dom/src/common/microsyntaxes.ts
```diff
@@ -0,0 +1,45 @@
...
+  str: string
+): Ok<readonly [string, Array<string>]> {
+  return Ok.of([
+    "",
+    str
+      .trim()
+      .split(/\s+/)
+      .filter((s) => s !== ""),
+  ] as const);
+}
```
kasperisager, line 45: This would be well-suited as `Attribute#tokens(separator: string | RegExp = /\s+/): Array<string>`. This could also be used for parsing comma-separated tokens by doing `attribute.tokens(",")`.

packages/alfa-dom/src/common/microsyntaxes.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    return str === ""
+      ? Err.of(EnumeratedValueError.Missing)
+      : result.isNone()
+      ? Err.of(EnumeratedValueError.Invalid)
+      : Ok.of(["", result.get()] as const);
+  }
+
+  return parser;
+}
```
kasperisager, line 30: This would be well-suited as `Attribute#enumerate(): Option<string>`. The current API is a little unfortunate in that it requires a very specific type of mapping (specifically a `Map<K, V>`) to be given up front, which isn't the only conceivable thing we would want to do with an enumerated attribute. In general, we should aim to design APIs that make as few assumptions as possible in order to allow callers the most flexibility.

With the above API, the specific use case you have could then be supported by `attribute.enumerate().flatMap(keyword => mapping.get(keyword))` or even this:

```ts
attribute.enumerate().flatMap(keyword => {
  switch (keyword) {
    case "foo":
    case "bar":
      return Some.of(keyword);
    default:
      return None;
  }
});
```

`#enumerate()` would handle all the repetitiveness of lowercasing the attribute value and returning `None` when it's empty; the caller then knows that they receive a keyword that is ready to map to a state or do whatever else the caller wants to do.

packages/alfa-dom/src/node/element.ts
```diff
@@ -347,6 +369,32 @@ export namespace Element {
...
+          Err.of(EnumeratedValueError.Missing)
+        )
+        .mapOrElse(
+          ([_, result]) => Some.of(result),
+          (err) => mapping.get(err)
+        );
+    }
+
+    return parser;
+  }
```
kasperisager, line 397: With `Attribute#enumerate()` I don't see a need for this function.

packages/alfa-dom/src/common/predicate.ts
```diff
@@ -0,0 +1,32 @@
+import { Predicate } from "@siteimprove/alfa-predicate";
+import {Element, Namespace, Node} from "..";
+
+const { and, equals, property } = Predicate;
+
+export function hasNamespace(
+  predicate: Predicate<Namespace> = () => true
+): Predicate<Element> {
+  return (element) => element.namespace.some(predicate);
+}
```
kasperisager, line 10: This would be well-suited as `Element#hasNamespace(predicate: Predicate<Namespace>): boolean` and `Element.hasNamespace(predicate: Predicate<Namespace>): Predicate<Element>`.

packages/alfa-dom/src/common/predicate.ts
```diff
@@ -0,0 +1,32 @@
...
+    and(hasNamespace(equals(Namespace.HTML)), hasName(equals(...names)))
+  );
+}
+
+export function isDescendantOf(
+  node: Node,
+  options?: Node.Traversal
+): Predicate<Node> {
+  return (desc) => node.descendants(options).find(equals(desc)).isSome();
+}
```
kasperisager, line 32: This would be well-suited as `Node#includes(descendant: Node, options?: Node.Traversal): boolean` and `Node.includes(descendant: Node, options?: Node.Traversal): Predicate<Node>`.

kasperisager: Seeing how this actually looks for the inverse relation, I'd suggest both `hasDescendant()` and `hasAncestor()` as well.

Jym77:
I moved code around into classes/namespaces where it belongs.

I'm not super happy with what I did with `hasName`. The code is now duplicated in `alfa-dom/element` and `alfa-aria/role` because it is used on both elements and roles. Which is bad.

Given that it is used both in `alfa-rules` and `alfa-table`, there is no natural package to hold it as a local helper only (like it was before in `alfa-rules/predicate`.

I don't know how to do this better, to avoid the code duplication but keep it at a natural place.

packages/alfa-aria/src/role.ts
```diff
@@ -108,6 +108,10 @@ export class Role<N extends string = string> implements Equatable {
     return some(this.characteristics.name.from, predicate);
   }
 
+  public hasName(predicate: Predicate<N>): boolean {
+    return predicate(this.name);
+  }
```
kasperisager, line 113:
```suggestion
  public hasName(predicate: Predicate<string>): boolean {
    return predicate(this.name);
  }
```

packages/alfa-aria/src/role.ts
```diff
@@ -281,6 +285,30 @@ export namespace Role {
...
+
+  export function hasName<N extends string = string>(
+    name: string,
+    ...rest: Array<string>
+  ): Predicate<Role<N>>;
+
+  export function hasName<N extends string = string>(
+    nameOrPredicate: string | Predicate<string>,
+    ...names: Array<string>
+  ): Predicate<Role<N>> {
```
kasperisager, line 301:
```suggestion
  export function hasName(
    predicate: Predicate<string>
  ): Predicate<Role>;

  export function hasName(
    name: string,
    ...rest: Array<string>
  ): Predicate<Role>;

  export function hasName(
    nameOrPredicate: string | Predicate<string>,
    ...names: Array<string>
  ): Predicate<Role> {
```

kasperisager:
> I'm not super happy with what I did with `hasName`. The code is now duplicated in `alfa-dom/element` and `alfa-aria/role` because it is used on both elements and roles. Which is bad.

I don't think that's bad at all; many things have names after all 😄 Avoiding duplication is a worthy pursuit up to a point and I find this to be beyond that point. So, all is well 👍

packages/alfa-dom/src/node/element.ts
```diff
@@ -262,6 +262,47 @@ export class Element extends Node implements Slot, Slotable {
...
+
+    if (attribute.isNone() || attribute.get().value === "")
+      return Element.EnumeratedAttributeError.Missing;
+
+    const value = attribute.get().value.toLowerCase();
+
+    return keywords.includes(value)
+      ? value
+      : Element.EnumeratedAttributeError.Invalid;
+  }
```
kasperisager, line 296: As mentioned in https://github.com/Siteimprove/alfa/pull/197#discussion_r408673970, this belongs in `Attribute` as also made evident by the fact that the method is now postfixed with, well, `Attribute` 🙈 I do like this way of passing a predefined set of valid keywords though, so I'm thinking an overload would make sense:

```ts
Attribute#enumerate(): Option<string>;
Attribute#enumerate(valid: string, ...rest: Array<string>): Option<string>;
```

Jym77: The problem I had when trying to put it on `Attribute` is that it cannot handle the case where the attribute is missing, so the full `if (attribute.isNone())` part must still be done at the `Element` level. Either requiring a specific helper (like this one), or the need to systematically chain:
`element.attribute("foo").map(attribute => attribute.enumerate()).getOr(Err.of("missing"))`
instead of 
`element.enumerateAttribute("foo")`

Might still be better to have the helper split in two, however…

kasperisager: The chaining approach is indeed what should be used; that's the whole reason for the monadic approach:

```ts
const foo: string = element
  .attribute("foo")
  .map(attribute => 
    attribute.enumerate("bar", "baz").getOr("invalid")
  )
  .getOr("missing");
```

If I don't care about missing and invalid states, it's as simple as:

```ts
const foo: Option<string> = element
  .attribute("foo")
  .flatMap(attribute => attribute.enumerate("bar", "baz"));
```

If I only care about missing states, I can instead do:

```ts
const foo: Option<string> = element
  .attribute("foo")
  .map(attribute => attribute.enumerate("bar", "baz"))
  .getOr(Some.of("missing"));
```

By not making too many assumptions about what the caller wants to do, we've now made it possible to model a lot more things than if we had only designed the API with a single use case in mind.

packages/alfa-dom/src/node/attribute.ts
```diff
@@ -93,6 +96,29 @@ export class Attribute extends Node {
...
+  ): Result<string, Attribute.EnumeratedAttributeError> {
+    if (this._value === "")
+      return Err.of(Attribute.EnumeratedAttributeError.Missing);
+
+    const value = this._value.toLowerCase();
+
+    return keywords.length === 0 || keywords.includes(value)
+      ? Ok.of(value)
+      : Err.of(Attribute.EnumeratedAttributeError.Invalid);
+  }
```
kasperisager, line 120:
```suggestion
  /**
   * @see https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute
   */
  public enumerate(): Option<string>;

  /**
   * @see https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute
   */
  public enumerate(valid: string, ...rest: Array<string>): Option<string>;
  
  public enumerate(
    ...valid: Array<string>
  ): Option<string> {
    const value = this._value.toLowerCase();

    return valid.length === 0 || valid.includes(value)
      ? Some.of(value)
      : None;
  }
```

A return value of `None` will always mean "invalid"; it's only when the attribute _isn't_ specified that the "missing" state is relevant. If it's specified with an empty value then it's still considered specified. As it turns out, the empty string can actually be a valid keyword, as noted in the spec, so we shouldn't return `None` when the attribute value is empty.

Jym77: I had somehow missed the fact that empty string can be a keyword 🤦 
This indeed makes things much easier with each error coming from a single and separate source…

packages/alfa-comparable/src/comparable.ts
```diff
@@ -1,9 +1,26 @@
+
```
kasperisager, line 1:
```suggestion
```

A rogue newline 🙈

packages/alfa-comparable/src/comparable.ts
```diff
@@ -1,9 +1,26 @@
...
+    More
+  }
+
+  /**
+   * Ready made generic function to use with Array.sort
+   */
+  export function compare<T extends Comparable<U>, U>(a: T, b: U): Comparison {
+    return a.compare(b);
+  }
+
```
kasperisager, line 25:
```suggestion
```

A rogue newline 🙈

packages/alfa-dom/src/node/attribute.ts
```diff
@@ -93,6 +95,26 @@ export class Attribute extends Node {
...
+  public tokens(separator: string | RegExp = /\s+/): Array<string> {
+    return this._value.trim().split(separator).filter(not(isEmpty));
+  }
+
+  /**
+   * @see https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute
+   */
+  public enumerate(): Option<string>;
+
+  public enumerate<VALID extends string = string>(valid: VALID, ...rest: Array<VALID>): Option<VALID>;
```
kasperisager, line 110:
```suggestion
  /**
   * @see https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#enumerated-attribute
   */
  public enumerate<T extends string>(valid: T, ...rest: Array<T>): Option<T>;
```

A type variable is always declared using a single upper-case letter, with `T` being preferred as the type variable of highest importance, such as `Result<T, E>` and `Option<T>`.

The docblock should be included in both overloads to ensure that it always shows up in intellisense and not just when the first overload is used. The reason for making an overload instead of just relying on the implementation declaration is that the two overloads warrant separate documentation as they don't behave the same.

packages/alfa-dom/src/node/attribute.ts
```diff
@@ -137,6 +159,11 @@ export namespace Attribute {
       owner
     );
   }
+
+  export enum EnumeratedAttributeError {
+    Missing = "missing",
+    Invalid = "invalid",
+  }
```
kasperisager, line 166:
```suggestion
```

No longer needed.

packages/alfa-dom/test/node/attribute.spec.ts
```diff
@@ -0,0 +1,16 @@
+import { None } from "@siteimprove/alfa-option";
+import { Parser } from "@siteimprove/alfa-parser";
+import { Err, Ok } from "@siteimprove/alfa-result";
+import { test } from "@siteimprove/alfa-test";
+
+import { Attribute } from "../../src";
+
+const makeAttribute = (str: string) => Attribute.of(None, None, "dummy", str);
+
+test("parse space separated token list", (t) => {
```
kasperisager, line 10:
```suggestion
test("#tokens() parses a space separated token list", (t) => {
```

We should use proper sentences for test case descriptions and make sure to note which method/function is under test.

packages/alfa-table/src/helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+/**
+ * @see https://html.spec.whatwg.org/multipage/tables.html#empty-cell
+ */
+export function isEmpty(element: Element): boolean {
+  return (
+    element.children().isEmpty() &&
+    // \s seems to be close enough to "ASCII whitespace".
+    !!element.textContent().match(/^\s*$/)
```
kasperisager, line 39:
```suggestion
    /^\s*$/.test(element.textContent())
```

packages/alfa-table/src/helpers.ts
```diff
@@ -0,0 +1,51 @@
...
+    .getOr(Ok.of(failed))
+    .getOr(failed);
+}
+
+/**
+ * @see https://html.spec.whatwg.org/multipage/tables.html#empty-cell
+ */
+export function isEmpty(element: Element): boolean {
+  return (
+    element.children().isEmpty() &&
```
kasperisager, line 37: This will be `true` only if the element contains _no_ child nodes, including text nodes.

packages/alfa-table/src/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./header";
+export * from "./table";
+
```
kasperisager, line 3:
```suggestion
```

packages/alfa-table/src/microsyntaxes.ts
```diff
@@ -0,0 +1,36 @@
...
+  if (str.match(/^\s*$/)) {
+    return Outcomes.empty;
+  }
+  const raw = Number(str);
+  return isNaN(raw)
+    ? Outcomes.notNumber
+    : raw !== Math.floor(raw)
+    ? Outcomes.notInteger
+    : // 0 and -0 are different (but equal…) floats. Normalising.
+      Ok.of(raw === -0 ? 0 : raw);
```
kasperisager, line 18: This needs to properly parse an integer according to the HTML specification rather than rely on the rather odd number conversion rules of JavaScript.

packages/alfa-table/src/rowgroup.ts
```diff
@@ -0,0 +1,251 @@
...
+    /**
+     * @see https://html.spec.whatwg.org/multipage/tables.html#algorithm-for-processing-row-groups
+     */
+    export function from(group: Element): Result<Builder, string> {
+      if (
+        group.name !== "tfoot" &&
+        group.name !== "tbody" &&
+        group.name !== "thead"
+      )
+        return Err.of("This element is not a row group");
```
kasperisager, line 203:
```suggestion
      if (
        group.name !== "tfoot" &&
        group.name !== "tbody" &&
        group.name !== "thead"
      ) {
        return Err.of("This element is not a row group");
      }
```

Make sure not to rely on indentation like this; space sensitive code, especially in languages where it's optional, is a foot-gun as I'm sure Apple can attest to: https://nakedsecurity.sophos.com/2014/02/24/anatomy-of-a-goto-fail-apples-ssl-bug-explained-plus-an-unofficial-patch/ 🙈

packages/alfa-table/src/table.ts
```diff
@@ -0,0 +1,411 @@
...
+/**
+ * @see https://html.spec.whatwg.org/multipage/tables.html#table-processing-model
+ */
+export class Table implements Equatable, Serializable {
+  private readonly _width: number;
+  private readonly _height: number;
+  private readonly _element: Element;
+  private readonly _cells: Array<Cell>;
+  private readonly _rowGroups: Array<RowGroup>;
+  private readonly _colGroups: Array<ColGroup>;
```
kasperisager, line 27: This should be `_columnGroups`.

packages/alfa-table/src/rowgroup.ts
Jym77: Shouldn't this be caught be linter/prettier? Maybe I just forgot to run it… 🙄

packages/alfa-dom/src/node/attribute.ts
Jym77: Note that I do hate the "single letter possibly T" convention for cases where the type variable has some meaning (like here) (it does make full sense in `Option<T>` where it has no meaning).

I mean, if we compile litterature, we probably have thousands of pages on why and how we should give good names to variables. I'm having a hard time seeing this conclude by "oh yeah, if it's a type variable, just name it T"… That sounds very similar to advising to name every (code) variable i, j, k, …

kasperisager: I absolutely agree that variables in general should have proper names. The way I see it though, type variables don't fall in the category of general variables; there should never be more than at most a couple of them in any given context, unlike variables in general. If you need to name them anything beyond a single letter in order to distinguish them then there's too many.

If a certain type variable has "special" meaning then we simply pick a letter that we find more appropriate, such as `Parser<I, T, E>` which uses `I` for `Input` and `E` for `Error`.

packages/alfa-table/src/rowgroup.ts
kasperisager: ESLint has a rule that can catch it 👍

packages/alfa-table/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "@siteimprove/alfa-equatable": "^0.0.0",
+    "@siteimprove/alfa-iterable": "^0.0.0",
+    "@siteimprove/alfa-json": "^0.0.0",
+    "@siteimprove/alfa-option": "^0.0.0",
+    "@siteimprove/alfa-predicate": "^0.0.0"
+  },
+  "devDependencies": {
+    "@siteimprove/alfa-test": "^0.0.0"
+  }
+}
```
kasperisager, line 38:
```suggestion
{
  "$schema": "http://json.schemastore.org/package",
  "private": true,
  "name": "@siteimprove/alfa-table",
  "homepage": "https://siteimprove.com",
  "version": "0.0.0",
  "license": "MIT",
  "description": "Suite of open and standards-based tools for performing reliable accessibility conformance testing at scale",
  "repository": {
    "type": "git",
    "url": "https://github.com/siteimprove/alfa.git",
    "directory": "packages/alfa-table"
  },
  "bugs": "https://github.com/siteimprove/alfa/issues",
  "main": "src/index.js",
  "types": "src/index.d.ts",
  "files": [
    "src/**/*.js",
    "src/**/*.d.ts"
  ],
  "dependencies": {
    "@siteimprove/alfa-comparable": "^0.0.0",
    "@siteimprove/alfa-dom": "^0.0.0",
    "@siteimprove/alfa-equatable": "^0.0.0",
    "@siteimprove/alfa-iterable": "^0.0.0",
    "@siteimprove/alfa-json": "^0.0.0",
    "@siteimprove/alfa-option": "^0.0.0",
    "@siteimprove/alfa-predicate": "^0.0.0"
  },
  "devDependencies": {
    "@siteimprove/alfa-test": "^0.0.0"
  },
  "publishConfig": {
    "access": "public",
    "registry": "https://npm.pkg.github.com/"
  }
}
```

I did some updating of package manifests on `master`.

packages/alfa-table/src/microsyntaxes.ts
```diff
@@ -0,0 +1,105 @@
...
+  for (; position < input.length; position++) {
+    const char = input[position];
+    if (isAsciiDigit(char)) {
+      value = value * 10 + digitToNumber(char);
+    } else {
+      break;
+    }
+  }
+
+  return Ok.of(positive ? value : 0 - value);
```
kasperisager, line 89:
```suggestion
  return Ok.of(positive ? value : -value);
```

Jym77: Yeah, I chose `0-value` because the spec is stating "return the result of subtracting _value_ from zero"…

Jym77:
> Also, is there a reason why the lock file (`yarn.lock`) has changed?

I guess I ran `yarn install` after adding the `alfa-table` package and that decided to update other packages… 🤷

packages/alfa-table/src/microsyntaxes.ts
Jym77: Fun fact: we actually do need to use `0-value`.

`0` and `-0` are somehow different numbers 🙈 and `-value` will be `-0` if `value` is `0` (but `0-value` will be `0` (and when `value` is `-0`, both `-value` and `0-value` are `0`)).

kasperisager: Ah, that's fair then! That's also avoids having to deal with negative zero shenanigans I guess 🤷

PR #202, opened 2020-03-25
Resolve presentational role conflicts when building ARIA nodes

### LINKED ISSUE #184: Implement ARIA presentational roles conflict resolution
This bit of the ARIA spec is currently not considered when we compute roles for elements: https://www.w3.org/TR/wai-aria/#conflict_resolution_presentation_none

kasperisager:
Closes #184.

### DIFF
--- a/packages/alfa-aria/src/node.ts
+++ b/packages/alfa-aria/src/node.ts
@@ -189,7 +189,9 @@ export namespace Node {
 
       // Text nodes are _always_ exposed in the accessibility tree.
       if (dom.Text.isText(node)) {
-        accessibleNode = Branched.of(Text.of(node, node.data));
+        accessibleNode = Branched.of(
+          Text.of(node, node.data.replace(/\s+/g, " "))
+        );
       }
 
       // Element nodes are _sometimes_ exposed in the accessibility tree.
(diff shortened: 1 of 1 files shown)


kasperisager:
Both of those cases are actually made very explicit in ARIA:

> User agents MUST NOT expose elements having explicit or inherited presentational role in the accessibility tree, with these exceptions: [...]

As such, `` isn't even relevant to conflict resolution as it has neither an explicit nor an inherited presentational role. I'd much prefer not to deal with this case just yet though as it very much is an edge case; for now, Alfa will exclude such an `` element from its accessibility tree.

The point about Chrome ignoring `aria-hidden` when an element is focusable is more concerning though. The particular case of `<foo aria-hidden="true" tabindex=0>` isn't a question of conflict resolution either, as that element also doesn't have an explicit nor inherited presentational role. Per the exclusion criteria, it however shouldn't be included:

> If not already excluded from the accessibility tree per the above rules, user agents SHOULD NOT include the following elements in the accessibility tree:
>
> - Elements, including their descendants, that have `aria-hidden` set to true. In other words, `aria-hidden="true"` on a parent overrides `aria-hidden="false"` on descendants.

That's however a _SHOULD NOT_, not a _MUST_, so I'm guessing Chrome is actually free to expose elements with `aria-hidden="true"` if they are deemed relevant for operability purposes.

PR #205, opened 2020-04-08
SIA R45 and SIA R46

Jym77:
Implementation of the table rules.


packages/alfa-aria/src/feature.ts
```diff
@@ -633,23 +633,26 @@ Feature.register(
...
-          case Header.Variant.ColGroup: return "ColumnHeader";
+          case Header.Variant.ColGroup:
+            return "columnheader";
           case Header.Variant.Row:
-          case Header.Variant.RowGroup: return "RowHeader"
-
-        }}
-      );
+          case Header.Variant.RowGroup:
+            return "rowheader";
```
Jym77, line 653: The fact that I had to correct the casing here is somewhat screaming for a better solution (we shouldn't be able to return a name here which has not been registered as a role in https://github.com/Siteimprove/alfa/blob/master/packages/alfa-aria/src/role/structure/row-header.ts (or elsewhere).

I kinda guess that we can't have a dynamic type constraint that the string we return here must have been defined via `register` first…
I guess the easiest way would be to maintain a list of role (i.e. an enumeration) and only pick from there (both for `register` and for this).

That would also allow stricter typing of predicates such as `hasRole` where it can be good to prevent a typo like `hasRole("buttton")` (3 "t") that would make the test always fail. `hasRole(Role.Name.Button)` would be safer.

What do you think?

kasperisager: This is indeed a little unfortunate. I have an idea of how to implement this nicely but haven't gotten around to actually doing it yet. The key is incrementally building up some sort of container with a union of all the valid role names and then extracting the union from this container. The container would then serve as a global, strongly-typed registry of the valid roles.

packages/alfa-rules/src/sia-r45/rule.ts
```diff
@@ -0,0 +1,95 @@
...
+              if (not(isPerceivable(device))(table.get())) return result;
+
+              // if the table is exposed, record the ownership and return the headers attribute
+              const headers = cell.attribute("headers").get();
+              ownership = ownership.set(headers, table.get());
+              result = Some.of(headers);
+              return result;
+            })
+            .filter((option) => option.isSome())
+            .map((option) => option.get())
```
kasperisager, line 58: A cleaner approach to this would probably be to start from each `<table>` and then drill down into the cells, rather than the other way around. It's becomes a little awkward having to repeatedly walk back up the tree to look for the owning table, and repeatedly checking if the same table is perceivable.

kasperisager:
This needs a rebase on top of `master` as well :octocat:

packages/alfa-rules/src/sia-r45/rule.ts
```diff
@@ -0,0 +1,92 @@
...
+import { isPerceivable } from "../common/predicate/is-perceivable";
+
+const { isElement, hasName, hasNamespace } = Element;
+const { and } = Predicate;
+
+export default Rule.Atomic.of<Page, Attribute>({
+  uri: "https://siteimprove.github.io/sanshikan/rules/sia-r45.html",
+  evaluate({ device, document }) {
+    // records in which <table> is located each "headers" attribute
+    let ownership: Map<Attribute, Element> = Map.empty();
```
kasperisager, line 18:
```suggestion
    let ownership = Map.empty<Attribute, Element>();
```

packages/alfa-rules/src/sia-r46/rule.ts
```diff
@@ -0,0 +1,100 @@
...
+
+export default Rule.Atomic.of<Page, Element>({
+  uri: "https://siteimprove.github.io/sanshikan/rules/sia-r46.html",
+  // only working with HTML elements, not with roles.
+  // Thus https://act-rules.github.io/rules/d0f69e/#passed-example-2 will be ignored.
+  // Need to copy the table building algo for roles to handle it.
+  // Is it worth it?
+  evaluate({ device, document }) {
+    // records in which <table> is located each <th>
+    let ownership: Map<Element, Element> = Map.empty();
```
kasperisager, line 26:
```suggestion
    let ownership = Map.empty<Element, Element>();
```

PR #207, opened 2020-04-10
Factor isElement in isDocumentElement

Jym77:
`isDocumentElement` was always used together with `isElement`. Refactoring these together.


packages/alfa-rules/src/common/predicate/is-document-element.ts
```diff
@@ -1,11 +1,12 @@
-import { Element, Namespace } from "@siteimprove/alfa-dom";
+import { Element, Namespace, Node } from "@siteimprove/alfa-dom";
```
kasperisager, line 1:
```suggestion
import { Element, Namespace } from "@siteimprove/alfa-dom";
```

PR #208, opened 2020-04-10
Allow n-ary and

Jym77:
Many chaining of `and` is done with one narrowing predicate (`isElement`, usually), and several non-narrowing. This allows a n-ary `and` with predicates 2, …, n all having the same type. Can be used in several case to shorten and clarify code.

+ proof of concept of usage in one case. We probably don't need to refactor everything now and it can happen little by little.


packages/alfa-predicate/src/predicate.ts
```diff
@@ -50,6 +50,16 @@ export namespace Predicate {
...
 
+  export function and<T, U extends T, V extends U>(
+    left: Predicate<T, U>,
+    ...right: Array<Predicate<U, V>>
+  ): Predicate<T, V> {
+    return right.reduce<Predicate<T, V>>(
+      (acc, cur) => binanyAnd(acc, cur),
+      left
+    );
+  }
```
kasperisager, line 61: This does actually allow callers to narrow the type `T` to a `U` and then a `V`, but will only do the last bit for the first of the `right` predicates which doesn't seem very intuitive. Also, the right predicates are currently optional so one can do `and(left)` without supplying any `right`s. That's not really an API design that we want. Here's what I'm thinking we'd want though:

```ts
function and<T, U extends T, V extends U>(
  left: Predicate<T, U>,
  right: Predicate<U, V>
): Predicate<T, V>;

function and<T>(
  left: Predicate<T>,
  right: Predicate<T>,
  ...rest: Array<Predicate<T>>
): Predicate<T>;

function and<T>(
  ...predicates: Array<Predicate<T>>
): Predicate<T> {
  return (value) => predicates.reduce<boolean>(
    (holds, predicate) => holds && predicate(value),
    true
  );
}
```

This does two things:

1. Ensure that at least two predicates are passed.
2. Only perform narrowing when a full chain can be established, such as `T -> U -> V`. When it can't, no narrowing is performed.

We could even extend this with overloads for 2+ predicates, allowing for even longer narrowing chains. If variadic kinds are ever implemented in TypeScript, we could then rely on these to get rid of the overloads.

kasperisager: ...and let's do the same for `or()` now that we're at it 😄

Jym77: Yes, I did consider having `and(left, right, ...rest)` and somehow discarded it 🤷

packages/alfa-predicate/src/predicate.ts
```diff
@@ -44,19 +44,39 @@ export namespace Predicate {
...
-    return function and(value) {
-      return fold(left, value, right, contradiction);
-    };
+  ): Predicate<T, V>;
+
+  export function and<T, U extends T>(
+    left: Predicate<T, U>,
+    right: Predicate<U>,
+    ...rest: Array<Predicate<U>>
+  ): Predicate<T, U>;
```
kasperisager, line 53:
```suggestion
  export function and<T>(
    left: Predicate<T>,
    right: Predicate<T>,
    ...rest: Array<Predicate<T>>
  ): Predicate<T>;
```

This shouldn't perform any narrowing as we otherwise end up with an API that will only narrow according to the first predicate despite taking multiple predicates.

Jym77: But this is precisely the use case I wanted to cater to, typically in https://github.com/Siteimprove/alfa/pull/208/files#diff-77b9407fe13c6d0cbc183bade8fddb13L34-L50 (and many other places in rules)
the first predicate is usually `Predicate<Node, Element>` and the rest are `Predicate<Element, Element>` If we don't narrow on the first, we still have to keep it separate which is also pretty weird : `and(isElement, and(foo, bar, baz))`.

Moreover, if we have 3 or more narrowing predicates chained together we will have to chain `and`s anyway (`and(isFoo, isBar, isBaz)` doesn't narrow all the way down). Same with 2 narrowing predicates and more non-narrowing predicates (`and(isElement, isElementSubClass, foo, bar, baz))` won't work whether the first argument narrows or not).
How often do we have such cases? Should we rather solve them by having a `and(p1, p2, p3)`/`and(p1, p2, ...rest)` overload that would narrow thrice/twice?

I guess, my point is I do see a rather frequent use case that becomes possible by narrowing on first predicate and I think it's worth making this use case more natural to write.

kasperisager: I do see your point about such as use case. The current API is not generally intuitive though, which is my concern. `and()` should either narrow according to the entire predicate chain or it should not narrow at all; the current behaviour of narrowing only according to the first predicate when 2+ are given is not intuitive, even if it does support a specific use case. It might as well narrow according to the second predicate as well, given that at least two are required, but then it won't narrow according to the rest.

If narrowing is needed for chains of 2+ predicates then we need to add whichever number of overloads will be required to support our use cases and hope for variadic kinds to arrive in the future. Either that, or a combination of the narrowing and non-narrowing `and()`, such as `and(isNode, and(isElement, ...))`. I'm personally in favour of the latter as it's the simpler approach but I'd be okay with the former as well 😄

PR #209, opened 2020-04-10
SIA R53 and SIA R61

Jym77:


packages/alfa-rules/src/sia-r61/rule.ts
```diff
@@ -0,0 +1,62 @@
...
+            return [document];
+          }
+        }
+        return [];
+      },
+
+      expectations(target) {
+        return {
+          1: expectation(
+            Accessible.Node.from(firstHeading, device).every((accNode) =>
```
Jym77, line 39: As far as I understand, `every` here means "for every browser in te browser specific list that we get".

Do we have a way to instead map the test and return browser specific outcome?

packages/alfa-rules/src/common/expectation/get-aria-level.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export function getAriaLevel(
+  element: Element,
+  device: Device
+): Branched<number, Browser> {
+  return Node.from(element, device).map((accNode) =>
+    accNode
+      .attribute("aria-level")
+      .map((level) => Number(level))
+      .getOr(NaN)
```
Jym77, line 15: I'd really love if the aria attribute could be typed and this conversion was not needed here…

packages/alfa-rules/src/sia-r53/rule.ts
```diff
@@ -0,0 +1,59 @@
...
+          device
+        );
+
+        return {
+          1: expectation(
+            thisAriaLevel.flatMap((thisLevel) =>
+              previousAriaLevel.map(
+                (previousLevel) => previousLevel >= thisLevel - 1
+              )
+            ).every(b => b),
```
Jym77, line 45: Is there a way to return browser specific outcomes instead of collapsing everything?

packages/alfa-rules/src/sia-r61/rule.ts
```diff
@@ -0,0 +1,57 @@
...
+            return [document];
+          }
+        }
+        return [];
+      },
+
+      expectations(target) {
+        return {
+          1: expectation(
+            getAriaLevel(firstHeading, device).every((level) => level === 1),
```
Jym77, line 39: Is there a way to return browser specific outcomes instead of collapsing everything?

packages/alfa-rules/src/common/expectation/get-aria-level.ts
```diff
@@ -0,0 +1,17 @@
+import { Node } from "@siteimprove/alfa-aria";
+import { Branched } from "@siteimprove/alfa-branched";
+import { Browser } from "@siteimprove/alfa-compatibility";
+import { Device } from "@siteimprove/alfa-device";
+import { Element } from "@siteimprove/alfa-dom";
+
+export function getAriaLevel(
```
kasperisager, line 7: This should be a common predicate, probably named `hasHeadingLevel()`.

packages/alfa-rules/src/sia-r61/rule.ts
```diff
@@ -0,0 +1,57 @@
...
+  evaluate({ device, document }) {
+    let firstHeading: Element;
+
+    return {
+      applicability() {
+        if (hasChild(and(Element.isElement, isDocumentElement()))(document)) {
+          const myFirstHeading = document
+            .descendants({ flattened: true })
+            .filter(and(Element.isElement, hasRole(hasName(equals("heading")))))
+            .first();
```
kasperisager, line 27: `.filter().first()` -> `.find()`

packages/alfa-rules/src/sia-r61/rule.ts
```diff
@@ -0,0 +1,57 @@
...
+          if (myFirstHeading.isSome()) {
+            firstHeading = myFirstHeading.get();
+            return [document];
+          }
+        }
+        return [];
```
kasperisager, line 33: If you need to share things between the applicability and the expectations, run the needed code directly inside `evaluate()` instead of deferring it to `applicability()`.

packages/alfa-rules/test/sia-r53/rule.spec.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { passed, failed, inapplicable } from "../common/outcome";
+
+const { and } = Predicate;
+
+const device = Device.standard();
+
+function getById(document: Document, id: string): Element {
+  return document
+    .descendants()
+    .find(and(Element.isElement, (element) => element.id.getOr("") === id))
```
kasperisager, line 21:
```suggestion
    .find(and(Element.isElement, (element) => element.id.includes(id)))
```

Jym77:
Need re-approval after adaptation to other merged PR…

PR #210, opened 2020-04-12
Allow hasFoo to directly take a list of names

Jym77:
Several of the `hasFoo` predicates are almost exclusively used in the pattern `hasFoo(equals(foo, …, foo))`.
It is possible to overload the signature to avoid the repeated `equals` all over the place.

The resulting code (in `hasFoo`) is not super clean however (due to overload with rest parameter forcing to manually unpack `arguments`)… So I'm not entirely sure this is worth it…

PoC on `hasNamespace` to discuss whether we want to do it further or not.
Other possibilitIes:
* overload with array parameter, but forces array even if there is only one foo (`hasFoo([foo])`). Can further be overloaded with single foo but `hasFoo(foo)`/`hasFoo([foo, bar])` is not a nice pattern to use.
* add extra `hasFooEquals` predicate to avoid the overload.

### DIFF
--- a/packages/alfa-rules/src/common/predicate/has-namespace.ts
+++ b/packages/alfa-rules/src/common/predicate/has-namespace.ts
@@ -1,6 +1,7 @@
 import { Element, Namespace } from "@siteimprove/alfa-dom";
 import { Predicate } from "@siteimprove/alfa-predicate";
-import equals = Predicate.equals;
+
+const { equals } = Predicate;
 
 export function hasNamespace(
   predicate: Predicate<Namespace>
--- a/packages/alfa-rules/src/common/predicate/has-namespace.ts
+++ b/packages/alfa-rules/src/common/predicate/has-namespace.ts
@@ -1,8 +1,25 @@
 import { Element, Namespace } from "@siteimprove/alfa-dom";
 import { Predicate } from "@siteimprove/alfa-predicate";
+import equals = Predicate.equals;
 
 export function hasNamespace(
-  predicate: Predicate<Namespace> = () => true
-): Predicate<Element> {
+  predicate: Predicate<Namespace>
+): Predicate<Element>;
+export function hasNamespace(
+  ...namespaces: Array<Namespace>
+): Predicate<Element>;
+export function hasNamespace(): Predicate<Element> {
+  let predicate: Predicate<Namespace>;
+
+  if (arguments.length === 0) {
+    predicate = () => true;
+  }
+  if (arguments.length === 1 && typeof arguments[0] === "function") {
+    predicate = arguments[0];
+  }
+  if (arguments.length !== 0 && typeof arguments[0] !== "function") {
+    predicate = equals(...arguments);
+  }
+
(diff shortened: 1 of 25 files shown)


packages/alfa-rules/src/common/predicate/has-namespace.ts
```diff
@@ -1,8 +1,25 @@
...
+
+  if (arguments.length === 0) {
+    predicate = () => true;
+  }
+  if (arguments.length === 1 && typeof arguments[0] === "function") {
+    predicate = arguments[0];
+  }
+  if (arguments.length !== 0 && typeof arguments[0] !== "function") {
+    predicate = equals(...arguments);
+  }
```
kasperisager, line 22:
```suggestion
  predicate: Predicate<Namespace>
): Predicate<Element>;

export function hasNamespace(
  namespace: Namespace,
  ...rest: Array<Namespace>
): Predicate<Element>;

export function hasNamespace(
  namespaceOrPredicate: Namespace | Predicate<Namespace>,
  ...namespaces: Array<Namespace>
): Predicate<Element> {
  let predicate: Predicate<Namespace>;
  
  if (typeof namespaceOrPredicate === "function") {
    predicate = namespaceOrPredicate;
  } else {
    predicate = equals(namespaceOrPredicate, ...namespaces);
  }
```

Jym77: Yes, feels much less hacky (and forces at least one namespace…)
This does however prevent the use of a default predicate. But we don't seem to use it, so I guess it's OK.

kasperisager: If we need it, it's easy to introduce an additional overload and a branch that checks for `namespaceOrPredicate === undefined` 👍

Jym77:
Tried to do the same to `hasName` and `hasRole`.
`hasRole` is becoming somewhat ugly and I'm less sure of the value here…

packages/alfa-rules/src/common/predicate/has-role.ts
```diff
@@ -1,11 +1,51 @@
...
+  } else {
+    if (typeof nameOrPredicateOrOptions === "string") {
+      options = {};
+      predicate = hasName(equals(nameOrPredicateOrOptions, nameOrOptions, ...names));
+    } else {
+      options = nameOrPredicateOrOptions;
+      predicate = hasName(equals(nameOrOptions, ...names));
+    }
+  }
+
```
kasperisager, line 48: We need to factor out the options from this one, otherwise things just get too complex 😮 As such, I'm think that two additional predicates, `hasImplicitRole()` and `hasExplicitRole()`, would be a better way to go about this.

Jym77: Yes, this is clearly too ugly. And I agree that given the options we can have, making specific functions is probably better. It will also make the call easier to read (`hasExplicitRole()` reads better than `hasRole({implicit: false})` which needs to remember that the only other possibility is explicit).

PR #215, opened 2020-04-21
Is html element by name

Jym77:
This is a fairly common pattern that could be factored in…

### DIFF
--- a/packages/alfa-dom/src/node/element.ts
+++ b/packages/alfa-dom/src/node/element.ts
@@ -2,7 +2,6 @@ import { Iterable } from "@siteimprove/alfa-iterable";
 import { Mapper } from "@siteimprove/alfa-mapper";
 import { None, Option, Some } from "@siteimprove/alfa-option";
 import { Predicate } from "@siteimprove/alfa-predicate";
-import { Err, Result } from "@siteimprove/alfa-result";
 import { Sequence } from "@siteimprove/alfa-sequence";
 
 import { Namespace } from "../namespace";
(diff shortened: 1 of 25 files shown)


packages/alfa-dom/src/node/element.ts
```diff
@@ -342,6 +341,16 @@ export namespace Element {
...
+  export function isHtmlElementWithName(
+    name: string,
+    ...rest: Array<string>
+  ): Predicate<unknown, Element> {
+    return and(
+      isElement,
+      and(hasNamespace(Namespace.HTML), hasName(name, ...rest))
+    );
+  }
+
```
kasperisager, line 353: I'm not too keen on this, for two reasons:

1. The name is _really_ long and should also properly case `HTML`, making the name even more obscure.

2. This makes it difficult to disassemble the predicate to extend it at call sites. For example, if I have a call site that today looks for HTML `<a>` elements but tomorrow will need to look for SVG `<a>` elements as well, this would be the diff:

    ```diff
    -isHtmlElementWithName("a");
    +and(
    +  isElement,
    +  and(
    +    hasNamespace(Namespace.HTML, Namespace.SVG),
    +    hasName("a")
    +  )
    +);
    ```

    Had we not combined everything into one massive predicate, the diff would instead be:

    ```diff
     and(
       isElement,
       and(
    -    hasNamespace(Namespace.HTML),
    +    hasNamespace(Namespace.HTML, Namespace.SVG),
         hasName("a")
       )
     );
    ```

    I want us to have building blocks that are small enough to easily assemble _and_ disassemble as needed. Of course, they shouldn't be so small that they become cumbersome to work with, but we also shouldn't introduce building blocks that are too large just for the sake of avoiding repetition.

Jym77: Fair point.

PR #217, opened 2020-04-22
Use hasName in alfa-aria

Jym77:
Using the new predicate to clean up code a bit.

### DIFF
--- a/packages/alfa-aria/src/get-name.ts
+++ b/packages/alfa-aria/src/get-name.ts
@@ -9,7 +9,7 @@ import { Style } from "@siteimprove/alfa-style";
 
 import { Role } from "./role";
 
-const { isElement } = Element;
+const { hasName, isElement } = Element;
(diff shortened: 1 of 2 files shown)


packages/alfa-aria/src/get-name.ts
```diff
@@ -539,16 +537,14 @@ function getLabel(element: Element): Option<Element> {
...
-              and(
-                Element.isElement,
-                (element) =>
-                  element.name === "label" &&
-                  element.attribute("for").some((attr) => attr.value === id)
+          const label = root.descendants().find(
+            and(
+              Element.isElement,
+              and(hasName("label"), (element) =>
+                element.attribute("for").some((attr) => attr.value === id)
```
Jym77, line 544: This is of course a candidate for `hasAttribute` but one case is probably not enough reason to lift it from `alfa-rules` to `alfa-dom`…

PR #218, opened 2020-04-24
Update sia r21

Jym77:
Updating according to https://github.com/act-rules/act-rules.github.io/pull/1277 and https://github.com/Siteimprove/sanshikan/pull/100

### DIFF
--- a/packages/alfa-rules/src/sia-r21/rule.ts
+++ b/packages/alfa-rules/src/sia-r21/rule.ts
@@ -27,7 +27,7 @@ export default Rule.Atomic.of<Page, Attribute>({
               isElement,
               and(
                 hasNamespace(Namespace.HTML, Namespace.SVG),
-                hasAttribute("role", value => not(isEmpty)(value.trim())),
+                hasAttribute("role", (value) => not(isEmpty)(value.trim())),
                 not(isIgnored(device))
               )
             )
--- a/packages/alfa-rules/src/sia-r21/rule.ts
+++ b/packages/alfa-rules/src/sia-r21/rule.ts
@@ -27,7 +27,7 @@ export default Rule.Atomic.of<Page, Attribute>({
               isElement,
               and(
                 hasNamespace(Namespace.HTML, Namespace.SVG),
-                hasAttribute("role", not(isEmpty)),
+                hasAttribute("role", value => not(isEmpty)(value.trim())),
                 not(isIgnored(device))
               )
             )
--- a/packages/alfa-rules/src/sia-r21/rule.ts
+++ b/packages/alfa-rules/src/sia-r21/rule.ts
@@ -8,7 +8,7 @@ import { Page } from "@siteimprove/alfa-web";
 import { expectation } from "../common/expectation";
 
 import { hasAttribute } from "../common/predicate/has-attribute";
-import { hasExplicitRole } from "../common/predicate/has-role";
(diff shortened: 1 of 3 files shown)


packages/alfa-rules/src/sia-r21/rule.ts
```diff
@@ -40,7 +40,7 @@ export default Rule.Atomic.of<Page, Attribute>({
 
         return {
           1: expectation(
-            hasExplicitRole()(owner),
+            hasRole()(owner),
```
Jym77, line 43: Sanshikan is allowing defaulting to implicit role, copying it here.
https://github.com/Siteimprove/sanshikan/blob/master/rules/sia-r21.md#expectation
and second passed example.

kasperisager:
@Jym77 I was just updating our GDS benchmark and noticed that due to these changes we're no longer failing this case: https://alphagov.github.io/accessibility-tool-audit/tests/html-invalid-aria-role-names.html. This led me down the rabbit hole of figuring out _why_ we diverge from the ACT-R rule on the bit about implicit roles and then this turned up: https://github.com/act-rules/act-rules.github.io/issues/373. To summarise the conclusion of that issue, the rule should expect all tokens to be valid, which for some reason isn't yet the case for the ACT-R rule. This is of course coupled with the perfectly reasonable assumption that the `role` attribute is used to ensure WCAG compliance.

Jym77:
Ah yes, I see the point.
Essentially, there are 2 possible ways to handle this whole question (of `role` attribute):
* The strict way: every single token has to be valid. The fallback are meant for UA that don't handle some roles. "if you're using a role, only use existing ones".
Problem: in some cases, we flag an incorrect token but every single UA/AT is going to fallback to implicit role (or second token) and there is no actual accessibility issue. So this is more of a smell than an error. This is often the case with checks related to parsing where UA/AT tend to auto-fix a lot of things.
* The lenient way: if there is a `role` attribute, the element must have a `role`. Thus implicit role is also OK. "you wanted this to have a role, well it has one".
Problem: we don't detect typos in explicit roles which are probably producing something that was not intended by the author.

I agree that the middle way "if you use role, at least one explicit role must be OK" doesn't make a lot of sense. Essentially, `role="none presention"` should fail if we decide to be strict (and would be an actual accessibility issue for AT that don't implement `none`…)

I guess the fact that the strict solution is producing "false positive" (rule failing without an accessibility issue, e.g. `role="noen presentation"`) is the reason why ACT-R didn't go that way.

kasperisager:
I'm fairly certain that the ACT-R rule is the way it currently is as only half of the issue got solved; the assumption was added, but the expectation was never updated accordingly (see https://github.com/act-rules/act-rules.github.io/issues/373#issuecomment-448548507). If we assume that the `role` attribute is used for WCAG compliance then anything you put in there should be valid, of course with the option for fallbacks for older UA/AT. As with all assumptions we make, as long as the assumption holds then the results will be true positives.

The above would mean that `role="none presention"` would pass as both roles a valid and the second role added as a fallback, but `role="noen presention"` would fail as `noen` isn't a valid role and, by assumption, preventing WCAG conformance. This is of course a case where the assumption doesn't hold and as a consequence there's no actual accessibility issue, but that's something I'm quite OK with; `noen` is gibberish and should be removed to ensure that the assumption does hold.

PR #220, opened 2020-04-28
Include media conditions in `Sheet`

### LINKED ISSUE #219: The `Sheet` class does not include media conditions
The `Sheet` class of the @siteimprove/alfa-dom package does not currently include media conditions:

https://github.com/Siteimprove/alfa/blob/afe5190e56a3a2805bcaedb791272c4f239c27e7/packages/alfa-dom/src/style/sheet.ts#L7-L12

Media conditions can be added to style sheets by means of `<link rel="stylesheet" media="<conditions>">` and Alfa will currently not account for these conditions.

kasperisager:
Closes #219.

We need to coordinate this with @lukasbob as well as the crawler services that handle page rendering will need to include media conditions in their internal page representation if they don't already do so.

### DIFF
--- a/packages/alfa-cascade/src/selector-map.ts
+++ b/packages/alfa-cascade/src/selector-map.ts
@@ -120,6 +120,18 @@ export class SelectorMap {
     };
 
     for (const sheet of sheets) {
+      if (sheet.disabled) {
+        continue;
+      }
+
+      if (sheet.condition.isSome()) {
+        const query = Media.parse(sheet.condition.get());
+
+        if (query.isNone() || !query.get().matches(device)) {
+          continue;
+        }
+      }
+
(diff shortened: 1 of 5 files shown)


kasperisager:
This is on the crawler teams radar 👌

kasperisager:
For future reference: The changes introduced here are actually backwards incompatible as the `Sheet.JSON` interface now requires a `condition` property, meaning that the TypeScript compiler will no longer accept instances of `Sheet.JSON` _without_ a `condition` property. As we're however still at version 0, we won't be labelling this a major change.

kasperisager:
On second thought, let's label this a major change for posterity but state that as we're still at version 0 these changes will only cause a minor bump.

PR #222, opened 2020-04-28
Only look for `Element` nodes when determining ownership in SIA-R68

kasperisager:


packages/alfa-rules/test/sia-r68/rule.spec.tsx
```diff
@@ -48,6 +48,27 @@ test("evaluate() passes a list with a list item and a group of list items", asyn
   ]);
 });
 
+test("evaluate() ignores non-element children when determining ownership", async (t) => {
+  const document = Document.of((self) => [
+    Element.fromElement(
+      <div role="list">
+        <span role="listitem">Item 1</span>{" "}
```
Jym77, line 55: I'm pondering whether we should add a comment "don't remove that `{" "}`, you stupid!" to try and prevent ourselves from shooting our tests in the foot in the next big refactoring…
(I've seen my editor adding a bunch of these when copy/pasting HTML from other sources which includes spaces between tags, and tend to aggressively remove them as useless garbage…)

PR #223, opened 2020-04-30
Update sia r45

Jym77:
Update to ignore ref to self, according to https://github.com/act-rules/act-rules.github.io/pull/1274


Jym77:
Actually, not correct. The rule has now 2 expectation…

PR #227, opened 2020-05-01
Unify test names

Jym77:
Just a clean up/unification on test names…


kasperisager:
I'm leaving the patch label off this one as these changes aren't something we'll include in the changelog.

PR #228, opened 2020-05-04
Provide initial implementation of SIA-R69

kasperisager:


packages/alfa-rules/src/sia-r69/rule.ts
```diff
@@ -0,0 +1,388 @@
...
+                and(hasRole(Role.hasName(equals("group"))), isDisabled),
+                node
+              )
+            ) {
+              return;
+            }
+          }
+
+          if (Text.isText(node)) {
+            if (test(and(isVisible(device), not(isIgnored(device))), node)) {
```
Jym77, line 49: I think there is a `isPerceivable` predicate for this.

packages/alfa-rules/src/sia-r69/rule.ts
```diff
@@ -0,0 +1,388 @@
...
+
+export default Rule.Atomic.of<Page, Text, Question>({
+  uri: "https://siteimprove.github.io/sanshikan/rules/sia-r69.html",
+  evaluate({ device, document }) {
+    return {
+      applicability() {
+        return visit(document);
+
+        function* visit(node: Node): Iterable<Text> {
+          if (Element.isElement(node)) {
```
Jym77, line 35: This (and https://github.com/Siteimprove/alfa/pull/228/files#diff-3dbf3df10d60073193da7fa74c7f22fcR48) make me think we maybe should have an `node#kind` or something because
```javascript
switch (node.kind) {
  case "element": …
  case "text": …
}
```
might be more readable.

packages/alfa-rules/src/sia-r69/rule.ts
```diff
@@ -0,0 +1,388 @@
...
+          target,
+          "What are the foreground colors of the text node?"
+        );
+
+        const backgrounds = Question.of(
+          "background-colors",
+          "color[]",
+          target,
+          "What are the background colors of the text node?"
+        );
```
Jym77, line 75: I kinda fail to see how these questions are not ask every single time 🤔 
But it seem to work 🤷

packages/alfa-rules/src/sia-r69/rule.ts
```diff
@@ -0,0 +1,388 @@
...
+            );
+          })
+        );
+
+        const parent = target.parent({ flattened: true }).get() as Element;
+
+        return {
+          1: getForeground(parent, device)
+            .map((foreground) => result.answer(foreground))
+            .flatMap((result) =>
```
Jym77, line 104: Naming the local variable `result`, same as the global(-ish) variable that is used both on the previous line and after this call (l109) is rather confusing…

packages/alfa-rules/src/sia-r69/rule.ts
kasperisager: The magic of monads 😄 The type of the first `result` is basically `Question<Question<..>>` with the outer `Question` providing the foreground colours and the inner `Question` providing the background colours:

https://github.com/Siteimprove/alfa/blob/4669b7900bc4d5a3e716a00879cc42516238df08/packages/alfa-rules/src/sia-r69/rule.ts#L77-L97

We then get attempt to determine the foreground colours automatically and, if that succeeds, we use them to unwrap a single layer of `result` so that it becomes `Question<...>`:

https://github.com/Siteimprove/alfa/blob/4669b7900bc4d5a3e716a00879cc42516238df08/packages/alfa-rules/src/sia-r69/rule.ts#L102-L103

We then repeat the same thing for the background colours to unwrap the last layer of the _new_ `result`:

https://github.com/Siteimprove/alfa/blob/4669b7900bc4d5a3e716a00879cc42516238df08/packages/alfa-rules/src/sia-r69/rule.ts#L104-L108

That's also the reason for naming both variables `result`. If all else fails, we simply fall back to the old `result`:

https://github.com/Siteimprove/alfa/blob/4669b7900bc4d5a3e716a00879cc42516238df08/packages/alfa-rules/src/sia-r69/rule.ts#L109

The result, no pun intended, of all this is that we either end up with an entirely unwrapped `result`, a single `Question<...>` layer, or a double `Question<Question<...>>` layer.

packages/alfa-rules/src/sia-r69/rule.ts
kasperisager: Let me know if https://github.com/Siteimprove/alfa/pull/228#discussion_r420061683 provides some clarity.

packages/alfa-rules/src/sia-r69/rule.ts
kasperisager: I do agree that would be ideal. The issue is that we can't narrow `Node` based on a property as it's not a union type; the DOM types are modelled using a class hierarchy.

packages/alfa-rules/src/sia-r69/rule.ts
Jym77: I think I misunderstood l103/106 which actually means "Hey, I got some answer to your question called `result` and I'm giving it now"…
Make sense now.

PR #231, opened 2020-05-06
Provide initial implementation of SIA-R82

kasperisager:
I'd love your input on this rule, @May-N, as it's primarily based on external input. The way it currently works is as follows:

1. For every form field, ask:
    1. Where are the error indicators, if any, for the form field?  
        - _If no error indicators are present, the rule passes._
2. For every error indicator of every form field, ask:
    1. Does the error indicator identify, in text, the form field it relates to?  
        - _If none do, or the one that does isn't perceivable, the rule fails._
    2. Does the error indicator describe, in text, the cause of the error or how to resolve it?  
        - _If none do, or the one that does isn't perceivable, the rule fails._


packages/alfa-rules/test/sia-r82/rule.spec.tsx
```diff
@@ -0,0 +1,146 @@
...
+    Option.of(self)
+  ),
+]);
+
+const input = document
+  .descendants()
+  .find(and(isElement, hasName("input")))
+  .get();
+
+const [error, invisibleError, ignoredError] = document
```
Jym77, line 38:
```suggestion
const [visibleError, invisibleError, ignoredError] = document
```
would be more consistent naming and help parsing the examples further down.

packages/alfa-rules/src/sia-r82/rule.ts
```diff
@@ -0,0 +1,188 @@
...
+                  "combobox",
+                  "listbox",
+                  "menuitemcheckbox",
+                  "menuitemradio",
+                  "radio",
+                  "searchbox",
+                  "slider",
+                  "spinbutton",
+                  "switch",
+                  "textbox"
```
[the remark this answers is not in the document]
kasperisager: I'm not entirely convinced as it currently aligns well with the actual rule definition. This will hopefully change once we fork the rule in Sanshikan and come up with a good term for "semantic role that inherits from the abstract input or select role, and
does not have a required context role that inherits from the abstract input or select role".

packages/alfa-rules/src/sia-r82/rule.ts
```diff
@@ -0,0 +1,188 @@
...
+          "Where are the error indicators, if any, for the form field?"
+        ).map((indicators) => [...indicators]);
+
+        return {
+          1: indicators.map((indicators) =>
+            expectation(
+              indicators.length === 0,
+              () => Outcomes.HasNoErrorIndicator,
+              () => {
+                function identifiesTarget(
```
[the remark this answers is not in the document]
kasperisager: Hoisting them outside the expectations seems fine 👍

packages/alfa-rules/src/sia-r82/rule.ts
```diff
@@ -186,3 +119,62 @@ export namespace Outcomes {
...
+      } else {
+        error = Outcomes.ErrorIndicatorIdentifiesTargetButIsNotPerceivable;
+      }
+    }
+
+    return identifiesTarget(indicators.slice(1), error, device);
+  });
+}
+
+function describesResolution(
```
Jym77, line 152: This feels some much like a `reduce` if it wasn't for the early escape…

kasperisager: I agree 😄 It _could_ be implemented as a fold, but that seemed a little wasteful. I'll look into possibly encoding this "early return" fold as a common operation.

Jym77: Maybe a statefull function with `for (indicator of indicators)` and a `break` is enough…
I'm cringing so much at the recursive call to `f(array.slice(1))` which looks like we don't know how to iterate over elements in an array…

kasperisager: The issue is that it's really constructing a deeply nested chain of `.map()` invocations, so neither a fold nor a loop will actually do us any good 😕

PR #232, opened 2020-05-10
Provide initial implementation of SIA-R81

kasperisager:
This still lacks this bit from the programmatic link context:

> [...] or in a table header cell that is associated with the table cell that contains the link.


packages/alfa-rules/src/sia-r81/rule.ts
```diff
@@ -0,0 +1,154 @@
...
+import { isIgnored } from "../common/predicate/is-ignored";
+
+import { Question } from "../common/question";
+
+const { isElement, hasNamespace, hasId } = Element;
+const { map, flatMap, isEmpty } = Iterable;
+const { and, or, not, equals } = Predicate;
+
+export default Rule.Atomic.of<Page, Iterable<Element>, Question>({
+  uri: "https://siteimprove.github.io/sanshikan/rules/sia-r41.html",
```
Jym77, line 29:
```suggestion
  uri: "https://siteimprove.github.io/sanshikan/rules/sia-r81.html",
```

kasperisager:
Good catch! 🤦

packages/alfa-rules/src/sia-r81/rule.ts
```diff
@@ -0,0 +1,157 @@
...
+        const elements = document
+          .descendants({ flattened: true, nested: true })
+          .filter(
+            and(
+              isElement,
+              and(
+                hasNamespace(Namespace.HTML, Namespace.SVG),
+                hasRole(
+                  or(Role.hasName("link"), (role) =>
+                    role.inheritsFrom(Role.hasName("link"))
```
Jym77, line 42: Here also we might want an "inclusive inherits" helper as it happens quite a few times.

packages/alfa-rules/src/sia-r81/rule.ts
```diff
@@ -0,0 +1,157 @@
...
+                  groups
+                    .get(node.name())
+                    .getOrElse(() => List.empty<Element>())
+                    .append(element)
+                );
+              }
+
+              return groups;
+            }, Map.empty<Option<string>, List<Element>>())
+            .filter((elements) => elements.size > 1)
```
Jym77, line 70: I see this one we do the size check 😁

PR #234, opened 2020-05-13
Implement SIA-R71, SIA-R72, SIA-R73, SIA-R74, SIA-R75, and SIA-R80

kasperisager:


packages/alfa-rules/src/sia-r73/rule.ts
```diff
@@ -0,0 +1,76 @@
...
+                    lineHeight.type === "keyword" &&
+                      lineHeight.value === "normal",
+                    () => Outcomes.IsNormal,
+                    () => Outcomes.IsInsufficient
+                  )
+              )
+          ),
+        };
+      },
+    };
```
Jym77, line 59: Wondering if this would read better with a `switch (lineHeight.type)` instead. Notably for not going through everything when we pass the `type` test but fail the `value` one.
```suggestion
switch (lineHeight.type) {
  case "number":
    return expectation(
      lineHeight.value >= 1.5,
      () => Outcomes.IsSufficient,
      () => Outcomes.IsInsufficient
      );
    case "length":
      …
    defaut: return Outcomes.IsInsufficient;
```

PR #237, opened 2020-05-20
Improve table building

Jym77:
The algorithm for internal headers was O(n⁶) (or rather O(n³m³) for an n×m table):
1. Need to go through each slot to assign header to it => nm
2. Each slot is looking at slots on its left and top => nm
3. finding which cell is covering the slot was done by going through the full cell list and testing `isCovering` => nm

Step 3 was that way because that allowed a lot of the code to be pure functional with no side effect by not relying on pointing to cell or stuff.

This PR remove step 3. `Cell.Builder` are now mutable. `Table.Builder` holds a `slots` array mapping from slot (coordinates) to the cell covering it. Thus finding the cell covering a slot is immediate.
Obvious drawback is that `Cell.Builder` are now mutated during table building since we want to keep the pointers in the `slots` array pointing to the same object. This implies that some steps are now side effects.

Step 2 can be greatly improved by going through the slots from the bottom-right corner and remembering what we went through. Apparently this is not needed yet as performances seem OK now. Given that it is not trivial (and will make the algorithm diverges a bit from the specs in how it does stuff), I'd rather postpone this until we figure out it is needed.


kasperisager:
It definitely sounds like you'll want to use lists rather than arrays for storing the slots then. That gives you effectively constant time `.get()` and `.set()` operations while still retaining immutability of both cells and slots.

Jym77:
> It definitely sounds like you'll want to use lists rather than arrays for storing the slots then. That gives you effectively constant time `.get()` and `.set()` operations while still retaining immutability of both cells and slots.

I don't get it. My usage is:
```javascript
slots[0][0] = cell[0];
cell[0] = new cell
```
=> Now, `slots[0][0]` is pointing to something that doesn't really exist anymore
```javascript
slots[0][0] = cell[0];
cell[0].foo = new value
```
=> Now, `slots[0][0]` is still pointing to the cell that holds the correct values.

Would using a list instead of an array make the first case work?

kasperisager:
No, not when you mutate the slots matrix like that. The point is to do copy-on-write, but rather than copy the entire matrix on every write, you can use lists instead which enable structural sharing and effectively constant time random access and writes.

packages/alfa-table/src/table.ts
```diff
@@ -224,29 +270,116 @@ export namespace Table {
...
+      y: number,
+      w: number,
+      h: number
+    ): boolean {
+      return some(this.cells, (cell) => cell.isDataCoveringArea(x, y, w, h));
+    }
+
+    public addHeadersVariants(): Builder {
+      return this.update({
+        cells: List.from(
```
Jym77, line 294: Somehow, removing this `List.from` is breaking all tests. I have little clue why. This is a bit weird and worrying, but I have already spend too much time pointlessly on this…
(it breaks the implicit header detection)

packages/alfa-table/src/table.ts
```diff
@@ -224,29 +270,116 @@ export namespace Table {
...
+    }
+
+    /**
+     * If principal cell is in a group, get all group headers that are in this group and above+lift of principal cell.
+     */
+    public getAboveLeftGroupHeaders(
+      kind: "row" | "column"
+    ): (principalCell: Cell.Builder) => Iterable<Cell.Builder> {
+      let anchor: "x" | "y",
+        groups: Iterable<Covering>,
```
Jym77, line 313: I'm not a huge fan of this `Covering` interface. Does the job, but feels a bit dirty…

Jym77:
@kasperisager Should be better now.
Got a bit out of hands with making everything non-mutable and reducing exposed footprint…

kasperisager:
This a breaking change as it changes several class getters from being exposed as arrays to being exposed as iterables.

packages/alfa-table/src/table.ts
Jym77: 🤦 Of course 🙈

PR #239, opened 2020-05-22
Provide initial implementation of SIA-R84

kasperisager:


packages/alfa-rules/src/sia-r84/rule.ts
```diff
@@ -0,0 +1,97 @@
...
+      },
+
+      expectations(target) {
+        return {
+          1: expectation(
+            or(
+              isTabbable(device),
+              hasDescendant(and(isElement, isTabbable(device)), {
+                flattened: true,
+              })
```
Jym77, line 44: Do we have a notion of inclusive descendant (https://dom.spec.whatwg.org/#concept-tree-inclusive-descendant)? Should we? (maybe as an option for `hasDescendant`)

kasperisager: That's a great point! I definitely think we should 👍

PR #242, opened 2020-05-27
Initial implementation of SIA-R83

kasperisager:


packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -0,0 +1,143 @@
...
+                )
+              ),
+              node
+            )
+          ) {
+            return;
+          }
+
+          if (test(and(isText, isVisible(device)), node)) {
+            if (collect) {
```
Jym77, line 44: Wondering if `collect` should be moved up in the main test to benefit from lazy test evaluation and avoid the more costly `isText` and `isVisible`. Probably not a big gain…
```suggestion
          if (collect && test(and(isText, isVisible(device)), node)) {
```

packages/alfa-rules/test/sia-r83/rule.spec.tsx
```diff
@@ -0,0 +1,118 @@
...
+
+test("evaluate() passes a text node that truncates overflow using ellipsis", async (t) => {
+  const document = Document.of((self) => [
+    Element.fromElement(
+      <div style="overflow: hidden; white-space: nowrap; text-overflow: ellipsis;">
+        Hello world
+      </div>,
+      Option.of(self)
+    ),
+  ]);
```
Jym77, line 22: Completely unrelated, but given how often we use that pattern in test cases, we probably should have a helper:
```javascript
function documentFromElement(element) {
  return Document.of(self => [ Element.fromElement(element, Option.of(self)) ]);
}
```

kasperisager: I do agree that it has become a little repetitive. I haven't found a good abstraction yet though and actually prefer spelling it out over using a helper, simply because it does so little but requires yet another import 🤔

PR #244, opened 2020-06-03
Add `Document#frame` property to better model embedded documents

kasperisager:
The `Element#content` property is currently used for modelling elements that allow embedded content, such as `<iframe>` and `<object>` elements. However, we did not have a property for modelling such a relation from the perspective of the embedded content itself, such as a `Document` node inside an `<iframe>` element. This pull request adds the `Document#frame` property for that purpose, modelled after [`Window.frameElement`](https://developer.mozilla.org/en-US/docs/Web/API/Window/frameElement).


kasperisager:
Changing this to a major change due to a87bc67.

PR #246, opened 2020-06-03
Rework the way expectations in rules are modelled

kasperisager:
This pull request aims to rework the way expectations in rules are modelled with the goal of allowing rules to expose additional diagnostic information through their expectations. The pull request introduces a new class, `Diagnostic`, and changes the `Expectation` type to use this class rather than simple `string` messages:

```ts
type Expectation = Option<Result<Diagnostic>>;
```

In addition to the usual `string` message, which is now included in `Diagnostic#message`, the `Diagnostic` class also includes a `data` property which is basically a bag of arbitrary data that can be attached to a `Diagnostic` object.

The biggest drawback of this approach is of course the loose typing of the `data` property as it's just a `Map<string, unknown>`. Another option that I'm also exploring is to define `Diagnostic` as either an interface or an abstract class with a default implementation. Rules would then either implement or subclass the `Diagnostic` interface or class if they need to provide additional diagnostic information. This would allow consumers to do things like narrow the general `Diagnostic` type to things like `ContrastDiagnostic` in the case of contrast rules, for example.

I'd love your input on this @Jym77 and I'd also like for you @singingknight to provide your thoughts on how you envision this working together with the Siteimprove backend services that consume Alfa.


Jym77:
I like the idea.

1. On one hand, I think that `Diagnostic` should be allowed to be any JSON-like stuff. I can easily foresee that once we have this tool (and the backend/frontend to support it), we'll want more rules to send precise diagnostic (duration (audio, refresh, …), linked element (transcript, label, error indicator, focus indicator, …), elements inside a document (the repeated sections inside a document that can't be collapsed, …), …)
So I think we do need to allow almost any serialisable object as data.

2. On the other hand, I think we need these diagnostic to be somehow strongly typed, on a rule basis. The diagnostics are going to be the contract between Alfa and its consumers (primarily Dory). I can also easily foresee that these contracts will change at some point for various reasons and these changes should trigger a type error in a consumer using the wrong version of Alfa, not a runtime error. So that if you update Alfa from v0 to v1 and forget to update your diagnostic reading routine from `diagnostic.data.foo` to `diagnostic.data.bar`, you crash early and hard rather than get a `undefined` value ghosting around in the prod build…

These are a bit opposite goals…
I think 2 is extremely important to avoid building a gun pointing toward our own feet.

kasperisager:
That's great input @Jym77, thanks! f10e87c should hopefully address both points.

singingknight:
Looks good.

A lot better than having to reimplement Alfa logic in Page Report.

packages/alfa-rules/src/sia-r69/diagnostic/contrast.ts
```diff
@@ -0,0 +1,125 @@
+import { Diagnostic } from "@siteimprove/alfa-act";
```
Jym77, line 1: This might need to be moved to a common folder once we get a rule for "color contrast (enhanced)".

packages/alfa-rules/src/sia-r69/diagnostic/contrast.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  export function isContrast(value: unknown): value is Contrast {
+    return value instanceof Contrast;
+  }
+
+  export class Pairing implements Equatable, Serializable {
+    public static of(
+      foreground: RGB,
+      background: RGB,
+      contrast: number
```
Jym77, line 72: Should we move the contrast computation here?
That would especially allow to write rule for enhanced contrast by re-using the computation in the shared "contrast diagnostic" instead of repeating it in the rule.

Given that contrast computation is an "absolute" external formula, I think it could make sense to delegate it here. OTOH, this gives extra responsibilities (computing contrast) to the diagnostic which is only a structured data storage and this is maybe not a good way to go 🤔

packages/alfa-rules/src/sia-r69/diagnostic/contrast.ts
```diff
@@ -0,0 +1,125 @@
...
+  ): Contrast {
+    return new Contrast(message, threshold, Array.from(pairings));
+  }
+
+  private readonly _threshold: number;
+  private readonly _pairings: Array<Contrast.Pairing>;
+
+  private constructor(
+    message: string,
+    threshold: number = 4.5,
```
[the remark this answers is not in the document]
kasperisager: Bad copy paste, good catch 🙈

packages/alfa-rules/src/sia-r69/diagnostic/contrast.ts
kasperisager: Yep, I imagine shared diagnostics moving into `common/diagnostic`.

packages/alfa-rules/src/sia-r69/diagnostic/contrast.ts
kasperisager: Yeah, I'm not too keen on moving logic into the diagnostics 🤔 The contrast computation should absolutely be moved out of the current rule, but the diagnostics aren't the destination.

Jym77: Yes, I agree that keeping diagnostics as a clean "storage only" structure is a good idea to keep separation of concern. They're just the paper on which the doctor wrote the conclusion, but not really part of the actual examination…

As for contrast computation, it will probably move to `common/expectations/` once sharing is needed.

kasperisager:
We'll push this to after the upcoming v0.2.0 release.

packages/alfa-rules/src/sia-r11/rule.ts
```diff
@@ -48,7 +48,11 @@ export default Rule.Atomic.of<Page, Element>({
...
 export namespace Outcomes {
-  export const HasName = Ok.of("The link has an accessible name");
+  export const HasName = Ok.of(
+    Diagnostic.of(`The link has an accessible name`)
+  );
 
-  export const HasNoName = Err.of("The link does not have an accessible name");
+  export const HasNoName = Err.of(
+    Diagnostic.of(`The link does not have an accessible name`)
+  );
```
Jym77, line 57: The need to pile up `Ok.of(Diagnostic.of(…))` (or `Err.of(Diagnostic.of(…))`) is a bit boilerplating and annoying. But it's not clear we can do much better…

maybe a `foo(outcome: "passed" | "failed", msg, ...args): Result<Diagnostic>` could help, but it's not clear (and it won't really be useful for subclasses of `Diagnostic` anyway)…

kasperisager: > (and it won't really be useful for subclasses of Diagnostic anyway)…

That's precisely why we really can't make this any shorter than it already is. I'm not too worried though to be honest, as I do like the clarity it provides. As mentioned previously, abstraction is a worthy pursuit up to a point, but we shouldn't forget the value of composability.

PR #250, opened 2020-06-05
Implement a SARIF formatter

kasperisager:
Depends on #248.


kasperisager:
After this has been merged we'll push v0.2.0.

PR #256, opened 2020-06-09
Fix background computation in SIA-R69

kasperisager:


packages/alfa-rules/test/common/oracle.ts
```diff
@@ -7,25 +8,34 @@ import { Predicate } from "@siteimprove/alfa-predicate";
...
+    if (question.type === "node" && Option.isOption<Node>(answer)) {
+      return Future.now(Option.of(question.answer(answer)));
+    }
+
+    if (question.type === "node[]" && Iterable.isIterable<Node>(answer)) {
       return Future.now(Option.of(question.answer(answer)));
     }
 
-    if (question.type === "node" && isOption(answer)) {
+    if (question.type === "color" && Option.isOption<RGB>(answer)) {
```
Jym77, line 34: ❤️ 
So that was the trick to make these tests nice…

packages/alfa-rules/test/sia-r69/rule.spec.tsx
```diff
@@ -244,3 +246,63 @@ test("evaluate() is inapplicable to text nodes in disabled groups", async (t) =>
 
   t.deepEqual(await evaluate(R69, { document }), [inapplicable(R69)]);
 });
+
+test("evaluate() passes when a background color with sufficient contrast is input", async (t) => {
```
Jym77, line 250:
```suggestion
test("evaluate() passes when a background image with sufficient contrast is input", async (t) => {
```

packages/alfa-rules/test/sia-r69/rule.spec.tsx
```diff
@@ -244,3 +246,63 @@ test("evaluate() is inapplicable to text nodes in disabled groups", async (t) =>
...
+      passed(R69, target, {
+        1: Outcomes.HasSufficientContrast(21, 4.5, [
+          Contrast.Pairing.of(rgb(0, 0, 0), rgb(1, 1, 1), 21),
+        ]),
+      }),
+    ]
+  );
+});
+
+test("evaluate() fails when a background color with insufficient contrast is input", async (t) => {
```
Jym77, line 280:
```suggestion
test("evaluate() fails when a background image with insufficient contrast is input", async (t) => {
```

packages/alfa-rules/test/sia-r69/rule.spec.tsx
kasperisager: I meant for it to be "color" as that's what's being input.

packages/alfa-rules/test/sia-r69/rule.spec.tsx
kasperisager: Ditto on https://github.com/Siteimprove/alfa/pull/256#discussion_r437426810.

packages/alfa-rules/test/common/oracle.ts
kasperisager: Copy/pasta 😄

PR #258, opened 2020-06-09
Upgrade to Yarn Berry

kasperisager:
I also need to investigate if it's possible for us to move away from Lerna which would remove 600+ transitive dependencies. Lerna is currently only used when publishing packages, which I'm hoping will be possible to do easily with Yarn itself: https://yarnpkg.com/features/release-workflow


kasperisager:
Slating this until we're closer to an answer on https://github.com/yarnpkg/berry/issues/1229#issuecomment-648061910.

kasperisager:
Seeing how we're no closer to a solution on https://github.com/yarnpkg/berry/issues/1229#issuecomment-648061910, we'll stick to Lerna for now.

Jym77 (approved):

PR #264, opened 2020-06-15
Account for presentational role conflict resolution in ARIA feature mappings

Jym77:
On an ``, Presentational role conflict resolution triggered but returns `none` again as it is the implicit role with `alt=""`.

This should fix it.
Role aspect can now take optional arguments.

### DIFF
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -309,7 +309,7 @@ Feature.register(
   Feature.of("img", (element, { allowPresentational }) =>
     Option.of(
       element.attribute("alt").some((alt) => alt.value === "") &&
-        allowPresentational
+        (allowPresentational ?? true)
         ? "presentation"
         : "img"
     )
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -57,10 +57,10 @@ export class Feature<N extends string = string> {
 }
 
 export namespace Feature {
-  export type Aspect<T, U extends unknown = unknown> = Mapper<
+  export type Aspect<T, A extends Array<unknown> = []> = Mapper<
     Element,
     T,
-    Array<U>
+    A
   >;
 
   export interface Status {
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -67,7 +67,10 @@ export namespace Feature {
     readonly obsolete: boolean;
   }
 
-  export interface roleOptions {
+  export interface RoleOptions {
+    /**
+     * @internal
+     */
     readonly allowPresentational?: boolean;
   }
 
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -29,7 +29,7 @@ export class Feature<N extends string = string> {
 
   private constructor(
     name: N,
-    role: Feature.Aspect<Option<string>, Feature.roleOptions>,
+    role: Feature.Aspect<Option<string>, [Feature.RoleOptions]>,
     attributes: Feature.Aspect<Map<string, string>>,
     status: Feature.Status
   ) {
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -10,12 +10,12 @@ import { Scope, Table } from "@siteimprove/alfa-table";
 import { Role } from "./role";
 
 const { hasName, isElement } = Element;
-const { and, equals, test } = Predicate;
+const { and } = Predicate;
 
 export class Feature<N extends string = string> {
   public static of<N extends string>(
     name: N,
-    role: Feature.Aspect<Option<string>> = () => None,
+    role: Feature.Aspect<Option<string>, Role.feature.Options> = () => None,
     attributes: Feature.Aspect<Map<string, string>> = () => Map.empty(),
     status: Feature.Status = { obsolete: false }
   ): Feature<N> {
@@ -23,13 +23,13 @@ export class Feature<N extends string = string> {
   }
 
   private readonly _name: N;
-  private readonly _role: Feature.Aspect<Option<string>>;
+  private readonly _role: Feature.Aspect<Option<string>, Role.feature.Options>;
   private readonly _attributes: Feature.Aspect<Map<string, string>>;
   private readonly _status: Feature.Status;
 
   private constructor(
     name: N,
-    role: Feature.Aspect<Option<string>>,
+    role: Feature.Aspect<Option<string>, Role.feature.Options>,
     attributes: Feature.Aspect<Map<string, string>>,
     status: Feature.Status
   ) {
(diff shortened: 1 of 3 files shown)


packages/alfa-aria/src/feature.ts
```diff
@@ -57,12 +57,20 @@ export class Feature<N extends string = string> {
...
+    Array<U>
+  >;
 
   export interface Status {
     readonly obsolete: boolean;
   }
 
+  export interface roleOptions {
+    readonly allowPresentational?: boolean;
+  }
```
kasperisager, line 72:
```suggestion
  export interface RoleOptions {
    /**
     * @internal
     */
    readonly allowPresentational?: boolean;
  }
```

packages/alfa-aria/src/feature.ts
```diff
@@ -298,9 +306,10 @@ Feature.register(
 
 Feature.register(
   Namespace.HTML,
-  Feature.of("img", (element) =>
+  Feature.of("img", (element, { allowPresentational }) =>
```
kasperisager, line 309:
```suggestion
  Feature.of("img", (element, { allowPresentational = true }) =>
```

packages/alfa-aria/src/feature.ts
```diff
@@ -298,9 +306,10 @@ Feature.register(
...
     Option.of(
-      element.attribute("alt").some((alt) => alt.value === "")
+      element.attribute("alt").some((alt) => alt.value === "") &&
+        (allowPresentational ?? true)
```
kasperisager, line 312:
```suggestion
      allowPresentational && element.attribute("alt").some((alt) => alt.value === "")
```

packages/alfa-aria/src/role.ts
```diff
@@ -280,7 +284,7 @@ export namespace Role {
   }
 
   export namespace from {
-    export interface Options {
+    export interface Options extends Feature.roleOptions {
```
kasperisager, line 287:
```suggestion
    export interface Options extends Feature.RoleOptions {
```

packages/alfa-aria/src/feature.ts
```diff
@@ -10,26 +10,26 @@ import { Scope, Table } from "@siteimprove/alfa-table";
...
 
 const { hasName, isElement } = Element;
-const { and, equals, test } = Predicate;
+const { and } = Predicate;
 
 export class Feature<N extends string = string> {
   public static of<N extends string>(
     name: N,
-    role: Feature.Aspect<Option<string>> = () => None,
+    role: Feature.Aspect<Option<string>, Feature.roleOptions> = () => None,
```
kasperisager, line 18:
```suggestion
    role: Feature.Aspect<Option<string>, [Feature.RoleOptions]> = () => None,
```

packages/alfa-aria/src/feature.ts
```diff
@@ -10,26 +10,26 @@ import { Scope, Table } from "@siteimprove/alfa-table";
...
     attributes: Feature.Aspect<Map<string, string>> = () => Map.empty(),
     status: Feature.Status = { obsolete: false }
   ): Feature<N> {
     return new Feature(name, role, attributes, status);
   }
 
   private readonly _name: N;
-  private readonly _role: Feature.Aspect<Option<string>>;
+  private readonly _role: Feature.Aspect<Option<string>, Feature.roleOptions>;
```
kasperisager, line 26:
```suggestion
  private readonly _role: Feature.Aspect<Option<string>, [Feature.RoleOptions]>;
```

packages/alfa-aria/src/feature.ts
```diff
@@ -10,26 +10,26 @@ import { Scope, Table } from "@siteimprove/alfa-table";
...
   private readonly _attributes: Feature.Aspect<Map<string, string>>;
   private readonly _status: Feature.Status;
 
   private constructor(
     name: N,
-    role: Feature.Aspect<Option<string>>,
+    role: Feature.Aspect<Option<string>, Feature.roleOptions>,
```
kasperisager, line 32:
```suggestion
    role: Feature.Aspect<Option<string>, [Feature.RoleOptions]>,
```

packages/alfa-aria/src/feature.ts
```diff
@@ -43,7 +43,7 @@ export class Feature<N extends string = string> {
     return this._name;
   }
 
-  public get role(): Feature.Aspect<Option<string>> {
+  public get role(): Feature.Aspect<Option<string>, Feature.roleOptions> {
```
kasperisager, line 46:
```suggestion
  public get role(): Feature.Aspect<Option<string>, [Feature.RoleOptions]> {
```

packages/alfa-aria/src/feature.ts
```diff
@@ -57,12 +57,20 @@ export class Feature<N extends string = string> {
 }
 
 export namespace Feature {
-  export type Aspect<T> = Mapper<Element, T>;
+  export type Aspect<T, U extends unknown = unknown> = Mapper<
+    Element,
+    T,
+    Array<U>
+  >;
```
kasperisager, line 64:
```suggestion
  export type Aspect<T, A extends Array<unknown> = []> = Mapper<
    Element,
    T,
    A
  >;
```

This way, multiple additional arguments can be defined if needed.

Jym77: I do not understand the difference… 🤔 
Actually, I do not understand either why we have 
```javascript
export type Mapper<T, U = T, A extends Array<unknown> = Array<unknown>> = (
  value: T,
  ...args: A
) => U;
```
instead of
```javascript
export type MapperBis<T, U = T, A extends unknown = unknown> = (
  value: T,
  ...args: Array<A>
) => U;
```

what are the use-cases allowed by one and not the other?

```javascript
function f(x: number, ...args: Array<string>): boolean { return args.length === x }
```
is both a `Mapper<number, Array<string>, boolean>` and a `MapperBis<number, string, boolean>`.

---
I feel there is something in `A extends Array<unknown>` being different from `Array<X extends unknown>`… Not sure what however…

kasperisager: It may be easiest to illustrate with an example. Let's compare this...

```ts
type P<T> = (...args: Array<T>) => void;
```

...to this:

```ts
type Q<A extends Array<unknown>> = (...args: A) => void;
```

If I want a function that takes _one_ argument of type `number`, `Q<[number]>` would expand to `(a: number) => void`. If I want a function that takes _two_ arguments of type `number` followed by type `string`, `Q<[number, string]>` would expand to `(a: number, b: string) => void`.

In both of those cases, the best I can do with `P` is `P<number>` and `P<number | string>` and both would expand to variadic functions that take `number` and `number |  string` arguments, respectively. `Q` works by relying on _tuples_ rather than arrays and this is how we can control exactly which arguments the expanded function has.

Jym77: > `Q<[number, string]>` would expand to `(a: number, b: string) => void`.

Ah, that's it!
I thought that rest parameters must have the same type and that `function f(...args: [number, string])` wasn't allowed…
Makes sense now! 💡

PR #266, opened 2020-06-16
Prevent infinite instantiation of `Interview` type

kasperisager:
This pull request fixes a compiler crash in TypeScript 3.9 due to the recursive construction of the `Interview` type. This is fixed by limiting the depth of the recursion.

See https://github.com/microsoft/TypeScript/issues/38825.

Before:

https://github.com/Siteimprove/alfa/blob/f77af337f1ea1bc0ef5d100fbd2a06d34d2f32ef/packages/alfa-act/src/interview.ts#L8-L10

After:

https://github.com/Siteimprove/alfa/blob/907a9be4fb50c95307e1352a15fa9c22bfef8d79/packages/alfa-act/src/interview.ts#L16-L27

### DIFF
--- a/packages/alfa-act/src/interview.ts
+++ b/packages/alfa-act/src/interview.ts
@@ -5,9 +5,26 @@ import { Oracle } from "./oracle";
 import { Question } from "./question";
 import { Rule } from "./rule";
 
-export type Interview<Q, S, T> =
+/**
+ * As `Interview` is a recursive type that models nested chains of `Question`s,
+ * we need to limit the depth to which this recursion can happen to avoid the
+ * TypeScript compiler going into an infinite type resolution loop. This is done
+ * by maintaining a pointer, `D extends number` that down to -1 at which point
+ * the recursion will stop. Given a pointer `D`, the `Depths` tuple provides
+ * `D - 1 = Depths[D]`.
+ */
+type Depths = [-1, 0, 1, 2];
+
+export type Interview<Q, S, T, D extends number = 3> =
   | T
-  | { [K in keyof Q]: Question<K, Q[K], S, Interview<Q, S, T>> }[keyof Q];
+  | {
+      [K in keyof Q]: Question<
+        K,
+        Q[K],
+        S,
+        D extends -1 ? T : Interview<Q, S, T, Depths[D]>
+      >;
+    }[keyof Q];
 
 export namespace Interview {
   export function conduct<I, T, Q, A>(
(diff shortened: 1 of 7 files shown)


packages/alfa-act/src/interview.ts
```diff
@@ -5,9 +5,26 @@ import { Oracle } from "./oracle";
...
-export type Interview<Q, S, T> =
+/**
+ * As `Interview` is a recursive type that models nested chains of `Question`s,
+ * we need to limit the depth to which this recursion can happen to avoid the
+ * TypeScript compiler going into an infinite type resolution loop. This is done
+ * by maintaining a pointer, `D extends number` that down to -1 at which point
+ * the recursion will stop. Given a pointer `D`, the `Depths` tuple provides
+ * `D - 1 = Depths[D]`.
+ */
+type Depths = [-1, 0, 1, 2];
```
Jym77, line 16: For the record, this is an ugly trick 🤯 

Also, this means we now have a maximum number of nested questions we can have. Not a huge deal and can easily be augmented anyway…

packages/alfa-rules/src/sia-r23/rule.ts
```diff
@@ -25,53 +25,55 @@ export default Rule.Atomic.of<Page, Element, Question>({
...
-            target,
-            `Where is the transcript of the \`<audio>\` element?`
-          ).map((transcript) => {
-            if (transcript.isNone()) {
-              return Question.of(
-                "transcript-link",
-                "node",
-                target,
-                `Where is the link pointing to the transcript of the \`<audio>\`
+          1: <Interview<Question, Element, Option<Result<Diagnostic>>>>(
```
Jym77, line 28: I don't understand why the typing information is needed for some of the rules but not for all (looks like not all semi-automatic rules are here).

packages/alfa-act/src/interview.ts
kasperisager: The wonders of Turing completeness 🙈 Yes, it's an unfortunate but necessary trick that I snatched straight from the TypeScript compiler itself:

https://github.com/microsoft/TypeScript/blob/4bda7ce8e58e3e6393aff0f9bc23956e0eeb8a4e/lib/lib.es2019.array.d.ts#L21-L26

If we end up in a situation where we need to nest more than 4 levels of questions, we'll adjust the `Depths` type accordingly. 4 levels is the current need.

packages/alfa-rules/src/sia-r23/rule.ts
kasperisager: It's only needed for the cases where multiple different types of questions are nested as the outer question seems to dictate the inference of the rest.

Jym77: This actually makes some sort of sense…

PR #272, opened 2020-06-22
Update SIA R67

Jym77:
Update following
https://github.com/Siteimprove/sanshikan/pull/105


packages/alfa-rules/test/sia-r67/rule.spec.tsx
```diff
@@ -0,0 +1,112 @@
...
+        </svg>
+      </html>,
+      Option.of(self)
+    ),
+  ]);
+  const getById = getElementById(document);
+  const emptyAlt = getById("empty-alt");
+  const roleNone = getById("role-none");
+  const rolePresentation = getById("role-presentation");
+  const svg = getById("svg");
```
kasperisager, line 45: This can be done using destructuring and without additional helpers and need for IDs:

```ts
const [img, video, object, svg] = document
  .first()
  .get()
  .children()
  .filter(Element.isElement);
```

Jym77: True, but this is less resilient to changes in the HTML.
Additionally, ids are commenting (in the HTML) what this target is testing, and can be handy for debugging messages from inside the rule itself.

kasperisager: That's reasonable 👍 What we need then is a unified approach to this rather than ad-hoc helpers scattered throughout rule spec files. Test code is most annoying code to maintain so it needs to be _super_ maintainable 🙈

packages/alfa-aria/src/role.ts
```diff
@@ -261,7 +261,7 @@ export namespace Role {
                     // ...and it's not a presentational role in a forbidden context...
                     !(
                       role.some(Role.isPresentational) &&
-                      !isAllowedPresentational
+                      !allowedPresentational
```
Jym77, line 264: 🙈

packages/alfa-rules/test/sia-r67/rule.spec.tsx
Jym77: Yes, I totally agree. If we decide that IDs + getter is the way we wanna go, we should of course make the helper common and update existing tests to use it…
I've also been considering writing a wrapper for `Document.of`+`Element.fromElement` that we seem to use pretty often in tests…

kasperisager: #281 should hopefully help on the last point.

Jym77: Yeah it gets a bit better. Still need to 
```javascript
Document.fromDocument(
  h.document([
    <html>
      <head>
        <title>Hello world</title>
      </head>
    </html>
  ])
);
```
instead of
```javascript
documentFromHtml(<html>
      <head>
        <title>Hello world</title>
      </head>
    </html>);
```

kasperisager:
Just a heads up, @singingknight, that this is not ready for the platform yet.

PR #273, opened 2020-06-22
Move presentational role conflict resolution down to Role.from

Jym77:
We've built ourselves a nice footgun by putting the conflict resolution in `Node.build` which is bypassed by many functions such as `hasRole` calling `Role.from` directly…

Moving the resolution deeper to where role computation actually happens clean that up. And also avoids `Node.build` doing a double roundtrip to `Role.from`.

Also removing the default value of `isAllowedPresentational` from the individual features to eliminate the risk of them disagreeing on that…

### DIFF
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -57,11 +57,7 @@ export class Feature<N extends string = string> {
 }
 
 export namespace Feature {
-  export type Aspect<T, A extends Array<unknown> = []> = Mapper<
-    Element,
-    T,
-    A
-  >;
+  export type Aspect<T, A extends Array<unknown> = []> = Mapper<Element, T, A>;
 
   export interface Status {
     readonly obsolete: boolean;
@@ -71,7 +67,7 @@ export namespace Feature {
     /**
      * @internal
      */
-    readonly allowPresentational?: boolean;
+    readonly allowPresentational: boolean;
   }
 
   const features = Cache.empty<Namespace, Cache<string, Feature>>();
@@ -309,9 +305,10 @@ Feature.register(
 
 Feature.register(
   Namespace.HTML,
-  Feature.of("img", (element, { allowPresentational = true }) =>
+  Feature.of("img", (element, { allowPresentational }) =>
     Option.of(
-      allowPresentational && element.attribute("alt").some((alt) => alt.value === "")
+      allowPresentational &&
+        element.attribute("alt").some((alt) => alt.value === "")
         ? "presentation"
         : "img"
     )
--- a/packages/alfa-aria/src/node.ts
+++ b/packages/alfa-aria/src/node.ts
@@ -232,9 +232,6 @@ export namespace Node {
           accessibleNode = Branched.of(Container.of(node));
         } else {
           accessibleNode = Role.from(node)
-            .flatMap((role) => {
-              return Branched.of(role);
-            })
             .flatMap<Node>((role) => {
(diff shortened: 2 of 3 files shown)


packages/alfa-aria/src/role.ts
```diff
@@ -251,15 +254,22 @@ export namespace Role {
                   const role = Role.lookup(name);
 
                   if (
+                    // If the role is not abstract…
                     role.some(
                       (role) => role.category !== Role.Category.Abstract
+                    ) &&
+                    // …and it's not a presentational role in a forbidden context…
```
kasperisager, line 261:
```suggestion
                    // ...and it's not a presentational role in a forbidden context…
```

packages/alfa-aria/src/role.ts
```diff
@@ -251,15 +254,22 @@ export namespace Role {
                   const role = Role.lookup(name);
 
                   if (
+                    // If the role is not abstract…
```
kasperisager, line 257:
```suggestion
                    // If the role is not abstract...
```

packages/alfa-aria/src/role.ts
```diff
@@ -251,15 +254,22 @@ export namespace Role {
...
                     role.some(
                       (role) => role.category !== Role.Category.Abstract
+                    ) &&
+                    // …and it's not a presentational role in a forbidden context…
+                    !(
+                      role.some(Role.isPresentational) &&
+                      !isAllowedPresentational
                     )
                   ) {
+                    // …then we got ourselves a valid explicit role…
```
kasperisager, line 267:
```suggestion
                    // ...then we got ourselves a valid explicit role...
```

packages/alfa-aria/src/role.ts
```diff
@@ -251,15 +254,22 @@ export namespace Role {
...
                     return role;
                   }
                 }
               }
-
+              // …otherwise, default to implicit role computation.
```
kasperisager, line 272:
```suggestion
              // ...otherwise, default to implicit role computation.
```

packages/alfa-aria/src/node.ts
```diff
@@ -233,23 +233,10 @@ export namespace Node {
...
-                !isAllowedPresentational(node)
-              ) {
-                return Role.from(node, {
-                  explicit: false,
-                  allowPresentational: false,
-                });
-              }
-
               return Branched.of(role);
             })
```
kasperisager, line 237: This entire `.flatMap()` can be removed now.

packages/alfa-aria/src/role.ts
```diff
@@ -129,6 +129,7 @@ export namespace Role {
   /**
    * @see https://www.w3.org/TR/wai-aria/#roles_categorization
    */
+
```
kasperisager, line 132:
```suggestion
```

packages/alfa-aria/src/feature.ts
```diff
@@ -71,7 +67,7 @@ export namespace Feature {
     /**
      * @internal
      */
-    readonly allowPresentational?: boolean;
+    readonly allowPresentational: boolean;
```
kasperisager, line 70: I now realise that this has caused the public API to change in a way that makes it impossible to _not_ touch internal features from the outside as this internal property is now required 🙈

kasperisager:
I'll fix https://github.com/Siteimprove/alfa/pull/273/files#r448164685 on `master` and relabel this an internal change as that's really what it should have been.

PR #276, opened 2020-06-24
Speedup table building

Jym77:
This should significantly speed up table building by registering which row/column contains data cell and thus avoid going through all cells when computing scope.

More speedup can be achieved on the internal headers scanning. Looking into it.

Also improving code for `update` function with cleaner default parameters initialisation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,7 +39,7 @@ Items that are related, such as breaking changes, new features, or changes to ex
 
 ### Added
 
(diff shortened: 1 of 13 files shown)


Jym77:
Also optimizing the table model errors checks.
Instead of going through all cells and figuring out if another cell is covering the same slot (quadratic in the number of cells), we now just go through all slots and count coverage (linear in the number of cells, but with twice as much storage used).

PR #281, opened 2020-06-26
Introduce HyperScript-like DSL for constructing DOM

kasperisager:
This pull request introduces a [HyperScript](https://github.com/hyperhype/hyperscript)-like DSL of sorts for constructing DOM. It's been in the cards for some time now that we've needed a better approach to constructing DOM, in particular for test cases, as code like the following has become very common practice:

https://github.com/Siteimprove/alfa/blob/a2a909ea04a84615e37cd41dd4c6d9b9de913a47/packages/alfa-rules/test/sia-r1/rule.spec.tsx#L13-L22

The new `h()` function and the associated `h` namespace help alleviate this by allowing code that instead looks like this:

```ts
const document = Document.fromDocument(
  h.document([
    h("html", {}, [
      h("head", {}, [
        h("title", {}, ["Hello world"]),
      ])
    ])
  ])
);
```

This can of course be mixed with JSX as well for an even nicer look:

```tsx
const document = Document.fromDocument(
  h.document([
    <html>
      <head>
        <title>Hello world</title>
      </head>
    </html>
  ])
);
```

This pull request was in particular prompted by the need for also constructing CSSOM for testing the `@siteimprove/alfa-style` package:

```tsx
const document = Document.fromDocument(
  h.document(
    [<div />],
    [
      h.sheet([
        h.rule.style("div", { color: "red" }),
      ]),
    ]
  )
);
```

The changes are breaking as the `jsx()` function has also been modified and now requires that `style` attributes are provided as property-value records to remove the need for declaration parsing which greatly simplifies both the internal code and the exposed interface.

### DIFF
--- a/packages/alfa-aria/test/get-name.spec.tsx
+++ b/packages/alfa-aria/test/get-name.spec.tsx
@@ -35,7 +35,7 @@ test("getName() computes the text alternative of a button with text within a spa
 test("getName() ignores non-visible nodes", (t) => {
   const button = Element.fromElement(
     <button>
-      Button <span style="display: none">Hidden</span>
+      Button <span style={{ display: "none" }}>Hidden</span>
     </button>
   );
 
--- a/packages/alfa-dom/src/h.ts
+++ b/packages/alfa-dom/src/h.ts
@@ -31,58 +31,68 @@ export function h(
   children: Array<Node.JSON | string> = [],
   style: Array<Declaration.JSON> | Record<string, string> = []
 ): Element.JSON {
-  attributes = Array.isArray(attributes)
-    ? attributes
-    : entries(attributes).reduce<Array<Attribute.JSON>>(
-        (attributes, [name, value]) => {
-          if (value === false) {
-            return attributes;
-          }
-
-          return [
-            ...attributes,
-            h.attribute(hyphenate(name), value === true ? "" : value),
-          ];
-        },
-        []
-      );
-
-  style = h.block(style);
-
-  if (style.length > 0) {
-    attributes = [
-      ...attributes,
-      h.attribute("style", Block.fromBlock(style).toString()),
-    ];
-  }
-
-  const json: Element.JSON = {
-    type: "element",
-    namespace: Namespace.HTML,
-    prefix: null,
-    name,
-    attributes,
-    children: children.map((child) =>
-      typeof child === "string" ? h.text(child) : child
-    ),
-    style,
-    shadow: null,
-    content: null,
-  };
-
-  switch (name) {
-    case "svg":
-      rename(json, Namespace.SVG);
-      break;
-
-    case "math":
-      rename(json, Namespace.MathML);
-  }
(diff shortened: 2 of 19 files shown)


packages/alfa-aria/test/get-name.spec.tsx
```diff
@@ -35,7 +35,7 @@ test("getName() computes the text alternative of a button with text within a spa
 test("getName() ignores non-visible nodes", (t) => {
   const button = Element.fromElement(
     <button>
-      Button <span style="display: none">Hidden</span>
+      Button <span style={{ display: "none" }}>Hidden</span>
```
Jym77, line 38: I am a bit concerned by the need for double curly brackets. Notably because that make copy/pasting HTML a bit more difficult.

packages/alfa-dom/src/h.ts
```diff
@@ -0,0 +1,291 @@
...
+      return {
+        type: "supports",
+        condition,
+        rules,
+      };
+    }
+  }
+}
+
+function hyphenate(value: string): string {
```
Jym77, line 279: Should that be called `camelCaseToKebabCase`?

packages/alfa-dom/src/h.ts
```diff
@@ -0,0 +1,291 @@
...
+  return value.replace(/[A-Z]/g, (letter) => `-${letter.toLowerCase()}`);
+}
+
+function rename(element: Element.JSON, namespace: Namespace): void {
```
Jym77, line 283: `rename` is not a very good name as it only renames the namespace.

packages/alfa-aria/test/get-name.spec.tsx
kasperisager: That's how JSX works though. JSX isn't HTML and that's both good and sort of bad. In this case, passing a property-value record greatly simplifies that internal code and makes it easier for consumers to define styles as things aren't stringly typed anymore. In fact, we can improve the typing further by strongly typing the individual properties, though that would be an improvement for the future.

packages/alfa-dom/src/h.ts
kasperisager: That's a terribly long name for a one-line module-internal function 🙈 I would have inlined it but needed to call it twice.

packages/alfa-dom/src/h.ts
kasperisager: Agreed 👍 It's now called `renamespace()`.

packages/alfa-aria/test/get-name.spec.tsx
kasperisager: Also relevant: https://reactjs.org/docs/dom-elements.html#style

Jym77: True. And I have being doing `const style={…}; <div style={style}>` quite a few times… 🤷

PR #284, opened 2020-06-30
SIA R86: decorative element is not exposed

Jym77:
Handling the `disabled` state in presentational conflict resolution ended up in a bit of a quagmire…


packages/alfa-dom/src/node/element.ts
```diff
@@ -352,11 +352,7 @@ export namespace Element {
     );
   }
 
-  export const hasName = predicate.hasName;
-
-  export const hasNamespace = predicate.hasNamespace;
-
-  export const hasId = predicate.hasId;
+  export const { hasId, hasName, hasNamespace, hasTabIndex, isDisabled } = predicate;
```
Jym77, line 355: Is there a way to re-export everything inside a namespace?

packages/alfa-dom/src/node/element.ts
```diff
@@ -370,7 +366,10 @@ function isSuggestedFocusableElement(element: Element): boolean {
       return element.attribute("href").isSome();
 
     case "input":
-      return element.attribute("type").every((attr) => attr.value !== "hidden");
+      return element
+        .attribute("type")
+        .flatMap((attribute) => attribute.enumerate("hidden"))
+        .isNone();
```
Jym77, line 372: Not strictly needed because the attribute value goes through `foldCase`. But that was a bit random coincidence that two transformations just happened to be the same…

packages/alfa-aria/src/role.ts
```diff
@@ -330,21 +329,31 @@ export namespace Role {
...
  */
-const isAllowedPresentational: Predicate<Element> = (element) => {
-  if (element.tabIndex().isSome()) {
-    return false;
-  }
 
-  return Role.lookup("roletype").some((role) => {
+// An element is focusable if it has a tabindex and is not disabled.
+// Note that elements that are not rendered are still considered; however, these are not exposed in the accessibility
+// tree in the first place, so we should be fairly safe here.
```
kasperisager, line 335:
```suggestion
/**
 * An element is focusable if it has a tabindex and is not disabled. Note that 
 * elements that are not rendered are still considered; however, these are not 
 * exposed in the accessibility tree in the first place, so we should be fairly 
 * safe here.
 */
```

packages/alfa-aria/src/role.ts
```diff
@@ -330,21 +329,31 @@ export namespace Role {
...
       }
     }
 
-    return true;
+    return false;
   });
-};
+
+// An element is NOT allowed to be presentational if either
+// it is focusable OR it has some aria-* attributes
```
kasperisager, line 353:
```suggestion
/**
 * An element is NOT allowed to be presentational if either it is focusable OR 
 * it has some aria-* attributes
 */
```

packages/alfa-aria/src/role.ts
kasperisager: Top level declarations should use docblocks to enable intellisense and comments should be wrapped at 80 columns.

packages/alfa-dom/src/node/element.ts
kasperisager: I'm afraid not 😕

packages/alfa-dom/src/node/element/predicate/has-tab-index.ts
```diff
@@ -1,4 +1,4 @@
-import { Element } from "@siteimprove/alfa-dom";
+import { Element } from "../../element";
 import { Predicate } from "@siteimprove/alfa-predicate";
```
kasperisager, line 2:
```suggestion
import { Predicate } from "@siteimprove/alfa-predicate";

import { Element } from "../../element";
```

packages/alfa-rules/src/common/predicate/is-marked-as-decorative.ts
```diff
@@ -0,0 +1,14 @@
+import {Role} from "@siteimprove/alfa-aria";
+import {Element} from "@siteimprove/alfa-dom";
+
+/**
+ * Check if an element is marked as decorative by looking at its role but without conflict resolution.
+ * If the result is "none" or "presentation", then the element is marked as decorative.
+ */
+export function isMarkedAsDecorative(element: Element): boolean {
```
kasperisager, line 8:
```suggestion
export function isDecorative(element: Element): boolean {
```

packages/alfa-rules/src/common/predicate/is-marked-as-decorative.ts
```diff
@@ -0,0 +1,14 @@
+import {Role} from "@siteimprove/alfa-aria";
+import {Element} from "@siteimprove/alfa-dom";
+
+/**
+ * Check if an element is marked as decorative by looking at its role but without conflict resolution.
+ * If the result is "none" or "presentation", then the element is marked as decorative.
+ */
```
kasperisager, line 7:
```suggestion
/**
 * Check if an element is marked as decorative by looking at its role but 
 * without conflict resolution. If the result is "none" or "presentation", then 
 * the element is marked as decorative.
 */
```

packages/alfa-rules/src/common/predicate/is-marked-as-decorative.ts
Jym77: I disagree. This corresponds to our definition of "marked as decorative" and does not check if the element is indeed decorative. It does not even check if the element has a semantic role of none/presentation.

kasperisager: Then I'd call it `isMarkedDecorative()`. `MarkedAs` has an odd ring to it 🤔

kasperisager: `isIntendedDecorative()` is also an option.

PR #287, opened 2020-07-01
Redesign the @siteimprove/alfa-assert package

### LINKED ISSUE #270: Redesign the @siteimprove/alfa-assert package
After having noodled around with the current `@siteimprove/alfa-assert` package and its different offspring, `@siteimprove/alfa-jasmine`, `@siteimprove/alfa-jest`, et. al. it dawned on me that the current approach to accessibility assertions just doesn't work. The approach can be summed up as:

1. Given some input, such as a React node or an Angular test fixture, transform it into a `Page` object.
2. Run all available accessibility rules on the resulting `Page` object.
3. Either:
    - If any of the rules produce a failed outcome, grab the first failed outcome and pass an `Err` result.
    - Otherwise, pass an `Ok` result.
4. If an `Err` result was passed, present it using whatever means are appropriate for the context.

Of these, step 1 is the only one I'm really satisfied with.

Step 2 is problematic because not every rule will make sense to run in every conceivable context. For example, if I'm testing a component in isolation, it doesn't matter if the document has a `<title>` element as I'm not testing an entire document. If the component I'm testing is sufficiently atomic, it also doesn't matter that the component might be missing its required context role because it hasn't yet been put into a larger context. 

Step 3 is problematic as it risks concealing a large portion of failures that won't show up until the currently relevant failure has been fixed. This means re-running all the tests before additional failures become known rather than having a readily available view of _all_ failures.

Step 4 is problematic because it has proven incredibly difficult to present the kind of information we want to present for accessibility issues within the rather constrained test output of the various frameworks and libraries we've prototyped integrations for. Some of them only allow for very short messages, such as "expected target to be accessible", before things get unwieldy.

So, we need an new approach to this entire thing. With the points above in mind, I've identified the following criteria that a solution must meet:

1. The assertion output should be restricted to "something went wrong" or "everything is good" as this ensures we can present it across all the different frameworks and libraries we would want to integrate with. We shouldn't try to present the actual failures as part of the normal test output, which leads to the second point...

2. ...that all failures should be made available externally by relying on the existing formatter mechanism used by the CLI, i.e. `alfa-formatter-*` packages. This way, the actual failures can be presented to users using whatever tools they would normally use for digesting the outcome data.

kasperisager:
Closes #270. Depends on #286.


kasperisager:
This is not _done_ yet, but it's a big step in the right direction. I'd therefore like to include it in the 0.10 release so we can at least start gathering some feedback. @asjohnson mentioned in #671 that they were working on tests using the current Cypress integration and so might have some insights to share.

packages/alfa-assert/src/assertion.ts
```diff
@@ -0,0 +1,91 @@
...
+      .flatMap((outcomes) => {
+        const failures = [...outcomes].filter(Outcome.isFailed);
+
+        const count = failures.length;
+
+        const outcome = count === 1 ? "outcome" : "outcomes";
+
+        const was = count === 1 ? "was" : "were";
+
+        const message = `${count} failed ${outcome} ${was} found`;
```
Jym77, line 59: Looks like "outcome was"/"outcomes were" can be merged in a single string.

Jym77 (approved):
I'm having little knowledge on the internals of the various assertions libs, so this looks good…
Does the main readme need to be updated with newer examples?

kasperisager:
> Does the main readme need to be updated with newer examples?

We don't have examples of the assertion integrations in the README so nothing needs to change there 👍

Jym77 (approved):

PR #288, opened 2020-07-02
Refactor `Scraper#scrape()` internals to fix scraping of URLs with hash fragment

kasperisager:
Previously, scraping a page whose URL contained a hash fragment (`https://example.com/#foo`) would fail for reasons that are still not entirely clear. The internals of the `#scrape()` has grown fairly complicated to deal with immediate client side redirects; this has been simplified and requests are now simply retried with the new URL upon encountering a redirect.

### DIFF
--- a/packages/alfa-scraper/src/scraper.ts
+++ b/packages/alfa-scraper/src/scraper.ts
@@ -127,7 +127,9 @@ export class Scraper {
             .catch(() => null)
             .then((response) => response!);
 
-          const request = response.then((response) => response.request());
+          const request = response
+            .then((response) => response.request())
+            .catch(() => null);
 
           const result = await awaiter(page, timeout);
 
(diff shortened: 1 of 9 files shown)


packages/alfa-scraper/src/scraper.ts
```diff
@@ -36,6 +36,17 @@ export class Scraper {
     return new Scraper(await browser);
   }
 
+  public static async with<T>(
+    mapper: Mapper<Scraper, Promise<T>>,
+    browser?: Promise<puppeteer.Browser>
+  ): Promise<T> {
+    const scraper = await this.of(browser);
```
Jym77, line 43: Using `this` inside a static method triggered all my alarms… Then I realised that `of` is also static…
```suggestion
    const scraper = await Scraper.of(browser);
```
🤔 hmm… unless we do need to trickle down class hierarchy and call the correct `of`…

kasperisager: While the class has a private constructor and therefore cannot be subclassed, we _could_ call `Scraper.of()` just fine. But, for the sake of consistency with classes that _can_ be subclassed, using `this` makes sure we're always calling the right class.

Issue #293, opened 2020-07-10
Implement support for CSS Custom Properties


kasperisager:
Alfa currently lacks support for CSS Custom Properties: https://drafts.csswg.org/css-variables/. Implementing this will require a good amount of consideration as we currently diverge somewhat from typical UAs in our implementation of cascade due to our analytical use case.

PR #294, opened 2020-07-12
Very partial implementation of CSS custom properties

Jym77:
This is a very partial implementation of CSS custom properties.
In addition to the numerous `@TODO` that need to be solved, this should work for other values than colours…

The easy way would have been to simply keep any string containing `var(`, do the substitution at string level (using regexp or whatnot), _et voilà_. However, this is incorrect as the substitution happens at CSS token level, hence `--two: 2; width: var(--two)px` is **not** syntactically correct CSS but would be fine with low-level string substitution (https://drafts.csswg.org/css-variables/#using-variables)

My idea moving forward would be:
* Have a parser for `<declaration-value>` (https://drafts.csswg.org/css-syntax-3/#typedef-declaration-value).
* Use it to create a correct parser for `var()` syntax. This might be super annoying as a `var()` can contain nested `var()` in the fallback value, so recursive parser is needed… This might also lay the foundation of `attr()` and `calc()` parsers.
* All value parsers (in `alfa-css/src/value/*`) should parse `either(currentParser, varParser)` since a `var()` can be used in place of any value (as far as I understand). Probably need some clever workaround to avoid boiler-plate code that can easily be forgotten…
* Composite values (e.g. `rgb(R, G, B)`) that contain a `var()` are "tainted" (whether the `var()` is the full value or just a component). Once parsing is done to check basic syntax, we have to assume that the substitution is OK (https://drafts.csswg.org/css-variables/#using-variables just after Example 11), so we only keep the full string (as currently done in `alfa-css/src/value/color/var.ts`), or maybe some sort of AST (not really needed, as parsing will occur again anyway).
* At substitution time, "tainted" values are substituted (the AST might help in substituting only the `var()` part) and re-parsed (they can be invalid at computed-value time).

Relates to #293


kasperisager:
I appreciate the experimentation with this, though this is not the design that I've got in mind. Inspired by Servo, the key point to a solid implementation of this is that __no__ property definition should require awareness of the existence of custom properties and neither should callers of our APIs. As you've hinted at yourself, all the `var()` function does is perform token substitution. Each property definition is responsible for determining how to form a valid value from a list of tokens and so this substitution should take place _before_ we consult the property definition. Fortunately, our current design of the style system aligns perfectly with this; the property definitions are only consulted _just prior_ to computing styles for an individual element, which puts us in a perfect position to resolve custom properties and perform `var()` substitution right then and there.

I've still got some pieces I need to figure out and will do so first thing when I'm back from vacation.

kasperisager:
These changes are mostly superseded by #295.

PR #299, opened 2020-07-30
Implement support for ACT rule requirements and tags

### LINKED ISSUE #180: Explore use cases of including requirements in rules
When an ACT rule tests conformance to accessibility requirements, these requirements will be listed in the rule: https://www.w3.org/TR/act-rules-format/#accessibility-requirements-mapping. While we have previously included these requirements in the rule implementations in Alfa, they are no longer included following #165 as the use cases for them were not clear and their design very much reflected this.

As such, I'd like for us to explore possible use cases for including requirements in rules. As a client of Alfa, I will currently need to know about the requirements underlying a specific rule if I want to account for these requirements in my use of Alfa. One case where I may want to account for this would be in setting a bar for WCAG conformance. For example, I may wish to include all rules that test for success criteria required for level AA conformance, present and future. In order to do this today, I will have to cherry pick rules based on my knowledge of which ones test for success criteria required for level AA conformance.

Other use cases are likely to exist as well. If you come across one, please do share it by leaving a comment.

### LINKED ISSUE #236: Investigate ways of "tagging" rules
Possibly in extension of #180, it would be great with a generic way of assigning additional metadata to rules. Currently, it requires oracular knowledge to put together a set of rules for a given purpose, such as auditing a page or running a unit test on an isolated component. Something like the following would make it possible to select a subset of rules based on tags:

```ts
const rules = [
  Rule.Atomic.of<I, T>({
    uri: "foo",
    tags: [
      Context.of("page")
    ],
    evaluate(...) {}
  }),

  Rule.Atomic.of<I, T>({
    uri: "bar",
    tags: [
      Context.of("component")
    ],
    evaluate(...) {}
  }),
];

const componentRules = rules.filter(rule => 
  rule.tags.some(tag => Context.isContext(tag, "component")
);
```

kasperisager:
Closes #180.
Closes #236.


kasperisager:
@Jym77 I'll let you do an initial review of this when you're available, before I trek through the rest of the ~70 rules and add requirements.

packages/alfa-wcag/scripts/techniques.js
```diff
@@ -0,0 +1,49 @@
+const fs = require("fs");
+const path = require("path");
+const prettier = require("prettier");
+const puppeteer = require("puppeteer");
+
+puppeteer.launch().then(async (browser) => {
+  const page = await browser.newPage();
+
+  await page.goto("https://www.w3.org/WAI/WCAG21/Techniques/");
```
Jym77, line 9: I guess we might need 2.2 techniques at some point 🤔
If there's no technique that disappeared, we should probably only grab 2.2 techniques.

Jym77 (approved):

packages/alfa-wcag/scripts/techniques.js
kasperisager: The WCAG 2.2 techniques are currently the same as the 2.1 techniques, sans all the Flash techniques: https://www.w3.org/WAI/WCAG22/Techniques/

Jym77 (approved):

Jym77 (commented):
Looks good. Seems there is no tags in rules currently, but I guess that's on purpose and need refinement to decide which tags to use.

Jym77 (approved):

PR #301, opened 2020-08-04
Implement remaining `background-*` CSS long- and shorthand properties

### LINKED ISSUE #285: Implement remainder of the `background` CSS shorthand property grammar
While #277 introduced initial support for the `background` shorthand property, it did do with fairly limited scope and so the property currently only supports a single `<color>` layer.

kasperisager:
This pull request implements the remaining `background-*` long- and shorthand properties. It also introduces a shared, abstract `Value` class that captures some commonalities between the different value types.

The changes introduced are breaking as the public API of some CSS value types have either seen properties renamed or removed. The parsed type of property definitions has also changed to simplify things.

Closes #285.


packages/alfa-style/src/property/background.ts
```diff
@@ -70,45 +126,712 @@ export namespace Background {
...
+        | ["background-repeat-y", Repeat.Y.Specified.Item]
+        | ["background-attachment", Attachment.Specified.Item]
+        | ["background-origin", Origin.Specified.Item]
+        | ["background-clip", Clip.Specified.Item]
+      >,
+      string
+    >(
+      map(delimited(option(Token.parseWhitespace), parseColor), (color) => [
+        ["background-color", color],
+      ]),
```
Jym77, line 633: I think `parseBackgroundLayer` would be more readable by having the various components of `any` be extracted and named individually. That function is huge and reading parser combinators is not easy…

packages/alfa-style/src/property/background.ts
```diff
@@ -70,45 +126,712 @@ export namespace Background {
...
+        switch (position[0]) {
+          case "x":
+            x = position[1];
+            break;
+          case "y":
+            y = position[1];
+        }
+      }
+
+      return [x, y];
```
Jym77, line 332: I feel this could be streamlined in (roughly)
```javascript
const result = {x: "center"; y: "center"};
for (const [key, value] of positions) {
  result[key] = value;
}

return [result.x, result.y];
```
Not sure it is that much more readable in that specific case, but it does emphasise the similarity in treatment and the fact that `positions` is an array of `[key, value]`.

packages/alfa-style/src/property/background.ts
```diff
@@ -70,45 +126,712 @@ export namespace Background {
...
+              break;
+            case "background-attachment":
+              attachment.push(property[1]);
+              break;
+            case "background-origin":
+              origin.push(property[1]);
+              break;
+            case "background-clip":
+              clip.push(property[1]);
+          }
```
Jym77, line 819: This also feels very redundant and could maybe use
```javascript
for (const [key, value] of layer) {
  result[key].push(value)
}
```
(also need to replace `color` by an array and take its last value).

packages/alfa-style/src/property/combinator/any.ts
```diff
@@ -0,0 +1,44 @@
+import { Parser } from "@siteimprove/alfa-parser";
+import { Err, Result } from "@siteimprove/alfa-result";
+
+export function any<I, T, E, A extends [] = []>(
```
Jym77, line 4: Should that be part of `alfa-parser` instead? It feels like it should be together with the other parser combinators (`either`, …)

Maybe need to break `alfa-parser/src/parser.ts` into specific files for individual combinators (or related ones, e.g. `pair`, `left`, `right` make sense together).

Jym77, line 4: I had a hard time figuring out what this function does (parse a list of components where each component is present zero or one time, in an unspecified order).
I feel the name doesn't reflect that, but it might be a standard combinator name (couldn't find ref…) Plus, I don't really have anything else to suggest…
(maybe just a nice docstring…)

Jym77, line 4: Also, given that it's a tricky combinator (both to write and to understand, notably trickier than the ones we already have) it would benefits from some tests 😄

packages/alfa-style/src/property/combinator/any.ts
```diff
@@ -0,0 +1,44 @@
...
+  parser: Parser<I, T, E, A>,
+  ...rest: Array<Parser<I, T, E, A>>
+): Parser<I, Iterable<T>, E, A> {
+  const parsers = [parser, ...rest];
+
+  return (input, ...args) => {
+    const seen = parsers.map(() => false);
+    const values: Array<T> = [];
+
+    outer: while (true) {
```
Jym77, line 14: I think we can get rid of the `outer` loop (and the awful goto) simply be replacing `continue outer` with `i=0` (and maybe go all the way and replace the inner `for` by a real `while` in that case…)

packages/alfa-style/src/property/background.ts
kasperisager: That however messes up the type inference as `key` will be inferred as `"x" | "y"` and `value` will be inferred as `X.Specified.Item | Y.Specified.Item`. We need `"x"` to be assigned a value of `X.Specified.Item` and `"y"` a value of `Y.Specified.Item` which is only possible if we let TypeScript narrow the tuple type before we destruct it, as currently done.

packages/alfa-style/src/property/background.ts
kasperisager: That's not really possible for the same reason as https://github.com/Siteimprove/alfa/pull/301/files/ac71fefe5f47951cec8c139a6a31ecef8546ddc8#r468449936. In Servo, this stuff is generated via Python templating to avoid having to spell it out while still appeasing the type checker 🙈

packages/alfa-style/src/property/combinator/any.ts
kasperisager: It's a very CSS-specific combinator and so shouldn't live the generic combinators.

kasperisager: I'll add a `@see` annotation linking to https://drafts.csswg.org/css-values/#comb-any. The behaviour of this combinator is indeed a little odd, hence my comment in https://github.com/Siteimprove/alfa/pull/301#discussion_r468452732.

packages/alfa-style/src/property/combinator/any.ts
kasperisager: That won't work I'm afraid. We want to delay evaluating lines 37-41 until no results were found for any of the parsers. Every time a result is found, the process restarts by skipping the rest of the outer loop.

This is another piece of code that could be improved by some kind of "reduce with the ability to break early" as we could then reduce the array of parsers to either an error or the _first_ result, after which we'd break the reduction. For lack of that, I'm fine with some local side effects and a labelled continue.

packages/alfa-style/src/property/background.ts
kasperisager: I've added some comments that names the different components. The issue with the individual parsers out of the background layer parser is that they're really specialised versions of the individual component parsers. For example, `parseColor` already exists and calling the one used the background layer parser `parseColorInBackgroundLayer` or something similar seems like a bad idea.

packages/alfa-style/src/property/combinator/any.ts
Jym77: 🤔 but lines 37-41 would be after that loop…
```javascript
let i=0;
while (i<n) {
  if (seen[i]) continue; // skip parsers for components seen during a previous pass

  const result = parser[i](input); // try a new parser

  if (result.isErr()) { // parser is failing
    error = result;
    i++; // move on to next parser, within the same pass.
  } else {
    const [remainder, value] = result.get();
    values.push(value);
    input = remainder;
    seen[i] = true; // don't redo that parser during further passes.

    i = 0; // restarts a new pass from the start of the list of parsers.
  }
}
// we're only exiting the while if, during the last pass, each parser has either been seen or failed.

return values.length === 0 ? 
  error! // no parser has ever succeeded, returning last error found.
  : Result.of([input, values]); // some parsers have succeeded, returning result.

```

packages/alfa-style/src/property/background.ts
Jym77: Ah, yes. Now I see the structure of that parser a bit better 🙈 It's already just a bunch of wrappers around existing parser to add the `[key, value]` structure that is needed…
I guess the comments help 💯

packages/alfa-style/src/property/combinator/any.ts
kasperisager: Ah, now I see! That would indeed work 👍

PR #306, opened 2020-08-21
Implement explicit `aria-owns` relations in the accessibility tree

### LINKED ISSUE #173: Implement explicit `aria-owns` relations in the accessibility tree
The accessibility tree does not yet consider explicit relations between nodes made using the `aria-owns` attribute. This will have to be implemented in the tree construction hereabouts: 

https://github.com/Siteimprove/alfa/blob/40c81310ab959df019e4e2f8f603135f63e2aa22/packages/alfa-aria/src/node.ts#L219-L230

kasperisager:
Closes #173.

### DIFF
--- a/packages/alfa-aria/src/node.ts
+++ b/packages/alfa-aria/src/node.ts
@@ -8,6 +8,7 @@ import { Map } from "@siteimprove/alfa-map";
 import { None, Option } from "@siteimprove/alfa-option";
 import { Predicate } from "@siteimprove/alfa-predicate";
 import { Sequence } from "@siteimprove/alfa-sequence";
+import { Set } from "@siteimprove/alfa-set";
 import { Style } from "@siteimprove/alfa-style";
 
 import * as dom from "@siteimprove/alfa-dom";
@@ -20,7 +21,7 @@ import { Feature } from "./feature";
 
 import * as predicate from "./node/predicate";
 
-const { and, not, nor, equals } = Predicate;
+const { equals } = Predicate;
 
 /**
  * @see https://w3c.github.io/aria/#accessibility_tree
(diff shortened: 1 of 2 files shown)


packages/alfa-aria/src/node.ts
```diff
@@ -222,7 +223,65 @@ export namespace Node {
...
+    // https://w3c.github.io/aria/#aria-owns
+
+    // Find all elements in the tree. As explicit ownership is specified via ID
+    // references, it cannot cross shadow or document boundaries.
+    const elements = root.inclusiveDescendants().filter(dom.Element.isElement);
+
+    // Build a map from ID -> element to allow fast resolution of ID references.
+    // The collected references are added to the map in reverse order to ensure
+    // that the first occurrence of a given ID is what ends up in the map in
+    // event of duplicates.
```
Jym77, line 237: I'm wondering whether this map should be made a somewhat global object (probably attached to the `Document`) since there are many case where we need to access elements by ID… Having this available easily could speed things a bit.

packages/alfa-aria/src/node.ts
```diff
@@ -222,7 +223,65 @@ export namespace Node {
...
+              .reject((reference) => reference === element),
+          ] as const
+      )
+    );
+
+    // Refine the collected `aria-owns` references, constructing a set of
+    // claimed elements and resolving conflicting claims as needed.
+    const [claimed, owned] = references.reduce(
+      ([claimed, owned], [element, references]) => {
+        // Reject all element references that have already been claimed.
```
Jym77, line 269: As far as I've understood, specs don't say how to resolve multiple claim and we are making a choice to give them to the first claimer.
I guess we should document that this is our choice and a spec ambiguity so that it's easier to find and fix if specs are fixed…

packages/alfa-aria/src/node.ts
```diff
@@ -222,7 +223,65 @@ export namespace Node {
...
+
+        return [claimed, owned.set(element, references)];
+      },
+      [Set.empty<dom.Node>(), Map.empty<dom.Element, Sequence<dom.Node>>()]
+    );
+
+    build(root, device, claimed, owned);
 
     if (_cache.has(node)) {
       return _cache.get(node).get();
```
Jym77, line 292: Shouldn't we look in the cache first? If the node is already there, no need to process `aria-owns` again…

packages/alfa-aria/src/node.ts
```diff
@@ -222,7 +223,70 @@ export namespace Node {
...
+    const elements = root.inclusiveDescendants().filter(dom.Element.isElement);
+
+    // Build a map from ID -> element to allow fast resolution of ID references.
+    // The collected references are added to the map in reverse order to ensure
+    // that the first occurrence of a given ID is what ends up in the map in
+    // event of duplicates.
+    const ids = Map.from(
+      elements
+        .collect((element) => element.id.map((id) => [id, element] as const))
+        .reverse()
```
Jym77, line 241: Got to point out that this is going through the list twice while the first pass could put stuff to a LIFO accumulator to do the `reverse` at the same time. Not a big deal, I guess.

Jym77 (approved):

packages/alfa-aria/src/node.ts
kasperisager: D'oh 🤦

packages/alfa-aria/src/node.ts
kasperisager: This is specifically left as undefined behaviour:

> Authors **MUST** ensure that an element's ID is not specified in more than one other element's `aria-owns` attribute at any time. In other words, an element can have only one explicit owner.

I'll add an explanatory comment 👍

packages/alfa-aria/src/node.ts
kasperisager: Fortunately, it's not a list; it's a lazy sequence. All `.reverse()` does is therefore force the sequence in reverse order, applying the `#collect()` step at every value encountered. Now, `Map.from()` also needs to force the sequence and, as it's already forced, can simply traverse the existing sequence as a linked list. If we want to avoid both forcing _and_ separately traversing the sequence, `#collect()` could be followed by a `#reduce()` that builds the map and skips duplicate entries:


```ts
const ids = elements
  .collect((element) => element.id.map((id) => [id, element] as const))
  .reduce(
    (map, [id, element]) => (map.has(id) ? map : map.set(id, element)),
    Map.empty<string, dom.Element>()
  );
```

I don't know if any potential benefits outweigh the reduced clarity though 🤔

packages/alfa-aria/src/node.ts
kasperisager: Oh wait, I forgot that it already does 🤦 https://github.com/Siteimprove/alfa/blob/7d16fa50924bdf67db8ae53c878a27ad99e6223b/packages/alfa-aria/src/node.ts#L218-L222

kasperisager: We should however definitely cache the `Inert` node returned if the node isn't even part of the tree.

packages/alfa-aria/src/node.ts
kasperisager: I've thought about this as well, though I don't yet have a good answer on how to best go about it. I don't want to expose these kinds of lookup tables in public API as that feels like an awfully leaky abstraction. I'd also like to avoid APIs similar to [`Document.getElementById()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementById) as that also feels leaky. I'd much rather we come up with a general query API that uses structured queries that we can then optimise behind the scenes. That is, rather than have N different lookup tables to improve various queries, we have a single query API that we can optimise as needed to ensure a balanced trade-off between time and memory.

Jym77 (approved):

PR #309, opened 2020-08-27
Implement minimal crawler based on the existing scraper

kasperisager:
I've had this sitting in my local tree for a while now and wanted to ship it off before it got too stale. We've heard from several potential users that their use cases complicate passing a list of known URLs to either the scraper or the CLI. This pull request therefore introduces a _very_ simple crawler capable of producing a stream of `Page` objects for a given URL scope. This allows users to point the crawler at a list of seed URLs and then have the crawler discover and scrape pages. 

To use the crawler, a `Frontier` instance is first constructed with a URL scope and an optional list of seed URLs:

```ts
import { Frontier } from "@siteimprove/alfa-frontier";

const frontier = Frontier.of(
  // The URL scope. Only URLs that start with this URL are accepted by the
  // frontier when enqueuing URLs.
  "https://example.com",

  // The list of seed URLs which are immediately added to the frontier. If not
  // specified, the URL scope is used as the seed.
  ["https://example.com/foo.html", "https://example.com/bar.html"]
);
```

With a `Frontier` constructed, we can construct a `Crawler` instance and have it crawl the frontier:

```ts
import { Crawler } from "@siteimprove/alfa-crawler";

Crawler.with(async (crawler) => {
  for await (const result of crawler.crawl(frontier)) {
    if (result.isErr()) {
      console.error(result.getErr());
    } else {
      const page = result.get();
    }
  }
});
```

With these two building blocks, it should be easy for users to put together scripts that dynamically discover and audit pages. The question still stands on how to best support this functionality in the CLI, though that's a topic for another time.


packages/alfa-frontier/test/frontier.spec.ts
```diff
@@ -0,0 +1,120 @@
...
+  t.deepEqual(frontier.toJSON(), {
+    scope: "https://example.com/",
+    items: [],
+  });
+});
+
+test("#enqueue() adds a URL to the frontier", (t) => {
+  const frontier = Frontier.of("https://example.com/");
+
+  frontier.enqueue("https://example.com/foo.html");
```
Jym77, line 51: This (and next test) should also check the return value of `enqueue`.

packages/alfa-frontier/src/frontier.ts
```diff
@@ -0,0 +1,370 @@
...
+import { Option, None } from "@siteimprove/alfa-option";
+import { Predicate } from "@siteimprove/alfa-predicate";
+
+import * as json from "@siteimprove/alfa-json";
+
+const { equals, property } = Predicate;
+
+export class Frontier implements Equatable, Serializable {
+  public static of(
+    scope: string | URL,
```
Jym77, line 12: I can see the use case where the scope is more than one domain (URL). E.g. wantirg to crawl `example1.com` and `example.com`.

Jym77, line 12: Actually, `Frontier` and `Item` are pretty close to being a generic queue management system, not just for URLs.
Need to add a type parameter to both, and change `scope` to a `T => bool`. The rest of `Frontier` is pretty much agnostic to the actual stuff in it, and it looks like `Item` only need to know that it's URLs for its `matches` method which could as well be avoided by requiring `T` to be `Equatable`.

Jym77 (approved):
Looks good. Some minor possible improvements.

packages/alfa-frontier/src/frontier.ts
kasperisager: It's really not a generic queue of any sorts though as every operation has semantics specific to a crawl frontier. Sure, you might be able to abstract over the type of things stored in the frontier, though I don't see any use for it as its only purpose is to act as a URL frontier. Any attempt to abstract it beyond that seem like needless complexity 🤔

kasperisager: That seems like a quick way to explode the scope though, no pun intended. For example, one might want to use a single frontier to be scoped to subdomains as well, so then we need to support glob patterns like `https://*.example.com`.

I'd prefer to keep it simple and limit every frontier to a single URL prefix as its scope. It multiple distinct scopes are needed, multiple frontiers must be used.

kasperisager: It's also work keeping in mind that this is a structure that callers will interact with directly and so any additional layers of indirection we add will be layers that callers will have to unwind.

Jym77 (approved):

PR #355, opened 2020-09-04
Fix for `Lazy#map()` and `Lazy#flatMap()` possibly evaluating a `Lazy` multiple times

kasperisager:


packages/alfa-lazy/test/lazy.spec.ts
```diff
@@ -0,0 +1,85 @@
...
+  Lazy.of(() => {
+    throw new Error("The lazy was evaluated");
+  });
+});
+
+test("#force() evaluates a lazy value", (t) => {
+  t.equal(lazy.force(), 42);
+});
+
+test("#force() only evaluates a lazy value once, even if valled multiple times", (t) => {
```
Jym77, line 21:
```suggestion
test("#force() only evaluates a lazy value once, even if called multiple times", (t) => {
```

Jym77 (approved):
I'm not familiar enough with the implementation of `Lazy` to really get all that, but at least I'm buying in the tests and stuff does what it should do 😀

Jym77 (approved):

PR #356, opened 2020-09-04
Improve ARIA `Node` construction performance

kasperisager:
This pull request improves the construction performance of ARIA `Node` instances by introducing the following changes:

- When computing names for form controls with connected `<label>` elements, the list of `<label>` element is re-used, rather than re-computed, for each document scope.

- The `State` interface used for encoding the name computation state is now a class whose methods attempt to re-use the current `State` instance whenever possible to reduce allocations and pressure on the garbage collector.

- The attributes supported by `Role` instances are now computed and stored on construction.

- Elements without ARIA attributes and roles are now included as `Container` instances, hence skipping an expensive name computation as these elements shouldn't be included in the accessibility tree.

The last point also revealed an issue in the `isPerceivable()` predicate as it wasn't correctly accounting for elements that weren't included in the accessibility tree themselves but had children that were.

### DIFF
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -174,13 +174,13 @@ const labels = Cache.empty<Node, Sequence<Element>>();
 const nameFromLabel = (element: Element, device: Device, state: Name.State) => {
   const root = element.root();
 
+  const elements = root.inclusiveDescendants().filter(isElement);
+
   for (const id of element.id) {
     const target = ids
       .get(root, () =>
         Map.from(
-          root
-            .inclusiveDescendants()
-            .filter(isElement)
+          elements
             .collect((element) =>
               element.id.map((id) => [id, element] as const)
             )
@@ -197,9 +197,7 @@ const nameFromLabel = (element: Element, device: Device, state: Name.State) => {
   }
 
   const targets = labels
-    .get(root, () =>
-      root.inclusiveDescendants().filter(and(isElement, hasName("label")))
-    )
+    .get(root, () => elements.filter(hasName("label")))
     .filter(
       or(
         (label) => label.descendants().includes(element),
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -1,11 +1,14 @@
 import { Branched } from "@siteimprove/alfa-branched";
+import { Cache } from "@siteimprove/alfa-cache";
 import { Browser } from "@siteimprove/alfa-compatibility";
 import { Device } from "@siteimprove/alfa-device";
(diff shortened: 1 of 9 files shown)


packages/alfa-aria/src/feature.ts
```diff
@@ -167,13 +167,27 @@ const nameFromChild = (predicate: Predicate<Element>) => (
...
 };
 
+const ids = Cache.empty<Node, Map<string, Element>>();
+
+const labels = Cache.empty<Node, Sequence<Element>>();
+
 const nameFromLabel = (element: Element, device: Device, state: Name.State) => {
   const root = element.root();
 
+  const elements = root.inclusiveDescendants().filter(isElement);
```
Jym77, line 177: I think this should either be cached (like `ids` and `labels`) or rather just skip the computation if we `ids.has(root) && labels.has(root)` since it seems that `elements` is only used in these cache misses.

packages/alfa-aria/src/name.ts
```diff
@@ -336,48 +337,125 @@ export namespace Name {
...
+    public hasVisited(element: Element): boolean {
+      return this._visited.has(element);
+    }
+
+    public visit(element: Element): State {
+      const visited = this._visited.add(element);
+
+      if (this._visited === visited) {
+        return this;
+      }
```
Jym77, line 404: Could get the same logic with probably less computation with
```javascript
if (this.hasVisited(element)) {
  return this;
}
…
```
(would also make the logic more obvious, I feel: if it's already in, do nothing; otherwise, add it.

Jym77 (changes requested):

packages/alfa-aria/src/feature.ts
kasperisager: It's a lazy sequence so without enumerating it nothing is actually computed.

packages/alfa-aria/src/name.ts
kasperisager: That would require doing _two_ lookups though: First to check if the element is present, and second to add the element if it isn't. I'll add an explanatory comment though.

Jym77 (approved):

packages/alfa-aria/src/name.ts
Jym77: I guess the point I misjudged is the cost of `===` on complex structures (set, map, …)
I naively assumed it needed to compare content (which would also require lookup).

kasperisager: Ah, I see! Equality checks for object types are solely based on reference identity, not structure, and so will always be `O(1)`, yes. Structural equality will however be introduced for the proposed record and tuple types, currently at stage 2: https://github.com/tc39/proposal-record-tuple

PR #360, opened 2020-09-08
Remove step 2D of the accessible name computation

kasperisager:
Step 2D of the name computation causes a combinatorial explosion for at least one input and has therefore been removed. Its removal has had no effect on the current test cases and so if we do find a case actually covered by this step, we'll deal with that if and when it becomes relevant.

I've also included a small optimisation to how visited nodes are stored. These are now kept in an array rather than a set and a linear scan is performed when testing if a node has already been visited. As it turns out, only very few nodes are usually visited during name computation and so the overhead of set operations becomes more expensive than just using an array.


Jym77 (approved):

kasperisager:
Correction: The step in question was 2H, not 2D.

PR #362, opened 2020-09-08
Expose elements in accessibility tree if they have a tab index

Jym77:
Nodes that have a tab index but no role or ARIA attribute were not exposed in the accessibility tree.  Now they are.

This notably excluded `<iframe>` from the accessibility tree, thus making most of them inapplicable for their relevant rules…


kasperisager (approved):
💯

PR #367, opened 2020-09-09
Implement support for tree-structural pseudo-classes

### LINKED ISSUE #57: Implement support for An+B microsyntax

### LINKED ISSUE #134: Implement selector matching for child index selectors
~~With support for An+B microsyntax introduced in #131~~, child index selectors such as `:nth-child()` will also have to be supported during selector matching. Look into existing browser implementations for hints on optimizations, if any are employed.

kasperisager:
https://drafts.csswg.org/selectors/#structural-pseudos

Closes #57.
Closes #134.

### DIFF
--- a/packages/alfa-css/src/syntax.ts
+++ b/packages/alfa-css/src/syntax.ts
@@ -3,4 +3,5 @@ export * from "./syntax/component";
 export * from "./syntax/declaration";
 export * from "./syntax/function";
 export * from "./syntax/lexer";
+export * from "./syntax/nth";
 export * from "./syntax/token";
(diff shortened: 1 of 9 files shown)


kasperisager:
I'm adding some test cases for the `an+b` parser as well.

Jym77 (approved):
`selector.ts` is close to 2000 lines 😮 Which does raise the question of splitting it into mindsized bits soonish…

kasperisager:
We need to be careful as there are circular dependencies between the various selectors, so splitting out into several modules may not be straight forward.

PR #371, opened 2020-09-10
Fix `separator` role handling and attribute retrieval in R16

### LINKED ISSUE #366: Differentiate between static and focusable separators
The `separator` role (https://w3c.github.io/aria/#separator) has different requirements depending on whether the element carrying the role is focusable or not. There might not be a good way to encode this within the current structure of `Role` instances, so this may be something we should cater to at the rule level instead.

kasperisager:
Closes #366.

### DIFF
--- a/packages/alfa-rules/test/sia-r16/rule.spec.tsx
+++ b/packages/alfa-rules/test/sia-r16/rule.spec.tsx
@@ -1,0 +1,85 @@
+import { jsx } from "@siteimprove/alfa-dom/jsx";
+import { test } from "@siteimprove/alfa-test";
+
+import { Document } from "@siteimprove/alfa-dom";
+
+import R16, { Outcomes } from "../../src/sia-r16/rule";
+
+import { evaluate } from "../common/evaluate";
+import { passed, failed, inapplicable } from "../common/outcome";
+
+test(`evaluate() passes a <div> element with a role of checkbox and an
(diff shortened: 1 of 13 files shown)


packages/alfa-rules/test/sia-r16/rule.spec.tsx
```diff
@@ -0,0 +1,85 @@
...
+  const document = Document.of([checkbox]);
+
+  t.deepEqual(await evaluate(R16, { document }), [
+    failed(R16, checkbox, {
+      1: Outcomes.HasNotAllStates,
+    }),
+  ]);
+});
+
+test(`evaluate() failed a focusable <div> element with a role of separator and
```
Jym77, line 74: I think we should add a passed case with a non-focusable separator and no `aria-valuenow` since this is the problem we're fixing.

Jym77 (changes requested):

packages/alfa-rules/test/sia-r16/rule.spec.tsx
kasperisager: That's essentially the `<hr>` case, but I'll add another case with an explicit `separator` role.

Jym77 (approved):
Arf, failed to read the `<hr>` case correctly 😞 Sorry…

kasperisager:
No worries! Having both the implicit and explicit case covered is better 👍

Issue #138, opened 2019-08-20
Implement support for `calc()` expressions in CSS


kasperisager:
https://drafts.csswg.org/css-values/#calc-notation

kasperisager:
We're seeing false positives due to lack of support for `calc()` expressions so I'll up the priority on this one.

PR #376, opened 2020-09-11
Implement support for `calc()` expressions

### LINKED ISSUE #292: Ideas for parser quality of life improvement
* n-ary `either` Would make things like https://github.com/Siteimprove/alfa/blob/master/packages/alfa-css/src/value/color.ts#L55-L61 much nicer to write and read. Need that all parsers args return the same type.

* helpers functions in alfa-css, things such as https://github.com/Siteimprove/alfa/blob/style-custom-property/packages/alfa-css/src/value/color/var.ts#L121-L122 Could also use a `parseFunctionCall` taking care of matching the function name, parsing the args and separators, parsing the closing parenthesis, … would simplify a lot of code like https://github.com/Siteimprove/alfa/blob/master/packages/alfa-css/src/value/color/hsl.ts#L163 which is essentially duplicated in a few places.

kasperisager:
This pull request introduces initial support for `calc()` expressions and extends the `font-size` property definition with support for these. Only basic arithmetic has been implemented and no additional math functions, such as `min()`, `max()`, `clamp()` etc., are supported yet.

Closes #138.
Closes #292.


packages/alfa-style/src/property/font.ts
```diff
@@ -111,24 +118,47 @@ export namespace Font {
...
+
+              return Length.of(parent.value * value.value, parent.unit);
+            }
+
+            return value;
+          });
+
+          size = expression.toLength().or(expression.toPercentage()).get();
+        }
+
```
Jym77, line 161: Shouldn't the `calc` evaluation happen at a higher level (`compute`, I guess)?
As far as I understand, `calc` could happen almost anywhere and we probably don't want to handle the evaluation at property level… It's `Length` (and others) that should allow `Calculation` as values.

packages/alfa-css/src/value/calculation.ts
```diff
@@ -0,0 +1,848 @@
...
+
+    export abstract class Binary extends Operation<[Expression, Expression]> {
+      protected constructor(operands: [Expression, Expression], kind: Kind) {
+        super(operands, kind);
+      }
+    }
+  }
+
+  export class Sum extends Operation.Binary {
+    public static of(
```
Jym77, line 523: Could be overloaded as `of(fst: Expression, snd: Expression)` for flexibility of use (same with other binary operations).

Jym77 (approved):

packages/alfa-style/src/property/font.ts
kasperisager: Resolution of `calc()` expressions is specific to each property that accepts them and so is the way `calc()` expressions are parsed. For example, the `font-size` property only allows `calc()` expressions of type `<length-percentage>` and so `calc()` expressions with a different type should be rejected at parse time. At computed value time, the `font-size` property knows to resolve percentages in `calc()` expressions against the parent `font-size`, but this is different from property to property. For even other properties, such as `width` and `height`, `calc()` expressions can't even be fully resolved at computed value time and will carry over.

So, it's... complicated 🙈

packages/alfa-css/src/value/calculation.ts
kasperisager: That's equivalent to the current signature; `foo(...args: [string, number])` expands to `foo(args_0: string, args_1: number)`.

PR #383, opened 2020-09-15
Correctly handle children of `<iframe>` elements in accessibility tree and `isRendered()`

Jym77:
Nodes that represent nothing (`<template>` and descendants of `<iframe>`) are marked as not rendered and are not exposed in the accessibility tree.


packages/alfa-dom/src/node.ts
```diff
@@ -260,6 +262,30 @@ export abstract class Node
...
+      // @see https://html.spec.whatwg.org/multipage/scripting.html#the-template-element
+      // "In a rendering, the template element represents nothing."
+      and(Element.isElement, hasName("template"))(this) ||
+
+      // @see https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element
+      // "Descendants of iframe elements represent nothing. (In legacy user agents that do not support iframe elements,
+      // the contents would be parsed as markup that could act as fallback content.)"
+      this.ancestors().some(and(Element.isElement, hasName("iframe")))
+    )
+  }
```
kasperisager, line 288: I'm not too keen on introducing this API as it doesn't have a well-defined purpose other than catering to our immediate needs in the construction of the accessibility tree. As you note in the docblock, it's not that clear what the API _should do_.

I'd dig around Servo and Chromium for hints on how they handle this. If that doesn't provide any useful insight, just inline the needed bits so we don't introduce public API with no clear purpose.

kasperisager (dismissed):

packages/alfa-dom/src/node.ts
Jym77: Yes, I'm also far from happy to put that here…
The point was that it is used both in building the accessibility tree, and in `isRendering`. So I wasn't sure where to put it in order to share it…

kasperisager: Only the bit about `<iframe>` descendants is needed for those though; `<template>` elements are hidden by the UA style sheet:

https://github.com/Siteimprove/alfa/blob/7ad79c9899dc836df9d2b305ea27ba6ef8225141/packages/alfa-cascade/src/user-agent.ts#L14-L19

All that's left is then just a single line of code which is easily inlined.

Jym77: 🤔 but then shouldn't we rather just update that stylesheet to hide children of `<iframe>`?

kasperisager: That's one option, and probably the easiest solution for lack of a layout system 😄 I just dug through Chromium and it ignores all nodes without an associated layout object when building its accessibility tree:

https://github.com/chromium/chromium/blob/f9a52600a8d706ca524e8b10b2fc0fbbde992b57/third_party/blink/renderer/modules/accessibility/ax_layout_object.cc#L627-L631

Jym77: Tried to hide children of `<iframe>` in UA style sheet, but my CSS-fu is not good enough 😖 
https://github.com/Siteimprove/alfa/pull/383/files#diff-5be8aa0a1cd557655c16d24015883e65R229-R234

I'm not happy with the current solution that duplicates the logic, especially the but in `isRendered` is in a unlikely place and prone to be forgotten later 😢 

Do you have a better idea for handling this?

packages/alfa-cascade/src/user-agent.ts
```diff
@@ -219,4 +219,17 @@ export const UserAgent = h.sheet([
...
+   * "Descendants of iframe elements represent nothing. (In legacy user agents that do not support iframe elements,
+   * the contents would be parsed as markup that could act as fallback content.)"
+   */
+
+  // This is not enough as it won't hide the text children of iframe :-(
+  // <iframe> <span>hidden</span> Not hidden </iframe>
+
+  // h.rule.style("iframe > *", {
+  //   display: "none",
+  // }),
```
kasperisager, line 234: That's of course a very argument for why CSS won't do the trick 😄

packages/alfa-dom/src/node.ts
kasperisager: Until we have a layout system in place then any solution that doesn't introduce additional public API surface is a good solution to me.

packages/alfa-rules/src/common/predicate/is-rendered.ts
```diff
@@ -3,8 +3,18 @@ import { Element, Comment, Node } from "@siteimprove/alfa-dom";
...
+const { and } = Predicate;
+
 export function isRendered(device: Device): Predicate<Node> {
   return (node) => {
+    // descendants of iframe are not rendered, they are fallback content for legacy browsers…
+    // @see https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element
+
+    if (node.ancestors().some(and(Element.isElement, hasName("iframe")))) {
+      return false;
+    }
```
kasperisager, line 16: This will cause `isRendered()` to walk to the root of every node visited, making the complexity quadratic in the depth of the node for which the function was called. As `isRendered()` already traverses the ancestor chain, it's enough to check if the _parent_ is an `<iframe>` element.

packages/alfa-aria/src/node.ts
```diff
@@ -365,7 +365,10 @@ export namespace Node {
         // Get the children implicitly owned by the element. These are the
         // children in the flat tree that are neither claimed already nor
         // explicitly owned by the element.
-        const implicit = node
+
+        // children of iframe are not included, they are fallback content for legacy browsers…
+        // @see https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element
+        const implicit = node.name === "iframe" ? Sequence.empty<dom.Node>() : node
```
kasperisager, line 371: This only causes the implicit children to be excluded, not the explicit children.

packages/alfa-rules/src/common/predicate/is-rendered.ts
```diff
@@ -3,8 +3,18 @@ import { Element, Comment, Node } from "@siteimprove/alfa-dom";
 import { Predicate } from "@siteimprove/alfa-predicate";
 import { Style } from "@siteimprove/alfa-style";
 
+import { hasName } from "@siteimprove/alfa-dom/src/node/element/predicate/has-name";
```
kasperisager, line 6: Destructure the predicate from the `Element` class.

kasperisager (dismissed):

packages/alfa-dom/src/node.ts
Jym77: 👍 Then let's do it that way.

packages/alfa-aria/src/node.ts
Jym77: That was intentional as I assumed that `aria-own` should be respected, but I was indeed wrong…
```html
<iframe><span id="foo">Foo</span></iframe>
<div aria-label="foofoo" aria-owns="foo">Owns foo</div>
```
The `<div>` doesn't actually own `foo`…

kasperisager (approved):

PR #394, opened 2020-09-17
Add remaining name computation steps for `<input>` and `<textarea>` elements

### LINKED ISSUE #303: Include `placeholder` attribute in accessible name computation
https://github.com/w3c/html-aam/issues/167

### LINKED ISSUE #365: Add missing name computation steps for `<input>` elements
Currently, only connected `<label>` elements are considered when computing names for `<input>` elements: 

https://github.com/Siteimprove/alfa/blob/1858ef0d5c869da944353219af357078f177196d/packages/alfa-aria/src/feature.ts#L498

We're therefore missing the remaining steps outlined in https://w3c.github.io/html-aam/#accessible-name-and-description-computation.

- [ ] Follow up on DEV-10518.
- [ ] Follow up on DEV-10526.

kasperisager:
Closes #303.
Closes #365.


packages/alfa-aria/src/feature.ts
```diff
@@ -495,7 +458,47 @@ const Features: Features = {
...
+        if (
+          test<Element>(
+            hasInputType("text", "password", "search", "tel", "url"),
+            element
+          )
+        ) {
+          return Name.fromSteps(
+            () => nameFromLabel(element, device, state),
+            () => nameFromAttribute(element, "title"),
+            () => nameFromAttribute(element, "placeholder")
```
Jym77, line 471: I'm considering whether `nameFromAttribute` should accept several attribute names and test them in order: `nameFromAttribute(element, "title", "placeholder")`.

packages/alfa-aria/test/name.spec.tsx
```diff
@@ -540,6 +540,126 @@ test(`.from() determines the name of a <select> element with a <label> parent
...
+          Name.Source.label(input.attribute("value").get()),
+        ])
+      ),
+      [],
+    ],
+  ]);
+});
+
+test(`.from() determines the name of an <input type="reset"> element`, (t) => {
+  const input = <input type="submit" />;
```
Jym77, line 616: Is it "reset" or "submit"?

packages/alfa-aria/src/feature.ts
```diff
@@ -495,7 +458,47 @@ const Features: Features = {
...
+          return nameFromAttribute(element, "value");
+        }
+
+        if (test<Element>(hasInputType("submit"), element)) {
+          return Name.fromSteps(
+            () => nameFromAttribute(element, "value"),
+            () => Branched.of(Option.of(Name.of("Submit")))
+          );
+        }
+
```
Jym77, line 485: I'm not a huge fan of replacing the `switch` with a chain of `if` that I find harder to mentally parse and understand (+lot of repeated boilerplate).
I do agree that moving the input type computation out of this is good, though…

Maybe have a `inputType(element)` helper that can be used both here (`switch(inputType(element))` and in `hasInputType` and does the work?

Jym77 (changes requested):
One test doesn't do what's one the label.
The rest is optional polishing editorial.

packages/alfa-aria/test/name.spec.tsx
kasperisager: Bad copy/pasta; it's "reset".

packages/alfa-aria/src/feature.ts
kasperisager: Seems like a fine idea 👍

packages/alfa-aria/src/feature.ts
kasperisager: The only "boilerplate" I'm not too keen on having to repeat is the type annotation in `test()`, but that's caused by an underlying issue due to the conflation of predicates and type guards. I'm progressing on a way to solve this though. I do personally prefer the combination of `test()` and a predicate as predicates are much more expressive than `switch` statements. I'd therefore rather work on patterns for making this even easier to express using predicates than fit it into something that we can express using `switch`.

kasperisager: It might look like this:

```ts
const role = match(element)
  .when(hasInputType("button"), () => Role.of("button"))
  .when(hasInputType("checkbox"), () => Role.of("checkbox"))
  .getOrElse(() => Role.of("textbox"));
```

That looks surprisingly close to something `Option<T>`-like 🤔

kasperisager: Seeing how `Either<L, R>` is already right biased I may be able to tack this stuff directly onto it 🤔

kasperisager: @Jym77 Let me know what you think of the above. While it isn't something I'll be introducing in this pull request as it depends on some underlying changes to the `Predicate<T, U extends T>` type which I need to split in two, `Predicate<T>` and `Refinement<T, U extends T>`, it's something I can start working on separately.

kasperisager: Oh, another _really_ nice thing about the selective functor pattern used above: There's no implicit fall through foot gun lying around. That's another reason why I highly dislike the `switch` statement.

Jym77: Yeah, `switch` has always been the poor man's pattern matching…
I say this makes sense.

Jym77 (approved):

PR #398, opened 2020-09-18
Allow passing embedded and shadow content in `h()`

Jym77:


packages/alfa-dom/src/h.ts
```diff
@@ -37,6 +39,8 @@ export function h(
 }
 
 export namespace h {
+  import nor = Predicate.nor;
+
```
kasperisager, line 43: Use destructuring assignment and move to line 31.

packages/alfa-dom/src/h.ts
```diff
@@ -1,29 +1,31 @@
...
   NamespaceRule,
   PageRule,
+  Rule,
   StyleRule,
   SupportsRule,
 } from "./style/rule";
-import { Sheet } from "./style/sheet";
+import {Sheet} from "./style/sheet";
+import {Shadow} from "./node/shadow";
+import {Predicate} from "@siteimprove/alfa-predicate";
```
kasperisager, line 28: Looks like something overruled Prettier?

kasperisager, line 28: Move to line 2.

packages/alfa-dom/src/h.ts
```diff
@@ -72,16 +76,31 @@ export namespace h {
       .filter(Namespace.isNamespace)
       .getOr(Namespace.HTML);
 
-    return Element.of(
+    const contentDocument = children.find(Document.isDocument);
+    const shadowTree = children.find(Shadow.isShadow);
```
kasperisager, line 80:
```suggestion
    const document = children.find(Document.isDocument);
    const shadow = children.find(Shadow.isShadow);
```

packages/alfa-dom/src/h.ts
```diff
@@ -72,16 +76,31 @@ export namespace h {
...
+
+    const element = Element.of(
       Option.of(namespace),
       None,
       name,
       attributes,
-      children.map((child) =>
+      // filtering out children that are Document or Shadow
```
kasperisager, line 87:
```suggestion
```

packages/alfa-dom/src/h.ts
```diff
@@ -72,16 +76,31 @@ export namespace h {
...
       ),
       style.length === 0 ? None : Option.of(block)
     );
+
+    if (contentDocument !== undefined) {
+      element._attachContent(contentDocument);
+    }
+    if (shadowTree !== undefined) {
+      element._attachShadow(shadowTree);
+    }
```
kasperisager, line 101:
```suggestion
    if (content !== undefined) {
      element._attachContent(content);
    }

    if (shadow !== undefined) {
      element._attachShadow(shadow);
    }
```

packages/alfa-dom/src/h.ts
```diff
@@ -72,16 +76,31 @@ export namespace h {
...
       Option.of(namespace),
       None,
       name,
       attributes,
-      children.map((child) =>
+      // filtering out children that are Document or Shadow
+      children
+        .filter(nor(Document.isDocument, Shadow.isShadow))
+        .map((child) =>
         typeof child === "string" ? h.text(child) : child
```
kasperisager, line 91: Needs formatting.

packages/alfa-dom/src/h.ts
```diff
@@ -104,6 +123,19 @@ export namespace h {
     );
   }
 
+  export function shadow(
+    children: Array<Node | string>,
+    mode: Shadow.Mode = Shadow.Mode.Open,
+    style: Array<Sheet> = []
```
kasperisager, line 129:
```suggestion
    mode: Shadow.Mode,
    children: Array<Node | string>,
    style: Array<Sheet> = []
```

Matches `Shadow.of()` argument order.

kasperisager (changes requested):

kasperisager:
Changing to minor as this modifies public API, both behaviour and signature.

packages/alfa-dom/src/h.ts
Jym77: 🤯 
I am experiencing problems with my Prettier setup not working since the latest IDE upgrade 😞

packages/alfa-dom/src/h.ts
Jym77: 🤔 That was a conscious choice because I guess that our main use case won't care about the mode and I wanted to make it optional.
I've overloaded the signature instead, so we can have the quick `h.shadow([foo, bar])` which I guess we'll use often.

packages/alfa-dom/src/h.ts
```diff
@@ -104,6 +120,34 @@ export namespace h {
...
+
+    return Shadow.of(
+      mode,
+      children!.map((child) =>
+        typeof child === "string" ? text(child) : child
+      ),
+      style
+    );
+  }
+
```
kasperisager, line 150: This turned very complex very fast 🙈 I do see your point in the mode being less important than the children, and possibly the style, so let's take the chance to make it the last parameters in both `Shadow.of()` and `h.shadow()`. The coming release will contain breaking changes either way.

kasperisager (changes requested):

kasperisager (approved):

PR #399, opened 2020-09-18
Make `<object>` elements with a content document focusable

### LINKED ISSUE #388: `<object>` elements are not considered focusable
https://act-rules.github.io/rules/8fc3b6#failed-example-1
`<object title="" data="/test-assets/moon-audio/moon-speech.mp3"></object>`

The `<object>` has no role, no attribute, no name.
When rendered in Chrome, it has focusable content (the control buttons for the audio). But Alfa ignore these and therefore completely exclude the `<object>` from the accessibility tree. This result in that test case being Inapplicable instead of Failed.

More generally, `isSuggestedFocusableElement` should likely be refined for `<object>` .

Jym77:
Closes #388.


packages/alfa-rules/test/sia-r63/rule.spec.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R63, { document }), [
+    failed(R63, target, { 1: Outcomes.HasNoName})
+  ])
+});
+
+test("evaluate() fails an object with no name", async (t) => {
+  // Content document is needed to expose the object.
+  // const target = <object>{h.document(["Some nested document"])}</object>;
```
Jym77, line 31: And we already need #398 😁

packages/alfa-dom/src/node/element.ts
```diff
@@ -442,3 +445,4 @@ function isEditingHost(element: Element): boolean {
 function isBrowsingContextContainer(element: Element): boolean {
   return element.name === "iframe";
 }
+
```
kasperisager, line 448:
```suggestion
```

packages/alfa-rules/test/sia-r63/rule.spec.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const target = <object></object>;
+  target._attachContent(h.document(["Some nested document"]));
```
kasperisager, line 33: This is a package internal API which shouldn't be used so let's wait for #398.

kasperisager (changes requested):

packages/alfa-rules/test/sia-r63/rule.spec.tsx
kasperisager: All set 👍

packages/alfa-rules/test/sia-r63/rule.spec.tsx
```diff
@@ -0,0 +1,51 @@
+import { jsx } from "@siteimprove/alfa-dom/jsx";
+import { test } from "@siteimprove/alfa-test";
+
+import { Document, h } from "@siteimprove/alfa-dom";
```
kasperisager, line 4:
```suggestion
import { h } from "@siteimprove/alfa-dom/h";
import { jsx } from "@siteimprove/alfa-dom/jsx";
import { test } from "@siteimprove/alfa-test";

import { Document } from "@siteimprove/alfa-dom";
```

kasperisager (changes requested):
Looks good! 💯 Just a minor re-arrangement of some imports.

kasperisager (approved):

PR #400, opened 2020-09-18
Ignore header cells of presentational tables in SIA-R46

### LINKED ISSUE #390: SIA-R46 should not apply to cells in presentational tables
https://act-rules.github.io/rules/d0f69e#inapplicable-example-7
```html
<table role="presentation">
	<tr>
		<th>Time</th>
	</tr>
	<tr>
		<td>12:00</td>
	</tr>
</table>
```

The ACT rule requires the table to be included in the accessibility tree.
SIA-R46 uses `isPerceivable` which checks whether an inclusive descendant is included in the accessibility tree.

The `<th>` element is indeed exposed, making the case Applicable for Alfa…

Jym77:
Ignore tables and headers that are presentational, even if they have non-presentational children.
Closes #390.

### DIFF
--- a/packages/alfa-rules/src/sia-r46/rule.ts
+++ b/packages/alfa-rules/src/sia-r46/rule.ts
@@ -10,10 +10,10 @@ import { Page } from "@siteimprove/alfa-web";
 import { expectation } from "../common/expectation";
 
 import { hasRole } from "../common/predicate/has-role";
-import { isPerceivable } from "../common/predicate/is-perceivable";
+import { isIgnored } from "../common/predicate/is-ignored";
 
 const { isElement, hasName, hasNamespace } = Element;
-const { and, equals } = Predicate;
+const { and, equals, not } = Predicate;
 
 export default Rule.Atomic.of<Page, Element>({
   uri: "https://siteimprove.github.io/sanshikan/rules/sia-r46.html",
@@ -30,7 +30,7 @@ export default Rule.Atomic.of<Page, Element>({
               and(
                 hasNamespace(Namespace.HTML),
                 hasName("table"),
-                isPerceivable(device)
+                not(isIgnored(device))
               )
             )
           );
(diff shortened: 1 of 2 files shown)


packages/alfa-rules/src/sia-r46/rule.ts
```diff
@@ -44,7 +44,7 @@ export default Rule.Atomic.of<Page, Element>({
                 // The table model only works if the element is a th.
                 hasName("th"),
                 hasRole("rowheader", "columnheader"),
-                isPerceivable(device)
+                not(isIgnored(device))
```
kasperisager, line 47: This should still use `isPerceivable()`.

kasperisager (changes requested):

packages/alfa-rules/src/sia-r46/rule.ts
Jym77: That was a conscious choice because the rule (https://act-rules.github.io/rules/d0f69e) is requiring both the `th` and its `table ancestor to be "included in the accessibility tree" and I guess they should be handled the same way.

Difference will be for something like
```html
<table>
  <tr>
    <th role="none"> <img src="foo.jpg" alt="Foo" /></th>
  <tr>
</table>
```

🤔 Looking at it now, I wonder if the ACT rule is correct in ignoring that…
I'll file an issue on ACT rules so at least we know what we should do.

Jym77: Checking with NVDA, the header is not announce in that case. So it is correct to skip this.

This is already Inapplicable due the role on the `th`. I am not sure it is possible to exclude an "element with a role of `row/columnheader`" from the accessibility tree without making its content accessible (I can see how to exclude it by means of `display: none` or `aria-hidden` but both these methods are inherited to descendants).

So I am not sure it really makes a difference, and the NVDA test make me believe that `not(isIgnored)` is correct.

kasperisager: The header is required to be visible for it to be applicable, which is also checked by `isPerceivable()`.

kasperisager (approved):

PR #403, opened 2020-09-21
Implement matching for `(max|min)-(width|height)` features

### LINKED ISSUE #252: Implement media feature matching for dimensional viewport features
Currently, only `orientation` is supported: https://drafts.csswg.org/mediaqueries/#mf-dimensions

kasperisager:
Closes #252.

### DIFF
--- a/packages/alfa-media/src/resolver.ts
+++ b/packages/alfa-media/src/resolver.ts
@@ -1,0 +1,62 @@
+import { Converter, Length } from "@siteimprove/alfa-css";
+import { Device } from "@siteimprove/alfa-device";
+
+/**
+ * Resolvers are functions that resolve values to their canonical, computed
+ * representation.
+ *
+ * @internal
+ */
+export namespace Resolver {
+  /**
+   * Resolve a length in an arbitrary unit to a length in pixels.
+   *
+   * @see https://drafts.csswg.org/css-values/#lengths
+   *
+   * @remarks
+   * Relative lengths in media queries are based on initial values of the
+   * associated properties.
+   */
+  export function length(length: Length, device: Device): Length<"px"> {
+    const { unit, value } = length;
+    const { viewport } = device;
+
+    switch (unit) {
+      // https://www.w3.org/TR/css-values/#em
+      case "em":
+      // https://www.w3.org/TR/css-values/#rem
+      case "rem":
+        return Length.of(16 * value, "px");
+
+      // https://www.w3.org/TR/css-values/#ex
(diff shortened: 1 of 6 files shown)


packages/alfa-media/src/resolver.ts
```diff
@@ -0,0 +1,62 @@
+import { Converter, Length } from "@siteimprove/alfa-css";
+import { Device } from "@siteimprove/alfa-device";
+
+/**
+ * Resolvers are functions that resolve values to their canonical, computed
+ * representation.
+ *
+ * @internal
+ */
+export namespace Resolver {
```
Jym77, line 10: I'm a bit worried that this is almost reimplementing the length resolver from `alfa-style` (except using the initial value of 16). Somehow, we shouldn't have two places to resolve lengths the same way.

Looks like the resolver from `alfa-style` could have a third argument for `targetUnit` with default value of `length.unit`, and then this is essentially `@alfa-style/Resolver.length(length, Style.of("font-size: 16px", device) , "px")` (by crudely injecting style).

Jym77 (changes requested):

packages/alfa-media/src/resolver.ts
kasperisager: What we're seeing here is really a feature more than a design flaw 😄 One reason for separating everything into small packages was to enforce very strict boundaries between packages. As `alfa-style` depends on `alfa-cascade` which depends on `alfa-media`, value resolvers _cannot_ live in `alfa-style` if we want to use them in `alfa-media` as that would create a circular dependency. But, the value resolvers in `alfa-style` depend on style information and so _shouldn't_ be used in `alfa-media` as value resolution for media queries is _not_ allowed to be based on style information. So, the circular dependency is really a sign that what we're trying to do shouldn't be done, at least not in the way we're attempting 💡

The way to solve this would be to abstract the value resolution into a separate package that knows about neither style information nor media queries. Instead, it should rely on some generic interface for getting answers to questions such as "what is the current font size" or "what is the root font size". That needs more design work though and is beyond the scope of the changes I'm introducing here so I'll open a separate issue.

Jym77: Yes, I agree on the conclusion that value resolution should live somewhere else. Locally, we need a `Resolver.length(length, currentSize, rootSize, device /* for viewport*/, targetUnit)` and both usages are fairly simple calls to that.
I also agree that refactoring it is out of scope here.

Jym77 (approved):

PR #406, opened 2020-09-22
Don't require answers for `<iframe>` elements that definitely embed the same resource

Jym77:
* fix `Document#parent` to handle `{ nested: true }` option.
* fix SIA R15 to look at attribute value for sources set, given that identical attributes are actually different.
* fix SIA R15 to consider `srcdoc` attribute
* fix SIA R15 to be inapplicable on single iframe
* fix SIA R15 to look at all iframe, independently of nesting depth
* add tests for SIA R15


packages/alfa-rules/src/sia-r15/rule.ts
```diff
@@ -61,11 +63,23 @@ export default Rule.Atomic.of<Page, Iterable<Element>, Question>({
             }, Map.empty<Option<string>, List<Element>>())
             .values()
         );
+        return filter(groups, (group) => group.size > 1);
       },
 
       expectations(target) {
+        // @see https://html.spec.whatwg.org/multipage/iframe-embed-object.html#process-the-iframe-attributes
         const sources = Set.from(
-          map(target, (iframe) => iframe.attribute("src"))
```
Jym77, line 68: Looking only at the attribute was bad as two attributes with the same value end up being different, thus we were never hitting the "same resource" case.

packages/alfa-rules/src/sia-r15/rule.ts
```diff
@@ -40,9 +40,11 @@ export default Rule.Atomic.of<Page, Iterable<Element>, Question>({
             )
           );
 
-        const roots = iframes.groupBy((iframe) => iframe.root());
+        const roots = iframes.groupBy((iframe) =>
+          iframe.root({ nested: true })
+        );
```
kasperisager, line 45: There's no need to group the iframes by root if the rule is to consider all iframes across the page as a single bag.

packages/alfa-rules/src/sia-r15/rule.ts
```diff
@@ -61,11 +63,23 @@ export default Rule.Atomic.of<Page, Iterable<Element>, Question>({
...
+            iframe
+              .attribute("srcdoc")
+              .map((srcdoc) => srcdoc.value)
+              .getOr(
+                iframe
+                  .attribute("src")
+                  .map((source) => source.value)
+                  .getOr("about:blank")
+              )
+          )
```
kasperisager, line 82: For `<iframe>` elements without a `src` attribute, we should either do a structural comparison to determine content document equality, or simply always pose a question when there's more than one such iframe.

kasperisager (changes requested):

packages/alfa-rules/src/sia-r15/rule.ts
Jym77: 🙈

packages/alfa-dom/src/node.ts
```diff
@@ -226,8 +226,22 @@ export abstract class Node
...
-    return value === this;
+  protected _structurallyEquals(value: unknown): value is this {
+    return (
+      value instanceof Node &&
+      // We can't use Sequence#equals because it is oblivious to structural equality and will just call item.equals
+      Array.from(
+        this.children({ nested: true, composed: true })
+      ).every((thisChild, idx) =>
+        thisChild._structurallyEquals(
+          Array.from(value.children({ nested: true, composed: true }))[idx]
```
kasperisager, line 237: Child lists aren't memoized so this will go through the work of constructing the child list for `value` for every child of `this`.

packages/alfa-dom/src/node.ts
```diff
@@ -226,8 +226,22 @@ export abstract class Node
...
+        )
+      )
+    );
+  }
+
+  public equals(value: unknown, structural: boolean = false): value is this {
+    return structural ? this._structurallyEquals(value) : value === this;
```
kasperisager, line 244: It holds that if `value === this` then `value` also structurally equals `this`. That's an efficient of short circuiting the check.

packages/alfa-dom/src/node/document.ts
```diff
@@ -44,6 +49,14 @@ export class Document extends Node {
     return "/";
   }
 
+  protected _structurallyEquals(value: unknown): value is this {
+    return (
+      value instanceof Document &&
+      super._structurallyEquals(value) &&
+      Iterable.equals(this.style, value.style)
```
kasperisager, line 56: We also need to implement structural equality for the CSS structures as documents with style sheets otherwise won't be considered structurally equal unless `value === this`.

packages/alfa-dom/src/node/element.ts
```diff
@@ -255,6 +256,31 @@ export class Element extends Node implements Slot, Slotable {
...
+      this._attributes.every((thisAttribute) =>
+        value._attributes.some((valueAttribute) =>
+          thisAttribute.equals(valueAttribute, true)
+        )
+      ) &&
+      value._attributes.every((valueAttribute) =>
+        this._attributes.some((thisAttribute) =>
+          thisAttribute.equals(valueAttribute, true)
+        )
+      ) &&
```
kasperisager, line 277: These checks are `O(n * m)` with `n` being the number of attributes in `value` and `m` being the number of attributes in `this`. We can reduce that to `O(n)` by using a `Map<string, string>`.

packages/alfa-dom/src/node/element.ts
```diff
@@ -255,6 +256,31 @@ export class Element extends Node implements Slot, Slotable {
...
+      this.style.equals(value.style) &&
+      this.id.equals(value.id) &&
+      Iterable.equals(this.classes, value.classes)
```
kasperisager, line 280: These are already encoded within the attributes of `Element` and don't need to be checked.

packages/alfa-dom/src/node/element.ts
```diff
@@ -255,6 +256,31 @@ export class Element extends Node implements Slot, Slotable {
...
+        value._attributes.some((valueAttribute) =>
+          thisAttribute.equals(valueAttribute, true)
+        )
+      ) &&
+      value._attributes.every((valueAttribute) =>
+        this._attributes.some((thisAttribute) =>
+          thisAttribute.equals(valueAttribute, true)
+        )
+      ) &&
+      this.style.equals(value.style) &&
```
kasperisager, line 278: We also need to implement structural equality for the CSS structures as elements with inline styles otherwise won't be considered structurally equal unless `value === this`.

kasperisager (dismissed):
On second thought, let's not go with a structural equality check of `<iframe>` elements without a `src` attribute as it's evidently introducing too much complexity for too little gain. I've left some comments on the code either way, though we need to revert the changes.

Instead, for `<iframe>` elements without a `src` attribute we'll always pose a question.

packages/alfa-rules/src/sia-r15/rule.ts
```diff
@@ -107,3 +106,71 @@ export namespace Outcomes {
...
+  }
+
+  public hash(hash: Hash) {
+    Hash.writeString(hash, this._source);
+    this._value.hash(hash);
+  }
+
+  public toString() {
+    return `{source: ${this._source}, value: ${this._value}}`;
+  }
```
kasperisager, line 144:
```suggestion
```

packages/alfa-rules/src/sia-r15/rule.ts
```diff
@@ -107,3 +106,71 @@ export namespace Outcomes {
...
+
+  public get value(): Option<string> {
+    return this._value;
+  }
+
+  public equals(value: unknown): value is this {
+    return (
+      value instanceof EmbeddedResource &&
+      this._source === value._source &&
+      this._value.equals(value._value)
```
kasperisager, line 133:
```suggestion
      value._source === this._source &&
      value._value.equals(this._value)
```

packages/alfa-rules/src/sia-r15/rule.ts
```diff
@@ -107,3 +106,71 @@ export namespace Outcomes {
...
+
+  public hash(hash: Hash) {
+    Hash.writeString(hash, this._source);
+    this._value.hash(hash);
+  }
+
+  public toString() {
+    return `{source: ${this._source}, value: ${this._value}}`;
+  }
+}
```
kasperisager, line 145: This class looks like a rather complicated wrapper around a `string` and so whatever it solves could just as well be solved by a `Map` 🤔

packages/alfa-rules/src/sia-r15/rule.ts
```diff
@@ -107,3 +106,71 @@ export namespace Outcomes {
...
+
+  // Otherwise, grab the src attribute.
+  function getUrl(value: string): EmbeddedResource {
+    try {
+      return EmbeddedResource.of("src", new URL(value).href);
+    } catch (error: unknown) {
+      if (error instanceof TypeError) {
+        return EmbeddedResource.of("invalid");
+      } else {
+        throw error;
```
kasperisager, line 167: In Alfa, `throw` will cause the equivalent of a panic as we rely on `Result<T, E>` for error handling. I imagine that a panic isn't what we want in this case 🙈

packages/alfa-rules/src/sia-r15/rule.ts
```diff
@@ -107,3 +106,71 @@ export namespace Outcomes {
...
+    return EmbeddedResource.of(
+      "srcdoc",
+      iframe.attribute("srcdoc").get().value
+    );
+  }
+
+  // Otherwise, grab the src attribute.
+  function getUrl(value: string): EmbeddedResource {
+    try {
+      return EmbeddedResource.of("src", new URL(value).href);
```
kasperisager, line 162: We can resolve the reference against the URL of the `Response` instance.

kasperisager (changes requested):

packages/alfa-rules/src/sia-r15/rule.ts
Jym77: 🤔 this is a couple of `string` where the second one may be missing. Structural equality is needed, so `[string, string]` is not really good. I'm not sure a `Map` is the best way to model that (as in, in every case there will always be exactly one key with a value).

packages/alfa-rules/src/sia-r15/rule.ts
Jym77: Would be cool, but I don't know where to access that 🙃

packages/alfa-rules/src/sia-r15/rule.ts
Jym77: Side question: is there a reason why we want to have `value` on the left side of comparison (other than unifying with other `equals` in the codebase)?

packages/alfa-rules/src/sia-r15/rule.ts
```diff
@@ -105,3 +106,29 @@ export namespace Outcomes {
...
+    } catch {
+      return Map.of(["invalid", ""]);
+    }
+  }
+
+  return iframe
+    .attribute("src")
+    .map((attribute) => getUrl(attribute.value))
+    .getOrElse(() => Map.of(["nothing", ""]));
+}
```
kasperisager, line 134: This wasn't _quite_ what I was trying to convey, but it's an interesting take nonetheless 😄 My point was that the combination of `Source` and whatever value is associated with that `Source` might as well be _one_ string, such as `"src" + href`. You could then use that as the _key_ in a `Map`, with the values being the `<iframe>` elements that share that particular `Source` + value combination.

Either way, I realised that we've been staring ourselves blind on what to stuff in a `Set` or `Map` when the better solution is much simpler. Since we're just trying to figure out if all `<iframe>` elements definitely point to the same thing, it's just a matter of taking the first `<iframe>` element of the group and comparing it to the rest. If they're all the "same" as the first `<iframe>` then we don't need to ask a question.

kasperisager (changes requested):

packages/alfa-rules/src/sia-r15/rule.ts
kasperisager: See https://github.com/Siteimprove/alfa/issues/420#issuecomment-701479693.

packages/alfa-rules/src/sia-r15/rule.ts
kasperisager: I'll introduce an `alfa-url` package that will deal with this stuff. Other packages are already juggling around WHATWG `URL` instances and these really don't work all that well with Alfa's architecture.

packages/alfa-rules/src/sia-r15/rule.ts
kasperisager: See https://github.com/Siteimprove/alfa/pull/406#discussion_r497617734.

packages/alfa-rules/src/sia-r15/rule.ts
kasperisager: The very first `equals()` method introduced put `value` on the left and `this` on the right as that aligns with the return type: `value is this`.

packages/alfa-rules/src/sia-r15/rule.ts
kasperisager: Tracking this in #421.

packages/alfa-rules/src/sia-r15/rule.ts
Jym77: 👍 Makes sense.
I was mimicking the order of `this.equals(value)` => `this === value` 😀

packages/alfa-rules/src/sia-r15/rule.ts
kasperisager: All set!

packages/alfa-rules/src/sia-r15/rule.ts
```diff
@@ -38,37 +40,39 @@ export default Rule.Atomic.of<Page, Iterable<Element>, Question>({
...
+          )
         );
 
         return {
           1: expectation(
-            sources.size === 1,
+            source.isSome() &&
+              // always ask question if we can't find source, presumably the content doc is fed another way into the iframes
+              source.get() !== "invalid:" &&
+              source.get() !== "nothing:",
```
kasperisager, line 75: Seems like these cases should simply be `None`.

kasperisager (changes requested):

packages/alfa-rules/src/sia-r15/rule.ts
Jym77: 🤔 that would break the agnosticity of `commonValue`.
OTOH, `embeddedResource` should have returned a `Result` from the start and that makes this test much nicer indeed 👍

packages/alfa-rules/src/sia-r15/rule.ts
```diff
@@ -57,18 +53,24 @@ export default Rule.Atomic.of<Page, Iterable<Element>, Question>({
...
       expectations(target) {
-        const sources = Set.from(
-          map(target, (iframe) => iframe.attribute("src"))
+        const embedSameResource = [...target].every(
+          (element, i, elements) =>
+            // This is either the first element...
+            i === 0 ||
+            // ...or an element that embeds the same resource as the element
+            // before it.
+            referenceSameResource(response.url)(element, elements[i - 1])
```
kasperisager, line 68: This is closer to what I had in mind: A pairwise comparison that bails out as soon as possible.

kasperisager (approved):

kasperisager (approved):

packages/alfa-rules/src/sia-r15/rule.ts
Jym77: That's actually computing the embedded resource of each element twice (for comparing with previous and next) while first mapping an embedded resource computation and then comparing them would do it only once (and should also bail out asap if done lazily).
Not really important because this is probably not costly…

kasperisager: Nah, it's just comparing some strings and parsing some URLs 🤷

Issue #410, opened 2020-09-23
Investigate structured conditional computations


kasperisager:
https://github.com/Siteimprove/alfa/pull/394#discussion_r490285836

> Inspired by https://www.staff.ncl.ac.uk/andrey.mokhov/selective-functors.pdf, I drafted this API:
> 
> ```ts
> interface Selective<S, T = never> {
>   value: Either<S, T>;
> 
>   when<P extends S, U>(
>     when: Refinement<S, P>,
>     mapper: Mapper<P, U>
>   ): Selective<S, T | U>;
> 
>   get(): T;
> 
>   getOr<U>(value: U): T | U;
> 
>   getOrElse<U>(thunk: Thunk<U>): T | U;
> };
> ```
> 
> The idea is that `Selective` models selective function application of type `S -> T`. Given a type refinement `S -> P`, `Selective#when()` applies a function `P -> U`, producing a new `Selective` of type `S -> T | U`. It looks like this in action:
> 
> ```ts
> declare const string: Selective<string>;
> 
> declare const isFoo: Refinement<string, "foo">;
> 
> declare const isBar: Refinement<string, "bar">;
> 
> const number = string
>   // If string is "foo", produce 42
>   .when(isFoo, (foo) => 42)
>   // If string is "bar", produce 56
>   .when(isBar, (bar) => 56)
>   // If string is neither "bar" nor "foo", produce 23
>   .getOr(23);
> ```

PR #415, opened 2020-09-24
Split `Predicate<T, U extends T>` into `Predicate<T>` and `Refinement<T, U extends T>`

kasperisager:
This pull request splits the current `Predicate<T, U extends T>` into two separate types:

1. `Predicate<T>`.
2. `Refinement<T, U extends T>`, living in the new `@siteimprove/alfa-refinement` package.

`Predicate<T>` now always acts as a plain predicate, performing no type narrowing, and `Refinement<T, U extends T>` always acts as a type predicate, narrowing the passed `T` to a `U`. This solves a long-standing issue wherein plain predicates of type `Predicate<T, T>` were, in certain situations, being inferred as type guards, causing the `else` branch of whatever code was using the predicate to narrow the corresponding value to `never`:

```ts
const isFoo: Predicate<string, string> = (value) => value === "foo";

const value: string = "bar";

if (isFoo(value)) {
  // value is `string`
} else {
  // value is `never`
}
```

To circumvent this, the `Predicate.test()` method had to be used, often requiring an explicit type annotation to avoid the incorrect inference, as seen here:

https://github.com/Siteimprove/alfa/blob/f07ab0eb791d7dbdd8a8c6647518cda090f1ca56/packages/alfa-aria/src/feature.ts#L378-L417

By separating these two behaviours, narrowing and non-narrowing, into separate types, we ensure that stuff like this doesn't happen. This separation also paves the way for #410, which will allow us to further clean up code like seen above.


packages/alfa-refinement/test/refinement.spec.ts
```diff
@@ -0,0 +1,75 @@
+// import { test } from "@siteimprove/alfa-test";
```
Jym77, line 1: Tests should be active.

packages/alfa-predicate/test/predicate.spec.ts
```diff
@@ -1,75 +1,75 @@
...
-      | null
-      | undefined
-      | symbol = value;
-
-    t.equal(foo, 123);
-  } else {
-    t.fail();
-  }
-});
+// import { test } from "@siteimprove/alfa-test";
```
Jym77, line 1: Looks like these are all tests for refinement. Should be completely killed here (or replaced by Predicate test).

Jym77 (changes requested):
Looks fine.

packages/alfa-refinement/test/refinement.spec.ts
kasperisager: Good catch, completely forgot about those! 🤦

packages/alfa-predicate/test/predicate.spec.ts
kasperisager: Ditto on https://github.com/Siteimprove/alfa/pull/415#discussion_r496573586.

Jym77 (approved):

PR #418, opened 2020-09-25
SIA-R68: Exclude elements with an `aria-busy=true` ancestor

Jym77:
as per the ACT rule.

### DIFF
--- a/packages/alfa-rules/src/sia-r68/rule.ts
+++ b/packages/alfa-rules/src/sia-r68/rule.ts
@@ -1,7 +1,7 @@
 import { Rule, Diagnostic } from "@siteimprove/alfa-act";
-import { Node, Role } from "@siteimprove/alfa-aria";
+import { Node as AccNode, Role } from "@siteimprove/alfa-aria";
 import { Device } from "@siteimprove/alfa-device";
-import { Element, Namespace } from "@siteimprove/alfa-dom";
+import { Element, Namespace, Node } from "@siteimprove/alfa-dom";
 import { Iterable } from "@siteimprove/alfa-iterable";
 import { Predicate } from "@siteimprove/alfa-predicate";
 import { Ok, Err } from "@siteimprove/alfa-result";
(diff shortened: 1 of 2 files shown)


packages/alfa-rules/src/sia-r68/rule.ts
```diff
@@ -97,3 +103,37 @@ function isRequiredChild(
...
   }
 }
+
+/**
+ * Collect all descendants of node that:
+ * * match includeNode; and
+ * * do not have an ancestor matching excludeSubtree; and
+ * * have an ancestor matching collectSubtree
+ */
+function visit<T extends Node = Node>(
```
Jym77, line 113: This is meant to be generic enough to re-factor `visit` in R83 (and maybe other). But I want feedback before refactoring 😉

packages/alfa-rules/src/sia-r68/rule.ts
```diff
@@ -97,3 +103,37 @@ function isRequiredChild(
...
+
+    const children = node.children(options);
+
+    for (const child of children) {
+      yield* doVisit(child, collect);
+    }
+  }
+
+  return doVisit(node, false);
+}
```
kasperisager, line 139: This is definitely a case of premature abstraction so let's not do that 🙈 This might provide some good insights: https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction

kasperisager (changes requested):

packages/alfa-rules/src/sia-r68/rule.ts
```diff
@@ -97,3 +89,33 @@ function isRequiredChild(
...
 }
+
+/**
+ * Collect all descendants of node that:
+ * * are non-ignored HTML or SVG elements with a role requiring specific children; and
+ * * do not have an aria-busy ancestor
+ */
+function* visit(node: Node, device: Device): Iterable<Element> {
+  if (and(isElement, hasAttribute("aria-busy", equals("true")))(node)) {
+    return [];
```
kasperisager, line 100:
```suggestion
    return;
```

For generator functions, the concepts of "yielding" and "returning" are different. Whatever a generator function returns will be available as the _final_ value _after_ it is done yielding values. In this case, we only care about the yielded values of type `Element` and so might as well have the generator return `void` rather than `never[]`.

kasperisager (dismissed):

packages/alfa-rules/src/sia-r68/rule.ts
```diff
@@ -97,3 +92,37 @@ function isRequiredChild(
...
+/**
+ * Collect all descendants of the given node where the descendant:
+ *
+ * - is a non-ignored HTML or SVG element with a role requiring specific
+ *   children; and
+ * - does not have an `aria-busy` ancestor.
+ */
+function* visit(node: Node, device: Device): Iterable<Element> {
+  if (
+    Refinement.and(isElement, hasAttribute("aria-busy", equals("true")))(node)
```
kasperisager, line 105: `Refinement.and()` falls back to the signature of `Predicate.and()` when no refinements are involved. You can therefore pull `and()` from `Refinement` rather than `Predicate` to avoid using both.

kasperisager (changes requested):

packages/alfa-rules/src/sia-r68/rule.ts
Jym77: Defaulted to `Refinement` for all logical connectors as it seems easier. But not sure it really is the correct thing to do (vs pulling `and` from `Refinement` and `equals, or` from `Predicate`…)

kasperisager: They _all_ fall back to the equivalent `Predicate` signature so that's fine 👍

kasperisager (approved):
💯

PR #421, opened 2020-09-30
Introduce `@siteimprove/alfa-url` package

kasperisager:
This pull request introduces a new `@siteimprove/alfa-url` package with an implementation of an immutable URL class modelled on the WHATWG URL specification. An obvious question might be "why not just use the WHATWG URL DOM API?" to which the obvious answer is "well, it kinda does". That is, `URL.parse()` relies on the WHATWG URL DOM API for parsing URL strings and extracting URL records. From there, the `URL` class also adds some convenience like implementations of `Equatable`, `Hashable`, and `Serializable`, making it easy to use `URL` instances in the various data structures Alfa provides.

The caveat to this approach is of course that we now rely on the WHATWG URL specification being implemented by whatever runtime Alfa is used in, which shouldn't really be a problem. It's of course readily available in browsers and is also supported in Node.js 10+ as well as Deno.

### DIFF
--- a/packages/alfa-url/src/url.ts
+++ b/packages/alfa-url/src/url.ts
@@ -1,0 +1,312 @@
+import { Equatable } from "@siteimprove/alfa-equatable";
+import { Hash, Hashable } from "@siteimprove/alfa-hash";
+import { Iterable } from "@siteimprove/alfa-iterable";
+import { Serializable } from "@siteimprove/alfa-json";
+import { Option, None } from "@siteimprove/alfa-option";
+import { Result, Err } from "@siteimprove/alfa-result";
+import { Sequence } from "@siteimprove/alfa-sequence";
+
+import * as json from "@siteimprove/alfa-json";
+
+const { isEmpty } = Iterable;
+
+/**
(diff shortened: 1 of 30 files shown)


packages/alfa-url/src/url.ts
```diff
@@ -0,0 +1,364 @@
...
+      path: this._path.toArray(),
+      query: this._query.getOr(null),
+      fragment: this._fragment.getOr(null),
+    };
+  }
+
+  /**
+   * @see https://url.spec.whatwg.org/#concept-url-serializer
+   */
+  public toString(): string {
```
Jym77, line 208: This is a piece of code with a lot of logic and conditionals which is therefore super prone to regressions and would most benefit from unit tests.
I think this is OK as I don't think that bugs here are going to be tricky to track down, or likely to cause big trouble. So we can probably leave it as is for now and add tests if we encounter bugs later…

Jym77 (approved):

packages/alfa-url/src/url.ts
kasperisager: Agreed 😄

PR #426, opened 2020-10-02
Implement traversal API for `Graph`

kasperisager:
This pull request implements a traversal API for the `Graph` class, which is needed for correctly dealing with cyclic `aria-owns` references:

https://github.com/Siteimprove/alfa/blob/2db1ec794a867b858a8f971e48229332d5a8dd46/packages/alfa-aria/test/node.spec.tsx#L182-L188

The changes are breaking as `Graph#neighbors()` now returns `Iterable<T>` rather than `Option<Set<T>>`, which seemed like a strong promise.

### DIFF
--- a/packages/alfa-graph/src/graph.ts
+++ b/packages/alfa-graph/src/graph.ts
@@ -2,7 +2,7 @@ import { Equatable } from "@siteimprove/alfa-equatable";
 import { Iterable } from "@siteimprove/alfa-iterable";
 import { Serializable } from "@siteimprove/alfa-json";
 import { Map } from "@siteimprove/alfa-map";
-import { Option } from "@siteimprove/alfa-option";
+import { Sequence } from "@siteimprove/alfa-sequence";
 import { Set } from "@siteimprove/alfa-set";
 
 import * as json from "@siteimprove/alfa-json";
@@ -33,8 +33,8 @@ export class Graph<T>
     return this._nodes.keys();
   }
 
-  public neighbors(node: T): Option<Set<T>> {
-    return this._nodes.get(node);
+  public neighbors(node: T): Iterable<T> {
+    return this._nodes.get(node).getOr([]);
   }
 
   public has(node: T): boolean {
@@ -42,13 +42,11 @@ export class Graph<T>
   }
 
   public add(node: T): Graph<T> {
-    const nodes = this._nodes;
-
-    if (nodes.has(node)) {
+    if (this.has(node)) {
       return this;
     }
 
-    return new Graph(nodes.set(node, Set.empty()));
+    return new Graph(this._nodes.set(node, Set.empty()));
   }
 
   public delete(node: T): Graph<T> {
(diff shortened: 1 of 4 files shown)


packages/alfa-graph/src/graph.ts
```diff
@@ -158,4 +171,60 @@ export namespace Graph {
...
+
+  export interface Traversal {
+    <T>(graph: Graph<T>, root: T): Iterable<T>;
+  }
+
+  /**
+   * @see https://en.wikipedia.org/wiki/Depth-first_search
+   */
+  export const DepthFirst: Traversal = function* <T>(graph: Graph<T>, root: T) {
+    const queue = [root];
```
Jym77, line 183: Should this rather be called `stack` given that it's how it's used?

packages/alfa-graph/test/graph.spec.ts
```diff
@@ -25,8 +42,86 @@ test("#disconnect() disconnects two nodes in a graph", (t) => {
...
+      3, //    |- 3
+      6, //    |- 6
+      7, //    |- 7
+    ]
+  );
+});
+
+test("#hasPath() checks if there's a path from one node to another", (t) => {
+  // foo always has a path to itself
+  t.equal(graph.hasPath("foo", "foo"), true);
```
Jym77, line 117: This should be tested on `bar` given that there is a cycle foo -> baz -> foo.

packages/alfa-graph/src/graph.ts
```diff
@@ -158,4 +171,60 @@ export namespace Graph {
...
+      const next = queue.shift()!;
+
+      yield next;
+
+      for (const neighbor of graph.neighbors(next)) {
+        if (seen.has(neighbor)) {
+          continue;
+        }
+
+        seen = seen.add(neighbor);
```
Jym77, line 225: I'm really bothered by the fact that the handling of `seen` is done at different time by each traversal (enqueuing or dequeuing). As far as I can tell, breadth first search would work as well if handling `seen` was done at dequeuing time instead (depth first need `seen` to be handled at dequeuing time). Having this here is an optimisation to keep the queue smaller (it's not going to save on processing time) but ends up making the two traversal very different when the difference could be minimal (stack vs queue).

I don't know how big the graphs we handle are and how important this optimisation is… I don't know if we should align the code or maybe just add a comment on each ("need to be done at dequeuing time" / "done at enqueuing time to save space")… 🤷

Jym77 (changes requested):
Nitpicking a bit…

packages/alfa-graph/src/graph.ts
kasperisager: This is quite literally the text book difference between iterative DFS and BFS though:

> This non-recursive implementation is similar to the non-recursive implementation of depth-first search, but differs from it in two ways:
> 
> - it uses a queue (First In First Out) instead of a stack and
> - it checks whether a vertex has been discovered before enqueueing the vertex rather than delaying this check until the vertex is dequeued from the queue.

https://en.wikipedia.org/wiki/Breadth-first_search#More_details

Jym77 (approved):

PR #428, opened 2020-10-02
Introduce `@siteimprove/alfa-protractor` package

kasperisager:
This is currently blocked on #287.


packages/alfa-protractor/src/protractor.ts
```diff
@@ -0,0 +1,277 @@
...
+} from "@siteimprove/alfa-dom";
+import { Request, Response } from "@siteimprove/alfa-http";
+import { Page } from "@siteimprove/alfa-web";
+
+import { ElementFinder } from "protractor";
+
+export namespace Protractor {
+  export type Type = ElementFinder;
+
+  export async function toPage(value: Type): Promise<Page> {
```
Jym77, line 35: I'm seeing this (and companion functions) in most test integration packages, without seeing obvious differences. Should it be refactored?

Jym77 (approved):

packages/alfa-protractor/src/protractor.ts
kasperisager: It _should_, yes, I just haven't yet found a good way of sharing that particular bit of functionality. The issue is that the functions need to be defined _within_ the lambda passed to APIs like `#executeScript()` as the lambdas will be stringified and sent over the wire. Stuff like this won't work:

```ts
// packages/alfa-dom/native.ts

export function toNode(node: globalThis.Node): Node;
```

```ts
// packages/alfa-protractor/src/protractor.ts

import { toNode } from "@siteimprove/alfa-dom/native";

const node = browser.executeScript(() => toNode(window.document));
```

PR #445, opened 2020-10-07
Fix several `Element#tabIndex()` cases

### LINKED ISSUE #444: `Element#tabIndex()` should disregard the `draggable` attribute
For the following `` element, `Element#tabIndex()` returns `Some { 0 }` when it should return `None`:

```html
<img draggable="false">
```

- [ ] Follow up on DEV-10652.

kasperisager:
Closes #444.

### DIFF
--- a/packages/alfa-rules/test/sia-r2/rule.spec.tsx
+++ b/packages/alfa-rules/test/sia-r2/rule.spec.tsx
@@ -38,13 +38,6 @@ test("evaluate() is inapplicable to an <img> element with a presentational role"
   t.deepEqual(await evaluate(R2, { document }), [inapplicable(R2)]);
 });
 
-test(`evaluate() is inapplicable to a non-draggable <img> element with a
-      presentational role`, async (t) => {
-  const document = Document.of([<img role="none" draggable="false"></img>]);
-
-  t.deepEqual(await evaluate(R2, { document }), [inapplicable(R2)]);
-});
-
 test("evaluate() is inapplicable to an <img> element that is hidden", async (t) => {
   const document = Document.of([<img hidden alt="Hello world"></img>]);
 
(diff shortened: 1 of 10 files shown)


packages/alfa-rules/test/sia-r2/rule.spec.tsx
```diff
@@ -32,20 +32,27 @@ test("evaluate() fails an <img> element without an accessible name", async (t) =
...
+  const document = Document.of([<img role="none"></img>]);
 
   t.deepEqual(await evaluate(R2, { document }), [inapplicable(R2)]);
 });
 
-test("evaluate() is inapplicable to an image that is marked as decorative", async (t) => {
-  const document = Document.of([<img role="none"></img>]);
+test(`evaluate() is inapplicable to a non-draggable <img> element with a
+      presentational role`, async (t) => {
+  const document = Document.of([<img role="none" draggable="false"></img>]);
```
Jym77, line 43: How is `draggable="false"` affecting the rule? Given that these two tests give the same result whether it's here or not…
If we want to illustrate `draggable` in the rule there should be a test where it does make a difference to have it.

Jym77 (changes requested):

packages/alfa-rules/test/sia-r2/rule.spec.tsx
kasperisager: The point is precisely that it _shouldn't_ affect the result, as per #444. `` was previously considered focusable, meaning you couldn't assign it a presentational role.

kasperisager: All add a reference to the issue in a comment.

Jym77: 😅 
🤔 shouldn't this test rather be in `alfa-dom` as a test for `isSuggestedFocusable`? Since the error was there rather than in the rule itself.

kasperisager: Yes and no; there are some tests for `Element#tabIndex()` is `alfa-dom`, but the test case in `alfa-rules` is really there to document a prior bug and prevent regression.

Jym77 (approved):

PR #448, opened 2020-10-08
Implement style resolution for stateful pseudo classes

### LINKED ISSUE #447: Implement style resolution for stateful pseudo classes

kasperisager:
This pull request implements style resolution for the following stateful pseudo classes:

- `:hover`
- `:active`
- `:focus`
- `:focus-within`
- `:focus-visible`
- `:link`
- `:visited`

A new class, `Context`, has been introduced to model context-dependant information during selector matching, such as which element is currently in focus, which elements have been visited, and so on. This class is heavily inspired by Servo ([`components/selectors/context.rs`][components/selectors/context.rs]) and works by modelling state as bit flags associated with elements. As an example, this means that it's up to the caller, not `Selector#matches()`, to decide if a given element links to a resource that has been visited; `Selector#matches()` simply checks the associated state, not the actual URL of the element being matched against.

This pull request also introduces changes to the way cascade is managed to allow efficiently re-computing cascade for elements when context changes. The `Cascade` class is now no longer immutable, but instead incrementally builds up a rule tree and caches entries for elements keyed on context. A "baseline" cascade is performed at construction time and computes cascade for all elements in a document with an empty context to benefit from ancestor filtering during selector matching. From then, matching declarations for individual elements can be queried with an optional context that defaults to an empty context. Whenever a non-yet-seen context is passed for an element, the rule tree is extended with matching declarations for the element and associated context.

Closes #447.

[components/selectors/context.rs]: https://github.com/servo/servo/blob/4886788a8ed975cbc1ae4f05cd8ce0ab41cd30e2/components/selectors/context.rs

### DIFF
--- a/packages/alfa-selector/src/context.ts
+++ b/packages/alfa-selector/src/context.ts
@@ -1,0 +1,96 @@
+import { Element } from "@siteimprove/alfa-dom";
+import { Map } from "@siteimprove/alfa-map";
+
+export class Context {
+  public static of(state: Iterable<[Element, Context.State]>): Context {
+    return new Context(Map.from(state));
+  }
+
+  private static _empty = new Context(Map.empty());
+
+  public static empty(): Context {
+    return this._empty;
+  }
+
+  private readonly _state: Map<Element, Context.State>;
+
+  private constructor(state: Map<Element, Context.State>) {
+    this._state = state;
+  }
+
+  public hasState(element: Element, state: Context.State): boolean {
+    return this._state.get(element).some((found) => (found & state) !== 0);
+  }
+
+  public getState(element: Element): Context.State {
+    return this._state.get(element).getOr(Context.State.None);
+  }
+
+  public setState(element: Element, state: Context.State): Context {
+    return new Context(this._state.set(element, state));
+  }
+
+  public hover(element: Element): Context {
+    return this.setState(element, this.getState(element) | Context.State.Hover);
+  }
+
+  public static hover(element: Element): Context {
+    return this.empty().hover(element);
+  }
+
+  public isHovered(element: Element): boolean {
+    return this.hasState(element, Context.State.Hover);
+  }
+
+  public active(element: Element): Context {
+    return this.setState(
+      element,
+      this.getState(element) | Context.State.Active
+    );
+  }
+
+  public static active(element: Element): Context {
+    return this.empty().active(element);
+  }
+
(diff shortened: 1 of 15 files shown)


packages/alfa-selector/src/context.ts
```diff
@@ -0,0 +1,96 @@
...
+    return this._state.get(element).some((found) => (found & state) !== 0);
+  }
+
+  public getState(element: Element): Context.State {
+    return this._state.get(element).getOr(Context.State.None);
+  }
+
+  public setState(element: Element, state: Context.State): Context {
+    return new Context(this._state.set(element, state));
+  }
```
Jym77, line 31: Given the usages we have below, I wonder if we'd need a helper do to the `this.setState(element, this.getState(element) | state)` part. And then just have `hover = foo(Context.State.Hover)`, …

packages/alfa-selector/src/selector.ts
```diff
@@ -1086,6 +1091,53 @@ export namespace Selector {
...
+    }
+
+    private constructor() {
+      super("focus-within");
+    }
+
+    public matches(element: Element, context: Context): boolean {
+      return (
+        context.isFocused(element) ||
+        element
```
Jym77, line 1115: Don't we have an `inclusiveDescendants` method?

packages/alfa-selector/src/selector.ts
```diff
@@ -1086,6 +1091,53 @@ export namespace Selector {
...
+          .descendants({ flattened: true })
+          .filter(isElement)
+          .some((element) => context.isFocused(element))
```
Jym77, line 1118: Shouldn't that be eta-contracted (`some(context.isFocused)`)?

Jym77 (changes requested):

packages/alfa-selector/src/selector.ts
kasperisager: That would break the `this` binding of `Context#isFocused`.

packages/alfa-selector/src/context.ts
kasperisager: Agreed 👍 I'll add an `Context#addState()` method.

packages/alfa-selector/src/selector.ts
kasperisager: We sure do!

Jym77 (approved):

PR #469, opened 2020-10-20
Accept table model errors

Jym77:
Instead of crashing on tables with model errors, we just store the errors for possible later processing.
Tables with empty rows or columns are usually not a problem and are now treated as OK.
Tables with doubly covered cells might create other problems. The internal scanning algorithm accounts for this possibility, so it should be OK-ish to just accept them…

- [ ] Follow up on DEV-10706.


kasperisager (changes requested):
I'd be surprised if the HTML specification included a processing algorithm that didn't anticipate and handle invalid user input so let's just ditch the errors entirely. If a table can always be formed either way, we're good 👍

Jym77:
> I'd be surprised if the HTML specification included a processing algorithm that didn't anticipate and handle invalid user input so let's just ditch the errors entirely. If a table can always be formed either way, we're good 👍

I'm a bit reluctant to completely ditch them… It feels like ditching valuable diagnostic info.
Especially, the "colliding cells" one is likely to create hard to track issues later (because the slot is skipped during headers association, so we'll end up with failing the header rules for hard to track reasons).

kasperisager:
Helping diagnose why rules are failing is not something the table package should concern itself with so I'm not too concerned about that aspect. Tracking these errors "just in case" does however add needless complexity to a package that is already non-trivially complex.

Similar to HTML validation, the _specific_ cases of table validation errors that might cause accessibility issues should be encoded in rules that clearly outline the error condition and remedy.

packages/alfa-table/src/cell.ts
```diff
@@ -285,19 +285,22 @@ export namespace Cell {
...
-        explicitHeaders,
-        implicitHeaders
+        List.empty()
       );
+      this._downwardGrowing = downwardGrowing;
+      this._scope = scope;
+      this._explicitHeaders = List.from(explicitHeaders);
+      this._implicitHeaders = List.from(implicitHeaders);
+
+      return this; // for chaining
```
kasperisager, line 303: Why break immutability here?

kasperisager (changes requested):

packages/alfa-table/src/cell.ts
Jym77: Cells (the concept) de facto mutate during the build process (notably adding the scope on header cells).
When `Table#slots` was returning an `Option<Cell>` that was OK since "mutating" the cell could just be done by mutating `Table#slots` instead.
But now, we need `Table#slots` to return a `Set<Cells>` to accept these table model errors. When the cell change, we would need to remove the old one from the set, and add the new one, which is more complex than just updating the cell.

OTOH, I agree than only the `Cell.Builder` should mutate, the product should be immutable 😓

kasperisager: If preserving immutability can be done simply be removing the old cell and adding the new one then let's do that. Shortcutting this at the cost of losing immutability seems like the wrong tradeoff.

kasperisager (changes requested):

packages/alfa-table/src/cell.ts
Jym77: I'm afraid that "simply" is not really going to be simple…
There's many places where cells are created and updated, often in bulk by sub-steps, and it's not that easy to track what the "old" cell is. The easiest way I see is to add an `id` to each cell that is kept through all incarnations and use this for comparing.

kasperisager: This really sounds like something we should fix regardless then as it seems more complex than it has to be 🤔 First things first then: Why are the cells stored in a set? With a set, we need to worry about keeping a reference to the old cell. If we were using a list, we'd simply identify cells by their index, which is effectively the ID you're describing.

Jym77: I don't see an easy way to fetch the index of the old version, unless I already have the old version and then whatever the collection, I can just `reject` the old one.
I've initially chose `Set` precisely to let it handle the "no duplicate" part for me.

packages/alfa-table/src/table.ts
```diff
@@ -331,7 +332,9 @@ export namespace Table {
             this._slots[x] = [];
           }
           for (let y = cell.anchor.y; y < cell.anchor.y + cell.height; y++) {
-            this._slots[x][y] = Some.of(cell);
+            this._slots[x][y] = (this._slots[x][y] ?? Set.empty())
+              .reject((item) => item.equals(cell))
```
kasperisager, line 336: This method will cause the entire set to be reconstructed.

packages/alfa-table/src/cell.ts
kasperisager: The point was that you'd always use the same index for the "same" cell and then simply update the cell at a given index. Either way, there seems to be some fundamental architectural hurdles in this package that we need to sort out before we pile on additional complexity.

packages/alfa-table/src/table.ts
Jym77: Which is why I ended up mutating the cell builder instead…
(also the set is supposed to have 0 or 1 element, unless there is a table model error)

packages/alfa-table/src/cell.ts
Jym77: Currently, cell mutation happens deeply at various places, so it's not easy to track where the cell comes from and what was its position in the list for when we need to update this list.
Additionally, a given cell might be present in several of these sets (if it covers several slots, which AFAIU must be the case for slots that are covered multiple time, i.e. the case where a set is needed rather than an option), potentially at different positions each time. So passing the information of where it was in each seems pretty much over complicated also.

packages/alfa-table/src/table.ts
kasperisager: Yeah, that does seem to confirm my suspicions in https://github.com/Siteimprove/alfa/pull/469#discussion_r509235445. I need to have a much deeper look at this package before deciding on an approach here 🤔 Given that this deals with an edge case that we've at least diagnosed as relating to invalid table markup, I'll probably not get around to it until after the stuff I'm currently prioritising is out of the way.

kasperisager:
Pulling this out of progress as it needs further refinement. I'm aiming to get a larger refactoring of the table package slated for next year. For now, seeing how resistant that package is to the changes we want to introduce here, I say we wait with introducing them.

kasperisager:
Superseded by #581.

packages/alfa-table/src/cell.ts
kasperisager: For posterity, this is what I had in mind:

https://github.com/Siteimprove/alfa/blob/4c5828ab8e9a74cc50f86760adc2982d91424106/packages/alfa-table/src/table.ts#L105-L106

This way, cells are stored in an array and referenced in the table grid by their index. When updating cells, they're simply overwritten in the cell array:

https://github.com/Siteimprove/alfa/blob/4c5828ab8e9a74cc50f86760adc2982d91424106/packages/alfa-table/src/table.ts#L527-L533

Header assignment is done at the very last step by mapping over the cell array and assigning headers:

https://github.com/Siteimprove/alfa/blob/4c5828ab8e9a74cc50f86760adc2982d91424106/packages/alfa-table/src/table.ts#L190-L192

PR #482, opened 2020-10-22
Reverse `wrapsText()` conditions in SIA-R83

kasperisager:
- [ ] Follow up on DEV-10706.


Jym77 (changes requested):
We should add a test to avoid regression since this was apparently not caught by existing tests…

Jym77 (approved):

PR #496, opened 2020-10-28
Implement `outline` CSS property

kasperisager:


packages/alfa-style/src/property/outline.ts
```diff
@@ -0,0 +1,209 @@
...
+   * @see https://drafts.csswg.org/css-ui/#outline
+   */
+  export const Shorthand = Property.Shorthand.of(
+    ["outline-width", "outline-style", "outline-color"],
+    (input) => {
+      let width: Width.Specified | undefined;
+      let style: Style.Specified | undefined;
+      let color: Color.Specified | undefined;
+
+      while (true) {
```
Jym77, line 160: We do seem to have a lot of these `while (true)` with `continue`s and `break` in style parsing. This is obviously due to the `||` operator in CSS which is order-independent.
I feel like we might need an helper to handle that to avoid duplicating the same pattern in all shorthand properties. Maybe something like `foo([Width.parse, Style.parse, Color.parse]) which would then return `[remaining, [width, style, color]])`? Typing might be annoying to due variadicity 🤔 

Anyway, not the topic for this PR…

Jym77 (approved):

PR #497, opened 2020-10-28
Implement SIA-R65

kasperisager:
The rule accounts for focus indicators implemented using the following properties:

- `outline`
- `text-decoration`


kasperisager:
@Jym77 There's still some outstanding work on this but feel free to start reviewing it when you have the time.

packages/alfa-rules/src/sia-r65/rule.ts
```diff
@@ -0,0 +1,114 @@
...
+        })
+      )
+      .filter(isElement)
+      .some((element) => {
+        const unset = Style.from(element, device);
+        const focus = Style.from(element, device, Context.focus(element));
+
+        // If the unset state does not have an outline, the focus state may use an
+        // outline as a focus indicator.
+        if (!hasOutline(unset) && hasOutline(focus)) {
```
Jym77, line 101: Should the rule also accept the reverse case (outline when not focused, no outline when focused), i.e. use a xor here?

packages/alfa-rules/src/sia-r65/rule.ts
```diff
@@ -0,0 +1,114 @@
...
+      })
+      .concat(
+        element.ancestors({
+          flattened: true,
+        })
+      )
+      .filter(isElement)
+      .some((element) => {
+        const unset = Style.from(element, device);
+        const focus = Style.from(element, device, Context.focus(element));
```
Jym77, line 97: There is something tricky with the descendant and ancestors here. The `Context` should put `focus` on the test target (the outer `element`) but is doing it here on the descendant/ancestor (the inner `element`) (if I understand correctly).

packages/alfa-rules/src/sia-r65/rule.ts
```diff
@@ -0,0 +1,114 @@
...
+  style
+    .computed("text-decoration-line")
+    .none((line) => line.type === "keyword") &&
+  style.computed("text-decoration-color").none(Color.isTransparent);
+
+function hasFocusIndicator(device: Device): Predicate<Element> {
+  return (element) => {
+    return element
+      .inclusiveDescendants({
+        flattened: true,
```
Jym77, line 87: Should this also look in nested browsing contexts?

packages/alfa-rules/src/sia-r65/rule.ts
```diff
@@ -0,0 +1,114 @@
...
+      })
+      .concat(
+        element.ancestors({
+          flattened: true,
```
Jym77, line 91: Should this also look in nested browsing contexts?

packages/alfa-rules/src/sia-r65/rule.ts
```diff
@@ -0,0 +1,114 @@
...
+  style.computed("outline-width").none(Length.isZero) &&
+  style
+    .computed("outline-color")
+    .none((color) => color.type === "color" && Color.isTransparent(color));
+
+const hasTextDecoration: Predicate<Style> = (style) =>
+  style
+    .computed("text-decoration-line")
+    .none((line) => line.type === "keyword") &&
+  style.computed("text-decoration-color").none(Color.isTransparent);
```
Jym77, line 81: These might end up being shared with "inline link is distinguishable" rule.

packages/alfa-rules/src/sia-r65/rule.ts
kasperisager: That's true and not at all what it should be doing, good catch! 👍

packages/alfa-rules/src/sia-r65/rule.ts
kasperisager: I don't hope that would ever happen, but those words have been repeated too often 😄 It's a tiny thing to account for so let's do that 👍

packages/alfa-rules/src/sia-r65/rule.ts
kasperisager: No, styles fortunately don't leak across browsing contexts.

packages/alfa-rules/src/sia-r65/rule.ts
kasperisager: Ditto on https://github.com/Siteimprove/alfa/pull/497#discussion_r516711016.

packages/alfa-rules/src/sia-r65/rule.ts
Jym77: 🤔 Shadow trees also offer a strong barrier for styles. But then, I guess `Style.from` handles that correctly…

kasperisager: While there's no way to reach _inside_ the shadow tree of a host, shadow trees can reach _outside_ their host using the `:host`, `:host()`, and `:host-context()` selectors: https://drafts.csswg.org/css-scoping-1/#host-selector

kasperisager:
I'll hold off on adding support for other types of focus indicators until we've had a chance to see how this rule fairs on actual sites.

Jym77 (approved):

PR #527, opened 2020-11-05
Ignore inclusive descendants of `aria-hidden=true` when computing accessible names

Jym77:


packages/alfa-aria/test/name.spec.tsx
```diff
@@ -961,3 +961,35 @@ test(`.from() correctly handles chained aria-labelledby references`, (t) => {
...
+  t.deepEqual(Name.from(heading, device).toArray(), [
+    [
+      Option.of(
+        Name.of("Hello", [
+          Name.Source.descendant(
+            heading,
+            Name.of("Hello", [
+              Name.Source.descendant(
+                exposed,
+                Name.of("Hello ", [Name.Source.data(text)])
```
Jym77, line 986: Interestingly, https://w3c.github.io/accname/#step2G does not require trimming of flattening the string, which means that accessible name of text nodes is not a flat string, contrary to what https://w3c.github.io/accname/#mapping_additional_nd_te claims:
> The purpose of the computation is to create a perceivable label or description for alternative presentations, in the form of a flat string of space separated textual tokens.

🤷

Jym77: https://github.com/w3c/accname/issues/63 🤣

packages/alfa-aria/src/name.ts
```diff
@@ -521,9 +521,15 @@ export namespace Name {
...
-      // Step 2A: Is the element hidden and not referenced?
+      // Step 2A: Is the element not referenced and either hidden or not exposed?
       // https://w3c.github.io/accname/#step2A
-      if (!state.isReferencing && !isRendered(element, device)) {
+      if (
+        !state.isReferencing &&
+        (!isRendered(element, device) ||
+          element
+            .attribute("aria-hidden")
+            .some((attribute) => attribute.value === "true"))
```
kasperisager, line 531: Let's stuff this in a predicate named `isHidden()` and reference https://w3c.github.io/accname/#dfn-hidden.

kasperisager (changes requested):

packages/alfa-aria/src/name.ts
```diff
@@ -521,15 +521,9 @@ export namespace Name {
...
+      // Step 2A: Is the element hidden and not referenced?
       // https://w3c.github.io/accname/#step2A
-      if (
-        !state.isReferencing &&
-        (!isRendered(element, device) ||
-          element
-            .attribute("aria-hidden")
-            .some((attribute) => attribute.value === "true"))
-      ) {
+      if (isHidden(element, device) && !state.isReferencing) {
```
kasperisager, line 526:
```suggestion
      if (!state.isReferencing && isHidden(element, device)) {
```

kasperisager (changes requested):

packages/alfa-aria/src/name.ts
Jym77: 🤦 🙈

kasperisager (approved):
Looks fantastic! 👌

PR #539, opened 2020-11-10
Support CSS text-indent property

### LINKED ISSUE #532: Implement `text-indent` CSS property
- [x] `text-indent` (https://drafts.csswg.org/css-text/#text-indent-property)

Jym77:
Closes #532

### DIFF
--- a/packages/alfa-style/src/property/text.ts
+++ b/packages/alfa-style/src/property/text.ts
@@ -7,6 +7,7 @@ import {
   System,
   Token,
 } from "@siteimprove/alfa-css";
+import { None, Option } from "@siteimprove/alfa-option";
 import { Parser } from "@siteimprove/alfa-parser";
 import { Err, Result } from "@siteimprove/alfa-result";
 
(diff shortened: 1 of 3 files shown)


packages/alfa-style/src/property/text.ts
```diff
@@ -86,6 +88,95 @@ export namespace Text {
...
+      Length | Percentage,
+      Option<Hanging>,
+      Option<EachLine>
+    ];
+
+    export type Computed = [
+      Length<"px"> | Percentage,
+      Option<Hanging>,
+      Option<EachLine>
+    ];
```
kasperisager, line 177:
```suggestion
    export type Specified = Length | Percentage;

    export type Computed = Length<"px"> | Percentage;
```

It turns out that no browsers support `each-line` and `hanging` (https://developer.mozilla.org/en-US/docs/Web/CSS/text-indent#Browser_compatibility) so we won't shouldn't support them either.

kasperisager (changes requested):

kasperisager (approved):

PR #549, opened 2020-11-18
Make elements with no content invisible

### LINKED ISSUE #520: `isVisible()` does not account for elements that have no visible content and therefore no dimensions
`isVisible()` returns `true` for the following cases:

```html
<div></div>
```

```html
<div>
  <span hidden>This text is hidden</span>
</div>
```

Jym77:
Closes #520.

Not sure this is what you had in mind 🤷


packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -70,3 +85,18 @@ function isClipped(
...
+  [
+    "audio",
+    "canvas",
+    "embed",
+    "iframe",
+    "img",
+    "input",
+    "object",
+    "video",
+  ].includes(element.name);
```
kasperisager, line 102:
```suggestion
const isReplaced = hasName(
  "audio",
  "canvas",
  "embed",
  "iframe",
  "img",
  "input",
  "object",
  "video",
);
```

packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -70,3 +85,18 @@ function isClipped(
       .some(isClipped);
   };
 }
+
+/**
+ * @see https://html.spec.whatwg.org/multipage/rendering.html#replaced-elements
```
kasperisager, line 90:
```suggestion
 * @see https://html.spec.whatwg.org/#replaced-elements
```

packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -7,6 +7,8 @@ import { Style } from "@siteimprove/alfa-style";
 
 import { isRendered } from "./is-rendered";
 import { isTransparent } from "./is-transparent";
+import { Iterable } from "@siteimprove/alfa-iterable";
+import every = Iterable.every;
```
kasperisager, line 11:
```suggestion
const { every } = Iterable;
```

packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -7,6 +7,8 @@ import { Style } from "@siteimprove/alfa-style";
 
 import { isRendered } from "./is-rendered";
 import { isTransparent } from "./is-transparent";
+import { Iterable } from "@siteimprove/alfa-iterable";
```
kasperisager, line 10: This should be moved up to other `@siteimprove/*` imports.

packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -33,7 +35,20 @@ export function isVisible(device: Device, context?: Context): Predicate<Node> {
...
+    },
+    // non-replaced elements with no visible child are not visible
+    // replaced elements are assumed to be replaced by something visible.
+    not(
+      and(
+        isElement,
+        and(not(isReplaced), (element) =>
+          every(
+            element.children({ nested: true, flattened: true }),
+            not(isVisible(device, context))
```
kasperisager, line 47: The complexity of `isVisible()` is currently something along the lines of `O(d)`, with `d` being the depth of the node, as it walks all the way up the ancestor chain. If we evaluate it for every child of the node as well, we'll end up with something along the lines of `O(d * n)`, with `n` being the size of the subtree of the node. That seems... excessive 🙈 We should probably throw a `Cache` in there somewhere to keep the complexity at `O(d)`.

kasperisager (changes requested):

packages/alfa-rules/src/common/predicate/is-visible.ts
Jym77: 🤦

packages/alfa-rules/src/common/predicate/is-visible.ts
Jym77: Gaaah 🙈

Jym77:
Gaah, `isTransparent` is also calling on parents… 🙈

Jym77:
Cache everything!

packages/alfa-rules/src/common/predicate/is-rendered.ts
```diff
@@ -6,34 +7,41 @@ import { Style } from "@siteimprove/alfa-style";
 
 const { isElement, hasName } = Element;
 
+const cache = Cache.empty<[Device, Context, Node], boolean>();
```
kasperisager, line 10:
```suggestion
const cache = Cache.empty<Device, Cache<Context, Cache<Node, boolean>>>();
```

The `Cache` class uses reference equality, not structural equality, in order to be able to only hold weak references to object types.

kasperisager (changes requested):

packages/alfa-rules/src/common/predicate/is-rendered.ts
Jym77: Yeah, I actually thought so yesterday night 🙈

kasperisager (changes requested):

kasperisager (approved):

PR #554, opened 2020-11-19
SIA-R10: ignore disabled elements

### LINKED ISSUE #516: SIA-R10 should not apply to disabled inputs
R10 is passing https://act-rules.github.io/rules/73f2c2#inapplicable-example-6
This is likely due to the fact that R10 looks literarily at the presence of an `aria-disabled` attribute on the element: https://github.com/Siteimprove/alfa/blob/master/packages/alfa-rules/src/sia-r10/rule.ts#L37
rather than on the value of the property which could be set by the HTML `disabled` attribute.

This could result in false positives in case of bad autocomplete on a disabled field.

Jym77:
Ignore elements not on the value of the `aria-disabled` attribute, but on the value of the `aria-disabled` state in the accessibility tree.
And reorder a bit the tests, to put the more generic (namespace), selective (element) and light at the start and only build the accessible node if needed.

Closes #516.


kasperisager (approved):
💯

PR #556, opened 2020-11-19
SIA R45: ignore tables with presentational role

### LINKED ISSUE #521: SIA-R45 should not apply to `<table>` elements that aren't exposed with a `table` role
R45 currently targets attributes in "`<table>` that are perceivable" while the ACT rule targets in "`<table>` that are included in the accessibility tree" (and visible). This results in disagreement on https://act-rules.github.io/rules/a25f45#inapplicable-example-2 where the `<table>` element is out of the accessibility tree due to its `role="presentation"` but is perceivable because it has descendants that are visible and included in the accessibility tree.

It's not clear that any of these is actually correct… A better Applicability might be about "`<table>` that are included in the AT with a role of `table` (+visible)" (or "`<table>` with a role of `table`" which seems to be the usual short form for it…)

Jym77:
Closes #521.


kasperisager (approved):
👏

PR #566, opened 2020-11-23
SIA R83: assume a used value for `normal` line-height

### LINKED ISSUE #515: SIA-R83 should assume a used value when the computed line height is `normal`
When the `line-height` is not specified, it defaults to the computed value `normal` which the rule cannot use to definitely find if there is a problem.
The rule should use a default for the used value of the property, so that it can detect these cases.

https://act-rules.github.io/rules/59br37#failed-example-1 and https://act-rules.github.io/rules/59br37#failed-example-2 are incorrectly handled because of this.

Jym77:
Closes #515.

Quick fix with a single used value, maybe not perfectly accurate.
I suggest we go with it for now and refine the code if this proves to be problematic.

### DIFF
--- a/packages/alfa-rules/src/sia-r83/rule.ts
+++ b/packages/alfa-rules/src/sia-r83/rule.ts
@@ -159,7 +159,7 @@ function wrapsText(device: Device): Predicate<Text> {
 }
 
 /**
- * "normal" line height seems to be everything but normal…
+ * "normal" line height is everything but normal…
  *
  * CSS recommends that it is "between 1.0 and 1.2"
  * @see https://drafts.csswg.org/css2/#valdef-line-height-normal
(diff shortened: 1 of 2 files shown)


packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -152,3 +157,22 @@ function wrapsText(device: Device): Predicate<Text> {
...
+ *
+ * MDN says that Desktop browsers use a value of roughly 1.2
+ * @see https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#normal
+ *
+ * This old post found out that it depend(ed?) on font-family and font-size :-/
+ * @see http://meyerweb.com/eric/thoughts/2008/05/06/line-height-abnormal/
+ *
+ * Keeping the simple version for now, if this causes troubles, we can later
+ * dig into browsers code and build a more complex version, possibly branched…
+ */
```
kasperisager, line 175: As explained in this article https://iamvdo.me/en/blog/css-font-metrics-line-height-and-vertical-align, the used value of `line-height: normal` depends on font metrics, as also verified by inspecting Servo: https://github.com/servo/servo/blob/6a3c3a4e18370d5604644d5491430b82d51c5be9/components/layout_2020/flow/inline.rs#L807. The exact metrics used are platform dependant, but overall match what's outlined in the article, that is the sum of the ascent, descent, and line gap:

- macOS: https://github.com/servo/servo/blob/8de1b8d3f4d0e814980a1f7836a9e865e0d0f53b/components/gfx/platform/macos/font.rs#L287
- Windows: https://github.com/servo/servo/blob/029049b48630f426ac5face4d9a9c0cd5195c66c/components/gfx/platform/windows/font.rs#L392

I also dug up this stuff in Chromium and it's about the same as in Servo: https://github.com/chromium/chromium/blob/53675c0acf7691ec1686a6aab6aab624f9b8a232/third_party/blink/renderer/core/layout/ng/inline/ng_line_height_metrics.h#L63. They've got some pretty involved hacks in place for determining platform specific font metrics though: https://github.com/chromium/chromium/blob/6efa1184771ace08f3e2162b0255c93526d1750d/third_party/blink/renderer/platform/fonts/font_metrics.cc

kasperisager (changes requested):

packages/alfa-rules/src/sia-r83/rule.ts
Jym77: Thanks for the links.
Do we have access to the font metrics to redo these computations?
My idea was to use the quick hack of 1.2 until it proves problematic 😅

kasperisager: > Do we have access to the font metrics to redo these computations?

We fortunately don't 🙈 Assuming a used value of 1.2 seems like the simplest approach and I'm perfectly happy with it 👍 We should adjust the comment to mention how things work in browsers with regards to font metrics and platform differences though.

kasperisager (approved):

PR #573, opened 2020-11-30
Implement CSS clip and clip-path properties

### LINKED ISSUE #533: Implement `clip` and `clip-path` CSS properties
- [ ] `clip` (https://drafts.fxtf.org/css-masking/#clip-property, deprecated)
- [ ] `clip-path` (https://drafts.fxtf.org/css-masking/#the-clip-path)

Jym77:
Closes #533.


packages/alfa-css/src/value/shape/rectangle.ts
```diff
@@ -0,0 +1,150 @@
...
+import { Hash } from "@siteimprove/alfa-hash";
+import { Parser } from "@siteimprove/alfa-parser";
+import { Err, Ok, Result } from "@siteimprove/alfa-result";
+import { Slice } from "@siteimprove/alfa-slice";
+
+import { Token } from "../../syntax/token";
+
+const { either, left, pair, peek, option, right, separatedList } = Parser;
+
+export class Rectangle extends Value<"shape"> {
```
Jym77, line 13: Not really a "shape" per se. But the plan is to group it with the shapes for `clip-path`, so I guess it makes sense.

packages/alfa-css/src/value/shape/rectangle.ts
```diff
@@ -0,0 +1,150 @@
...
+            pair(Token.parseComma, option(Token.parseWhitespace))
+          )
+        ),
+        Token.parseCloseParenthesis
+      )
+    )(input).flatMap(([remainder, result]) => {
+      const values = [...result];
+
+      const err: Result<[Slice<Token>, Rectangle], string> = Err.of(
+        "rect() must have exactly 4 arguments"
```
Jym77, line 139: Not sure if we want to accept more than 4 arguments (and just ditch the extra ones).

packages/alfa-style/src/property/clip.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * @see https://drafts.fxtf.org/css-masking/#clip-property
+ */
+export const Clip: Property<
+  Clip.Specified,
+  Clip.Computed
+> = Property.of(
+  Keyword.of("auto"),
+  either(Keyword.parse("auto"), Rectangle.parse),
+  (style) => style.specified("clip")
```
Jym77, line 25: Specs says computed is "as specified". Feels a bit weird to not absolutise anything. 🤷

packages/alfa-css/src/value/shape/rectangle.ts
```diff
@@ -0,0 +1,150 @@
...
+
+      if (values.length !== 4) {
+        return err;
+      }
+      return Ok.of<[Slice<Token>, Rectangle]>([
+        remainder,
+        Rectangle.of(values[0], values[1], values[2], values[3]),
+      ]);
+    });
+}
```
kasperisager, line 150: I've got a draft of the different shapes sitting in my local tree as I was experimenting a little while back, based on CSS Shapes: https://drafts.csswg.org/css-shapes/#supported-basic-shapes. I'll get them wrapped up and pushed.

kasperisager (changes requested):

packages/alfa-css/src/value/shape/rectangle.ts
Jym77: Perfect, implementing these would have been my next step 😀

kasperisager: The `rect()` shape corresponds to the new `inset()` shape from CSS Shapes:

https://github.com/Siteimprove/alfa/blob/18c3beb1d09bccb615b9d24c7cfbd1d09992ae87/packages/alfa-css/src/value/shape/inset.ts#L4-L36

Jym77: Looking more into it, I think that `rect()` and `inset()` are sufficiently different that two separate implementations will be easier. Notably:
* `inset()` arguments are `<length-percentage>`, `rect()` are `<length>` or `auto`. So, type `O` above would be different in both, even inheriting (to fix the radius at 0) would prove tricky.
* parse-wise, they are fairly different also, `inset()` accept less than 4 args (same meaning as for `border`) while `rect()` requires 4 args.

I think it's just easier to treat them as different shapes rather than trying to force `rect()` as being an `inset()`.
=> I think this PR is actually OK to merge as is and #591 keeps moving 😄

kasperisager: Apologies for stagnating on this! I think you're right that the `rect()` shape warrants its own separate definition, though then at the very least it should be added in #591 as a deprecated basic shape that only allows the `border-box` box value in `Shape`.

Jym77: I feel like I'm missing a connection 😅 
`rect()` can only be used in `clip`, and `clip` can only use `rect()` and `auto`. I don't see how `rect()` interacts with the various `<*-box>` types.

But then yes, putting an `export type Deprecated = Rectangle` in the `Shape` namespace is more or less part of the plan 😄

kasperisager: I added the basics so there are still some blanks to fill in: https://github.com/Siteimprove/alfa/blob/6578b25bcacb3d5aa982b5980e960b8b48923f45/packages/alfa-css/src/value/shape/rectangle.ts#L4-L24

kasperisager (approved):

kasperisager (approved):

PR #577, opened 2020-12-01
SIA R53: ignore headings that are not in the accessibility tree

### LINKED ISSUE #576: SIA-R53 should not apply to headings not in the accessibility tree as these have no level
The following case currently fails the rule:

```html
<h3 hidden>Foo</h3>
<h5>Bar</h5>
```

- [ ] Follow up on DEV-10889.

Jym77:
Closes #576.

### DIFF
--- a/packages/alfa-rules/src/sia-r53/rule.ts
+++ b/packages/alfa-rules/src/sia-r53/rule.ts
@@ -10,7 +10,7 @@ import { hasHeadingLevel } from "../common/predicate/has-heading-level";
 import { hasRole } from "../common/predicate/has-role";
 import { isIgnored } from "../common/predicate/is-ignored";
 
-const { equals, not } = Predicate;
+const { equals } = Predicate;
 const { isElement } = Element;
 
 export default Rule.Atomic.of<Page, Element>({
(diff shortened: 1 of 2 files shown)


packages/alfa-rules/src/sia-r53/rule.ts
```diff
@@ -18,7 +19,8 @@ export default Rule.Atomic.of<Page, Element>({
     const headings = document
       .descendants({ flattened: true })
       .filter(isElement)
-      .filter(hasRole("heading"));
+      .filter(hasRole("heading"))
+      .filter(not(isIgnored(device)));
```
kasperisager, line 23:
```suggestion
      .reject(isIgnored(device));
```

packages/alfa-rules/src/sia-r53/rule.ts
```diff
@@ -8,8 +8,9 @@ import { expectation } from "../common/expectation";
 
 import { hasHeadingLevel } from "../common/predicate/has-heading-level";
 import { hasRole } from "../common/predicate/has-role";
+import { isIgnored } from "../common/predicate/is-ignored";
 
-const { and, equals } = Predicate;
+const { equals, not } = Predicate;
```
kasperisager, line 13:
```suggestion
const { equals } = Predicate;
```

kasperisager (changes requested):
Looks great! 👌 Just a minor, erh, editorial?

kasperisager (approved):

PR #581, opened 2020-12-03
Rework the @siteimprove/alfa-table package

kasperisager:
This pull request rewrites the @siteimprove/alfa-table package with a goal of improved maintainability and performance. The functional center piece of the package is an implementation of the table forming algorithm, which has been rewritten to closely match the specification and to reduce the complexity of several computations from quadratic to linear. The end result is an implementation that reads much like the specification and is close to 2 orders of magnitude faster.

Being a complete rewrite with mostly 0 API compatibility with the previous version, these changes are most definitely breaking with too many changes to reasonably list.


kasperisager:
@Jym77 I'd still like some more test cases, but it should otherwise be ready for review. The only outstanding functionality not yet implemented is header assignment for row and column groups.

packages/alfa-table/src/table.ts
```diff
@@ -1,684 +1,854 @@
...
     }
 
-    public get colGroups(): Iterable<ColumnGroup> {
-      return this._table.columnGroups;
+    /**
+     * Get the cells at the given position.
+     */
+    function get(x: number, y: number): Array<number> {
+      ensure(x, y);
+      return table[y][x];
```
Jym77, line 235: The x/y inversion here (and other places) looks super error-prone.
I agree it's what we want, but this is the kind of 2-liner that triggers so many alarms when reading it. I would say we need a comment (either a detailled global one or a small one each time) to remind future selves that this is correct.

packages/alfa-table/src/table.ts
```diff
@@ -1,684 +1,854 @@
...
-        cells,
-        downwardGrowingCells,
-        slots,
-        rowGroups,
-        colGroups,
-        rowGroupHeaders,
-        columnGroupHeaders
-      );
+    function advance(): Option<Element> {
+      children = children.rest();
```
Jym77, line 269: I'm not a huge fan of the global-ish `children` variable modified by side effect, this has potential to blow up in many ways (just like any global variable…)

packages/alfa-table/src/table.ts
```diff
@@ -1,684 +1,854 @@
...
-      const rowGroupHeaders: Array<Cell.Builder> = [];
-      const columnGroupHeaders: Array<Cell.Builder> = [];
+    /**
+     * For all downwards growing cells, extend their height to the current
+     * y-coordinate.
+     *
+     * @see https://html.spec.whatwg.org/#algorithm-for-growing-downward-growing-cells
+     */
+    function growCells(): void {
+      for (let [cell, i] of downwardGrowing) {
```
Jym77, line 559: I'm wondering if `Cell` should have an `id` field that holds `i` since it is essentially what this is.
This would ensure the coupling between the cell and its index and could limit problems in building the `[cell, i]` tuple.

Cells definitely do not have an "index in an array", but it does make sense for them to have an ID (to track a cell through its various instances), and it is just serendipity that Tables are using that ID as an index in an array (or key in an object)…

packages/alfa-table/src/table.ts
```diff
@@ -1,684 +1,854 @@
...
         }
-        if (cell.variant.equals(Some.of(Scope.ColumnGroup))) {
-          columnGroupHeaders.push(cell);
+
+        if (cell.isHeader()) {
+          cells[i] = Cell.header(
+            cell.element,
+            cell.anchor,
+            cell.width,
+            cell.height
```
Jym77, line 573:
```suggestion
            height
```

packages/alfa-table/src/table.ts
```diff
@@ -1,684 +1,854 @@
...
+
+      if (ids.isSome()) {
+        // 3.1
+        // Nothing to do
+
+        // 3.2
+        headers.push(
+          ...ids
+            .get()
+            .collect((id) => Option.from(cells.find(hasElement(hasId(id)))))
```
Jym77, line 604: If the table is huge, this has potential for taking a lot of time (especially if header is on the last row). Could use some cache (not sure if many cells will have the same explicit header or not).
Or maybe a pre-processing to build a map id => cell once and do quick lookup afterwards.

packages/alfa-aria/src/feature.ts
```diff
@@ -657,49 +657,29 @@ const Features: Features = {
...
+              case "column":
+              case "column-group":
+                return Role.of("columnheader");
+
+              case "row":
+              case "row-group":
+                return Role.of("rowheader");
+
+              default:
+                return Role.of("cell");
```
Jym77, line 680: This actually depends on the role of the table element (its `gridcell` if the table is a `grid`).
```suggestion
for (const [role] of Role.from(table)) {
              if (role.isSome()) {
                switch (role.get().name) {
                  case "table":
                    return Role.of("cell");
                  case "grid":
                    return Role.of("gridcell");
                }
              }
            }
```

Jym77 (changes requested):
Looks good. Some minor changes and suggestions.

Somehow, I decided against having the `slots` array pointing to an index in a `cells` array at some point when building this… Guess that was a bad decision 😞

packages/alfa-table/src/table.ts
kasperisager: `fromTable()` defines a lot of mutable local variables, `children` included, as that aligns well with the way the algorithm is specified. It's all local to the function body though and is therefore by no means global. Trying to deviate from the specification to avoid mutable local variables seems like the greater evil to me, not the lesser.

packages/alfa-table/src/table.ts
kasperisager: Only in two places are the table contents actually accessed and those two places reside in functions defined right next to one another. I'll add a comment in both places explaining the "why" of it 👍

packages/alfa-table/src/table.ts
kasperisager: These indices have no use outside of `fromTable()` and so shouldn't leak into the `Cell` class. Beyond that, they're uniquely identifiable within a table by their anchor.

packages/alfa-table/src/table.ts
kasperisager: Good catch!

packages/alfa-table/src/table.ts
kasperisager: That's a great suggestion 👌

packages/alfa-aria/src/feature.ts
kasperisager: Good point, I'll get that cleared up!

packages/alfa-table/src/table.ts
```diff
@@ -1,684 +1,854 @@
...
 
-  public get rowGroups(): Iterable<RowGroup> {
-    return this._rowGroups;
+  public isEmpty(): boolean {
+    return this._cells.length === 0;
   }
 
-  public equals(value: unknown): value is this {
-    if (!(value instanceof Table)) return false;
+  public equals(table: Table): boolean;
```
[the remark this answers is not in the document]
kasperisager: I haven't yet gotten around to adding it to all classes that implement `Equatable`. The reason for adding the first overload is that without it stuff like this happens:

```ts
class Foo implements Equatable {}

const a: Foo;
const b: Foo;

if (a.equals(b)) {
  // `b` is narrowed from `Foo` to `Foo`
} else {
  // `b` is narrowed from `Foo` to `never`
}
```

The overload ensures that if `Foo` is equality tested against `Foo`, no narrowing occurs. Only when `Foo` is equality tested against something that isn't `Foo` does narrowing occur.

packages/alfa-table/src/table.ts
```diff
@@ -1,684 +1,854 @@
...
-    rowGroups: Iterable<RowGroup> = List.empty(),
-    columnGroups: Iterable<ColumnGroup> = List.empty()
+    cells: Iterable<Cell>,
+    groups: Iterable<Group>
   ): Table {
-    return new Table(element, width, height, cells, rowGroups, columnGroups);
+    return new Table(
+      element,
+      Array.from(cells).sort((a, b) => a.compare(b)),
+      Array.from(groups).sort((a, b) => a.compare(b))
```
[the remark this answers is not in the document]
kasperisager: Good point!

Jym77 (dismissed):

kasperisager:
This is ready for another look as I've wrapped up the last bits of functionality and added some more test cases.

Jym77 (approved):
Looks great.
I love the explanations of `data` and `jumps` ❤️

packages/alfa-table/src/table.ts
```diff
@@ -283,18 +340,21 @@ export namespace Table {
...
-        }
+      const cell = cells[i];
 
-        if (!jumps.y.has(y)) {
-          jumps.y.set(y, new Set());
+      if (cell.isHeader()) {
+        if (jumps.x.has(x)) {
+          jumps.x.get(x)!.push(y);
+        } else {
+          jumps.x.set(x, [y]);
```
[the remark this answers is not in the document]
kasperisager: These are built-in `Map` instances though, not those from `@siteimprove/alfa-map`.

packages/alfa-table/src/table.ts
```diff
@@ -306,24 +366,20 @@ export namespace Table {
     }
 
     /**
-     * Determine the position among the non-empty set of candidates that is
-     * lower than the given initial position. If no suitable candidate is found,
+     * Determine the last position among the candidates that is lower than or
+     * equal to the given initial position. If no suitable candidate is found,
```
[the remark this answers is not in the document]
kasperisager: I'll add both 👍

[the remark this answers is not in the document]
kasperisager: It now looks like this:

image

Jym77 (approved):

PR #586, opened 2020-12-04
SIA R90: Element with Presentational Children has no focusable descendants

Jym77:


kasperisager (dismissed):
Looks great! 👏 Just needs an export in `rules.ts` and then it's good to go.

kasperisager (approved):

PR #589, opened 2020-12-07
SIA R61: Ignore headings that are not in the accessibility tree

Jym77:


packages/alfa-rules/test/sia-r61/rule.spec.tsx
```diff
@@ -1,13 +1,15 @@
...
 import { test } from "@siteimprove/alfa-test";
 
-import { Document } from "@siteimprove/alfa-dom";
-
 import R61, { Outcomes } from "../../src/sia-r61/rule";
 
 import { evaluate } from "../common/evaluate";
 import { passed, failed, inapplicable } from "../common/outcome";
 
+const device = Device.standard();
```
kasperisager, line 11: There's no need to pass the standard device parameters to `evaluate()` and this is already the default if none are passed.

kasperisager (changes requested):

packages/alfa-rules/test/sia-r61/rule.spec.tsx
Jym77: Magic!
And I'm passing it so often 🙈

kasperisager: Object destructuring defaults are fantastic 😄 

https://github.com/Siteimprove/alfa/blob/7cfc5be90551302d4fc8beef34f4887f67d739de/packages/alfa-rules/test/common/evaluate.ts#L14-L19

kasperisager (approved):
💯

kasperisager (approved):

PR #591, opened 2020-12-07
Implement the `Shape` CSS value type

kasperisager:
- [x] `Inset`
- [X] `Circle`
- [x] `Ellipse`
- [x] `Polygon`
- ~~`Path`~~
- [x] `Rectangle`


packages/alfa-css/src/value/shape.ts
```diff
@@ -0,0 +1,94 @@
...
+import { Keyword } from "./keyword";
+import { Value } from "../value";
+
+import { Circle } from "./shape/circle";
+import { Inset } from "./shape/inset";
+import { Rectangle } from "./shape/rectangle";
+import { Ellipse } from "./shape/ellipse";
+import { Polygon } from "./shape/polygon";
+
+export class Shape<
```
Jym77, line 12: I'm actually not sure about keeping this class. I thing `Shape` should just be the CSS value `<basic-shape>`. The fact that it goes together with a `<geometry-box>` is imho more of something for the style property `clip-path` to handle.
We may still want an umbrella value for the values of `clip-path`, of course…

kasperisager: Operations on shapes, such as interpolation, go hand-in-hand with their associated reference box. When https://drafts.csswg.org/css-shapes/ talks about "shapes", it's most often referring to the combination of a basic shape and a reference box.

Jym77: Yes, let me rephrase it a bit 😁 
I think the implementation of `<basic-shape>` is a large enough chunk that we can review/merge that in and move on the implementation of the full shape. That does make the class `Shape` here not really part of the implementation of basic shapes and maybe could be extracted to the next PR…

(also, that does imply that maybe the basic shapes should have a `type` of `"basic-shape"`… 🤔 )

kasperisager:
I've had a go at the code so I'll leave the final sign-off to you, @Jym77 😄

Jym77 (approved):
Makes sense.
I'm really annoyed that we don't have import order rule in linting, this is always creating a lot of manual work 😕 (and I've already spent some time ordering them beforehand…)

PR #600, opened 2020-12-09
Improve performance of `:nth-*` selectors

### LINKED ISSUE #582: Improve performance of child index selector matching
Matching of child index selectors currently performs very poorly, which is especially problematic for large tables that alternate styling between even and odd rows. Child index selectors are currently matched somewhat naïvely in that indices are always recomputed on every match, resulting in a complexity of `O(n^2)` with `n` being the number of child elements relevant for the selector. In Servo, this is reduced to `O(n)` by keeping a cache of indices: https://github.com/servo/servo/blob/master/components/selectors/nth_index_cache.rs

kasperisager:
Closes #582.

### DIFF
--- a/packages/alfa-selector/src/selector.ts
+++ b/packages/alfa-selector/src/selector.ts
@@ -1249,6 +1249,8 @@ export namespace Selector {
       return new NthChild(index);
     }
 
+    private static readonly _indices = new WeakMap<Element, number>();
+
     private readonly _index: Nth;
 
     private constructor(index: Nth) {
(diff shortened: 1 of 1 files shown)


packages/alfa-selector/src/selector.ts
```diff
@@ -1249,6 +1249,8 @@ export namespace Selector {
       return new NthChild(index);
     }
 
+    private static readonly _indices = new WeakMap<Element, number>();
```
Jym77, line 1252: Not sure why this is directly a weak map rather than a cache (which seems to be mostly a wrapper for weak map).

I kinda guess this is because it's filled in bulk (in `matches`) rather than one by one when a cache miss happens.
I would argue using a cache would make the intention clearer.

Jym77 (changes requested):

packages/alfa-selector/src/selector.ts
kasperisager: `Cache` does provide some conveniences, but those come at a cost. Not a great cost, but a non-zero cost such as wrapping values in `Option`. As selector matching is one of our hottest paths, avoiding allocations like that is desirable and simple.

Jym77: Fair point 👍

Jym77 (approved):

PR #616, opened 2020-12-17
Account for additional techniques for hiding content in `isVisible()`

### LINKED ISSUE #519: `isVisible()` does not account for elements pulled offscreen via `position`

```html
<style>
  .hidden {
    position: absolute;
    left: -9999px;
  }
</style>

```

#531 is needed for this.

- [ ] Follow up on DEV-11054.

### LINKED ISSUE #524: `isVisible()` does not account for text hidden via `text-indent`

```html
<style>
  .hidden {
    overflow: hidden;
    white-space: nowrap;
    text-indent: 100%;
  }
</style>

```

#532 is needed for this.

- [ ] Follow up on DEV-10828.
- [ ] Follow up on DEV-10887.

kasperisager:
Closes #519.
Closes #524.


packages/alfa-rules/src/common/predicate/is-offscreen.ts
```diff
@@ -0,0 +1,46 @@
...
+          const { value: position } = style.computed("position");
+          const { value: left } = style.computed("left");
+          const { value: right } = style.computed("right");
+
+          if (position.value !== "static") {
+            for (const inset of [left, right]) {
+              switch (inset.type) {
+                case "length":
+                  return abs(inset.value) >= 9999;
+              }
```
Jym77, line 34: Why not looking also at `top` (I guess `bottom` is less doable due to infinity vertical scrolling)? (`position: absolute; top: -9999px` seems to be the preferred offscreen way in ACT rules at least).

packages/alfa-rules/src/common/predicate/is-offscreen.ts
```diff
@@ -0,0 +1,46 @@
...
+
+          const { value: position } = style.computed("position");
+          const { value: left } = style.computed("left");
+          const { value: right } = style.computed("right");
+
+          if (position.value !== "static") {
+            for (const inset of [left, right]) {
+              switch (inset.type) {
+                case "length":
+                  return abs(inset.value) >= 9999;
```
Jym77, line 33: Also, obvious prone to errors on huge displays, or with huge elements… Not sure we can do much better without a layout system.
(also, ACT rules is having `-999em` in at least two cases, which we won't catch here)

packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -6,107 +6,76 @@ import { Refinement } from "@siteimprove/alfa-refinement";
...
       and(
         isElement,
         and(not(isReplaced), (element) =>
           every(
-            element.children({ nested: true, flattened: true }),
+            element.children({
+              nested: true,
+              flattened: true,
+            }),
             not(isVisible(device, context))
```
Jym77, line 75: For the sake of readability of `isVisible`, I think we can go further in the subfunctions and extract these as two separate predicates.

packages/alfa-rules/test/common/predicate/is-visible.spec.tsx
```diff
@@ -94,3 +94,229 @@ test("isVisible() returns true for replaced elements with no child", (t) => {
...
+  const text = h.text("");
+
+  t.equal(isVisible(text), false);
+});
+
+test("isVisible() returns false for a text node with no data", (t) => {
+  const text = h.text("");
+
+  t.equal(isVisible(text), false);
+});
```
Jym77, line 172: Duplicate test.
```suggestion
test("isVisible() returns false for a text node with no data", (t) => {
  const text = h.text("");

  t.equal(isVisible(text), false);
});
```

packages/alfa-rules/test/common/predicate/is-visible.spec.tsx
```diff
@@ -94,3 +94,229 @@ test("isVisible() returns true for replaced elements with no child", (t) => {
...
+
+test("isVisible() returns false for a text node with only whitespace", (t) => {
+  const text = h.text(" ");
+
+  t.equal(isVisible(text), false);
+});
+
+test("isVisible() returns false for a text node with a transparent color", (t) => {
```
Jym77, line 180: For this and the following examples, I'm not sure the `<div>` element gets the same result as its text 🤔

Jym77 (changes requested):

packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -0,0 +1,74 @@
...
+              whitespace.value === "nowrap"
+            ) {
+              const { value: indent } = style.computed("text-indent");
+
+              switch (indent.type) {
+                case "percentage":
+                  return abs(indent.value) >= 1;
+
+                case "length":
+                  return abs(indent.value) >= 999;
```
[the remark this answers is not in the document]
kasperisager: It's a heuristic for lack of a layout system. If we had a way of knowing the width of the element, we'd of course be able to account for that.

packages/alfa-rules/src/common/predicate/is-offscreen.ts
kasperisager: As most pages grow downwards, not sideways, content should be offscreen'ed using `left` or `right` as the width of the viewport can be expected to remain constant. This article also talks about the perils of offscreen'ing using `top` and uses `left` in its examples: http://accessibilitytips.com/2008/03/04/positioning-content-offscreen/. This article from WebAIM uses `left` as well: https://webaim.org/techniques/css/invisiblecontent/

packages/alfa-rules/src/common/predicate/is-offscreen.ts
kasperisager: These are computed `px` values so `em` values will have been resolved by the time this condition is reached. `-999em` can be expected to compute to at least `-9999px` for reasonable font sizes; all it takes is `10px`, after all.

packages/alfa-rules/test/common/predicate/is-visible.spec.tsx
kasperisager: Very good point 👍 I've adjusted the test cases.

packages/alfa-rules/src/common/predicate/is-visible.ts
kasperisager: Agreed, though I'll leave that to a separate task.

packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -0,0 +1,74 @@
...
+          const { value: height } = style.computed("height");
+          const { value: width } = style.computed("width");
+          const { value: x } = style.computed("overflow-x");
+          const { value: y } = style.computed("overflow-y");
+
+          if (
+            height.type === "length" &&
+            height.value <= 1 &&
+            width.type === "length" &&
+            height.value <= 1 &&
```
Jym77, line 35:
```suggestion
            width.value <= 1 &&
```
🙈

packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -0,0 +1,74 @@
...
+        if (isElement(node)) {
+          const style = Style.from(node, device, context);
+
+          const { value: height } = style.computed("height");
+          const { value: width } = style.computed("width");
+          const { value: x } = style.computed("overflow-x");
+          const { value: y } = style.computed("overflow-y");
+
+          if (
+            height.type === "length" &&
```
Jym77, line 32: Also, wondering whether we should check for `height: 0%`. But maybe that's not really used for clipping.

Jym77 (changes requested):

packages/alfa-rules/src/common/predicate/is-clipped.ts
kasperisager: Great catch! :facepalm:

packages/alfa-rules/src/common/predicate/is-clipped.ts
kasperisager: Good point!

Jym77 (approved):

PR #617, opened 2020-12-17
Account for `<option>` elements in `isRendered()`

kasperisager:
- [ ] Follow up on DEV-10945.

### DIFF
--- a/packages/alfa-rules/src/common/predicate/is-rendered.ts
+++ b/packages/alfa-rules/src/common/predicate/is-rendered.ts
@@ -27,8 +27,20 @@ export function isRendered(
           return false;
         }
 
+        // <option> elements that are descendants of <select> elements will not
+        // be rendered as only their label is used during rendering.
+        // https://html.spec.whatwg.org/#the-select-element-2
         if (
-          Element.isElement(node) &&
+          hasName("option") &&
+          // <option> elements might be nested within <optgroup> elements so we
+          // need to look at the first two ancestors.
+          node.ancestors().filter(isElement).take(2).some(hasName("select"))
+        ) {
+          return false;
+        }
+
+        if (
+          isElement(node) &&
           Style.from(node, device, context)
             .computed("display")
             .some(([outside]) => outside.value === "none")


packages/alfa-rules/src/common/predicate/is-rendered.ts
```diff
@@ -27,8 +27,20 @@ export function isRendered(
           return false;
         }
 
+        // <option> elements that are descendants of <select> elements will not
```
Jym77, line 30: `<optgroup>` under `<select>` seem to be treated the same ay (rendered as label).

packages/alfa-rules/src/common/predicate/is-rendered.ts
```diff
@@ -27,8 +27,20 @@ export function isRendered(
...
+        // https://html.spec.whatwg.org/#the-select-element-2
         if (
-          Element.isElement(node) &&
+          hasName("option") &&
+          // <option> elements might be nested within <optgroup> elements so we
+          // need to look at the first two ancestors.
+          node.ancestors().filter(isElement).take(2).some(hasName("select"))
+        ) {
+          return false;
+        }
```
Jym77, line 40: 🤔 Won't that result in the text node in `<select><option>Foo</option></select>` being seen as not rendered?

Jym77 (changes requested):

packages/alfa-rules/src/common/predicate/is-rendered.ts
kasperisager: Yes, which is also intentional. The text within `<option>` elements isn't itself rendered, but is instead used as the label of the corresponding `<select>` element. Stuff like this, for example, has no effect:

```css
option {
  color: red;
}
```

That won't make the text rendered within the `<select>` element red; this will, however:

```css
select {
  color: red;
}
```

packages/alfa-rules/src/common/predicate/is-rendered.ts
kasperisager: Good point 👍

packages/alfa-rules/src/common/predicate/is-rendered.ts
Jym77: 🤔 but when the select is expanded, the various options are displayed individually, and in that case, the styling on  the option element does apply.
```html
<select style="color: green;">
  <option style="color: blue;">foo</option>
  <option style="color: red;">bar</option>
</select>
```
has its options in various colors when expanded (and its own value in green when collapsed).

OTOH, all this is apparently handled in shadow tree…
Plus, I'm not sure how much we can detect whether the select is expanded or not…

kasperisager: That seems UA specific then because this is what I'm seeing:

image

kasperisager: The plot thickens... The colours are applied on my end when I make it a multiselect:

image

This evidently needs some more refinement 🤔

Jym77: select
This is what Chrome shows me…

Jym77: > This evidently needs some more refinement 🤔

But in the meantime we might be safe-ish assuming that the select is collapsed since this is likely what Robin gives us currently 🤔

kasperisager:
I'll close this one as it has become apparent that it's not really clear how to best handle `<option>` elements in `isRendered()`. For now, the current behaviour isn't causing issues that can't be solved elsewhere.

Thanks so much for the thorough review, @Jym77! 🙏

PR #637, opened 2021-01-05
Keep source of explicit `initial` properties

Jym77:
Properties that are explicitly declared as `initial` were still recorded with no source. Among other, this causes Alfa to loose the `!important` flag in declarations like `<div style="foo: initial !important">` and therefore to incorrectly pass https://act-rules.github.io/rules/78fd32#failed-example-6

Testing on 
```html
<html lang="en">
	<style>
		p { background-color: initial !important;}
	</style>
<body>
	<p>White</p>
	<p style="background-color: blue;">Still white</p>
</body>
</html>
```
shows that the important flag on an explicit `initial` property is indeed respected.


kasperisager (approved):
Great catch! 💯

PR #640, opened 2021-01-06
Add `--cpu-profile` and `--heap-profile` flags to `alfa audit` command

kasperisager:


packages/alfa-cli/bin/alfa/profiler.ts
```diff
@@ -0,0 +1,59 @@
+import * as inspector from "inspector";
```
Jym77, line 1: Is the inspector file missing from the PR?

Jym77 (changes requested):

packages/alfa-cli/bin/alfa/profiler.ts
kasperisager: It's a Node.js built-in module: https://nodejs.org/api/inspector.html

Jym77: 🙈

Jym77 (approved):

PR #643, opened 2021-01-07
Add @siteimprove/alfa-performance package

kasperisager:
This pull request introduces a new package, @siteimprove/alfa-performance, for working with performance measurements. Its design is based on the [User Timing](https://www.w3.org/TR/user-timing/) API, but is implemented independently of runtime environment (browsers vs. Node.js) and aims to solve some of the shortcomings of the User Timing API:

- **Lack of memory bounds**: With the User Timing API, performance entries are added to and kept in an internal buffer whose size can grow without bound. As such, marks and measurements must be cleared out regularly, or disabled entirely, in long running processes to avoid running out of memory.

- **Everything is globally scoped**: With the User Timing API, any code that wishes to track performance measurements do so via a globally scoped `Performance` instance. While this aids discoverability for consumers that sit outside the main code path, such as Real User Measurement (RUM) collectors, it makes it impossible so do things like collecting performance measurements for a particular call tree only, such as a single audit of a given rule set. 

  This problem also relates to the previous point as a singleton `Performance` instance, and its associated performance entries, can never be garbage collected. With locally scoped `Performance` instances, one would be able to simply drop references to these instances when no longer needed and then have the garbage collector do its job.

The implementation of `Performance` introduced in this pull request solves these problems in the following way:

- No global `Performance` instance is available and specific `Performance` instances must be created through `Performance.of()`:

  ```ts
  const performance = Performance.of<string>();
  ```

  Note that keys are not restricted to the `string` type but can be arbitrarily chosen.

- Entries are never buffered, but are instead shipped straight to listeners. To perform measurements between marks, the marks must be passed explicitly:

  ```ts
  const start = performance.mark("start");
  const end = performance.mark("end");
  
  performance.measure("measurement", start, end);
  ```

### DIFF
--- a/packages/alfa-performance/src/performance.ts
+++ b/packages/alfa-performance/src/performance.ts
@@ -31,8 +31,15 @@ export class Performance<T>
     return now ? now - this._epoch : this.now();
   }
 
-  public mark(data: T): Performance.Mark<T> {
-    return this._emit(Performance.mark(data, this.now()));
+  public mark(
+    data: T,
+    start: Performance.Entry<T> | number = this.now()
+  ): Performance.Mark<T> {
+    if (typeof start !== "number") {
+      start = start.start;
+    }
+
+    return this._emit(Performance.mark(data, start));
   }
 
   public measure(
--- a/packages/alfa-performance/src/performance.ts
+++ b/packages/alfa-performance/src/performance.ts
@@ -65,6 +65,10 @@ export class Performance
     this._emitter.on(listener);
   }
 
+  public off(listener: Callback<Performance.Entry>): void {
+    this._emitter.off(listener);
+  }
+
   public asyncIterator(): AsyncIterator<Performance.Entry> {
     return this._emitter.asyncIterator();
   }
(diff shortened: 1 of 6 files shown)


packages/alfa-performance/src/performance.ts
```diff
@@ -0,0 +1,211 @@
...
+
+import * as json from "@siteimprove/alfa-json";
+
+import { now } from "./now";
+
+export class Performance<T>
+  implements
+    AsyncIterable<Performance.Entry<T>>,
+    Serializable<Performance.JSON> {
+  public static of<T>(): Performance<T> {
```
Jym77, line 13: Should `T` default to `string` which sounds like it's going to be a frequent case?
(would simplify calls by removing need to provide the type each time)

packages/alfa-performance/src/performance.ts
```diff
@@ -0,0 +1,211 @@
...
+    return value instanceof Performance;
+  }
+
+  export type Entry<T> = Mark<T> | Measure<T>;
+
+  export namespace Entry {
+    export type JSON<T> = Mark.JSON<T> | Measure.JSON<T>;
+  }
+
+  export class Mark<T> implements Serializable<Mark.JSON<T>> {
```
Jym77, line 116: Should this (and `Measure`) include a differentiating `type` (or similar) method?

Jym77 (approved):
Looks good. Some possible convenience improvements.

kasperisager:
Great suggestions, thanks! 🙏

Jym77 (approved):

PR #644, opened 2021-01-07
Make `#toJSON()` parametric for all container types

kasperisager:
This pull request makes the `#toJSON()` methods parametric for all container types to allow strongly typed JSON serialisation when working with these. The most common use case of this is by far when writing unit tests as these often do deep comparisons of JSON serialised structures. Consider, for example, a list of optional numbers:

```ts
const list: List<Option<number>>
```

Previously, when serialising this structure to JSON, you'd get the following type:

```ts
const json: List.JSON = list.toJSON()
```

Not... terribly helpful. It even becomes harmful when refactoring as TypeScript has no knowledge of the `Option.JSON` types nested within the serialised list structure. Had we typed out the structure by hand, as we often do when doing deep comparisons, TypeScript would not have applied the refactoring. This piece of code would therefore stay the same if had we chosen to rename `Some.JSON#value` to `Some.JSON#whatever`:

```ts
const json: List.JSON = [
  {
    type: "some",
    value: 42
  }
]
```

The result we really want is for TypeScript to intelligently apply to refactor to the above code as well, which this pull request enables by making the `List.JSON` type, and all other container types, parametric:

```ts
const json: List.JSON<Option.JSON<number>> = [
  {
    type: "some",
    value: 42
  }
]
```

If we now choose to rename `Some.JSON#value` to `Some.JSON#whatever`, the refactor will be applied to the above code as well as TypeScript knows that the inner object is of type `Some.JSON<number>`.

### DIFF
--- a/packages/alfa-json/src/serializable.ts
+++ b/packages/alfa-json/src/serializable.ts
@@ -20,8 +20,8 @@ export interface Serializable<T extends JSON = JSON> {
 }
 
 export namespace Serializable {
-  export type ToJSON<T> = T extends Serializable<infer T>
-    ? T
+  export type ToJSON<T> = T extends Serializable<infer U>
+    ? U
     : T extends JSON
     ? T
     : JSON;
--- a/packages/alfa-json/src/serializable.ts
+++ b/packages/alfa-json/src/serializable.ts
@@ -20,6 +20,12 @@ export interface Serializable<T extends JSON = JSON> {
 }
 
 export namespace Serializable {
+  export type ToJSON<T> = T extends Serializable<infer T>
+    ? T
+    : T extends JSON
+    ? T
+    : JSON;
+
   export function isSerializable<T extends JSON>(
     value: unknown
   ): value is Serializable<T> {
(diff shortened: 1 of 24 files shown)


packages/alfa-json/src/serializable.ts
```diff
@@ -20,6 +20,12 @@ export interface Serializable<T extends JSON = JSON> {
 }
 
 export namespace Serializable {
+  export type ToJSON<T> = T extends Serializable<infer T>
+    ? T
+    : T extends JSON
+    ? T
+    : JSON;
+
```
Jym77, line 28: Wait, are there two `T` here? One introduced by `ToJSON<T>` and one by `infer T`.
We should name them differently, especially since both `? T` then correspond to different `T` (if I understand correctly), which is definitely confusing…

packages/alfa-json/src/serializable.ts
```diff
@@ -20,6 +20,12 @@ export interface Serializable<T extends JSON = JSON> {
 }
 
 export namespace Serializable {
+  export type ToJSON<T> = T extends Serializable<infer T>
```
Jym77, line 23: Shouldn't the type be called `JSON<T>` rather than `ToJSON<T>`?
As far as I can tell, this is the type of serialised objects (JSON), not the type of serialising function (to JSON).

Jym77 (changes requested):
Stronger types ❤️

packages/alfa-json/src/serializable.ts
kasperisager: It's basically the type-level equivalent of the `Serializable.toJSON()` function as it transforms a type `T` into its corresponding JSON serialised type.

Jym77 (approved):

PR #648, opened 2021-01-12
Make SIA-R14 assume that non-essential text is imperceivable

kasperisager:
This pull request changes the expectation of SIA-R14 to assume that non-essential text, such as purely presentational text, has been made imperceivable. As such, imperceivable text is no longer required to be included in the accessible name of widget elements that receive their name from their contents.


Jym77 (approved):
Makes sense 👍

PR #650, opened 2021-01-13
Fix parsing of <position> CSS values

Jym77:
Needed for #591.

I've been hitting pretty strong with the test hammer as this is tricky and has a lot of corner cases so I want to limit the risks of regression…


kasperisager (approved):
Looks good 👍 I made some adjustments and switched the labels as these changes touch public API in a backwards-incompatible manner.

PR #665, opened 2021-01-21
Implement SIA R66

Jym77:
It wasn't obvious to refactor the applicability an expectation since it isn't obvious to have a good name for what they do 😕


packages/alfa-rules/src/common/expectation/get-colors.ts
```diff
@@ -0,0 +1,227 @@
...
+import { Current, Percentage, RGB, System } from "@siteimprove/alfa-css";
+import { Style } from "@siteimprove/alfa-style";
+import { Iterable } from "@siteimprove/alfa-iterable";
+
+const { map } = Iterable;
+
+/**
+ * Determine the approximate foreground color of an element if possible.
+ */
+export function getForeground(
```
Jym77, line 13: I decided against splitting `getFG` and `getBG` in separate files since they both depend on the same bits that would then need to be shared…

packages/alfa-rules/src/common/outcome/contrast.ts
```diff
@@ -0,0 +1,32 @@
+import { Contrast } from "../diagnostic/contrast";
+import { Err, Ok } from "@siteimprove/alfa-result";
+
+export namespace Outcomes {
```
Jym77, line 4: It would better be named `Contrast`, but that is already used for the `Diagnostic` class, and since these two are almost always imported together, that need renaming. And if we rename something every time we import it, it probably means that the name isn't great…

Still, not happy of having this just called `Outcomes` 😞

kasperisager: This should do the trick:

```ts
import { Contrast as ContrastDiagnostic } from "../diagnostic/contrast";
```

Jym77: Yes, of course.
What stopped me is that this will be needed at the 3 places where we import that (here and the two test files).
If we have to rename it every time we import it, it feels like the name was wrong in the first place 😕

kasperisager: > If we have to rename it every time we import it, it feels like the name was wrong in the first place 😕 

The beauty of modules is that they don't need to concern themselves with the world around them. They should choose the best possible export names _taking only their own context into account_. Importing modules can always rename individual imports or even namespace entire imported modules to avoid collisions _in their own context_.

"Effective Go" has a similar philosophy on package and export names: https://golang.org/doc/effective_go.html#package-names

kasperisager (approved):
Looks great! 👏 We can always refactor the internals as clearer patterns hopefully start to show for shared expectations.

PR #667, opened 2021-01-25
Make `Value.JSON` optionally parametric

Jym77:
In the line of #644, I figured `Value.JSON` could as well incorporate the type of value.

Not sure if we should extend further for sub-interfaces with common field (e.g. `format` for colours or `kind` for transforms).


kasperisager:
The type parameter `T` in `Value.JSON<T>` has a default value of string and so isn't required which maintains backwards compatibility. Swapping labels to minor.

kasperisager (approved):

PR #670, opened 2021-01-25
Rework `Hash` as an abstract class

kasperisager:
This pull request reworks the `Hash` interface as an abstract class that concrete hash algorithms, such as `FNV`, must now extend. Among other things, this allows for a chained API, making it much simpler to hash structures with many fields:

```ts
hash
  .writeString(foo.name)
  .writeNumber(foo.age)
  .writeBoolean(foo.isAwesome)
  .writeUnknown(foo.metadata);
```


packages/alfa-css/src/value/shadow.ts
```diff
@@ -99,7 +99,7 @@ export class Shadow<
     this._blur.hash(hash);
     this._spread.hash(hash);
     this._color.hash(hash);
-    Hash.writeBoolean(hash, this._isInset);
+    hash.writeBoolean(this._isInset);
```
Jym77, line 102: Would it makes sense to have that also chainable, likely with a
```javascript
writeHashable(foo: Hashable) {
  foo.hash(this);
  return this;
}
```
and then we can turn this (and similar) to
```javascript
hash.writeHashable(this._blur).writheHashable(this._spread).writeHashable(this._color).writeBoolean(this.isInset)
```

Jym77 (approved):

packages/alfa-css/src/value/shadow.ts
kasperisager: Oh, I actually already added that but forgot to use it:

https://github.com/Siteimprove/alfa/blob/4d3955dfa86b632f7153998b0b86005ad949e090/packages/alfa-hash/src/hash.ts#L104-L107

Jym77 (approved):

PR #675, opened 2021-01-28
SIA R7 ignore invisible text and fully overwritten attributes

### LINKED ISSUE #623: SIA R7 should ignore invisible text and fully overwritten attributes
- [x] https://github.com/act-rules/act-rules.github.io/pull/1383 wasn't backported. Notably, Alfa incorrectly fails `<span lang="invalid" hidden>`.

- [x] Alfa is also hit by https://github.com/act-rules/act-rules.github.io/issues/1525 : `<div lang="invalid"><span lang="en">Foo</span></div>` shouldn't fail the `<div>`.

vemolista:
Resolves #623


vemolista:
I'm guessing there is a way to do fewer passes through the tree, but it's a start 🚀

packages/alfa-rules/src/sia-r7/rule.ts
```diff
@@ -3,22 +3,26 @@ import { Attribute, Element, Namespace } from "@siteimprove/alfa-dom";
...
+import { hasDescendant, isIgnored } from "../common/predicate";
 
 import { hasAttribute } from "../common/predicate/has-attribute";
+import { isVisible } from "../common/predicate/is-visible";
 
 const { isElement, hasName, hasNamespace } = Element;
 const { isEmpty } = Iterable;
-const { and, not } = Predicate;
+const { and, not, or } = Predicate;
+const { and: andRefinement } = Refinement;
```
kasperisager, line 20:
```suggestion
const { not, or } = Predicate;
const { and } = Refinement;
```

When passed "normal" predicates, `Refinement.and()` behaves like `Predicate.and()`.

packages/alfa-rules/src/sia-r7/rule.ts
```diff
@@ -29,7 +33,19 @@ export default Rule.Atomic.of<Page, Attribute>({
             body
               .descendants()
               .filter(isElement)
+              .filter(or(isVisible(device), not(isIgnored(device))))
               .filter(hasAttribute("lang", not(isEmpty)))
+              .filter(
+                not(
+                  hasDescendant(
+                    andRefinement(
```
kasperisager, line 41:
```suggestion
                    and(
```

kasperisager (changes requested):
Solid progress, kudos! 👏 There are a couple of issues with the point on "overwritten attributes" though:

1.  As you note, the time complexity is indeed very high in the worst case. In fact, it's quadratic. As you also mention, the ideal scenario is doing just a single pass over the tree, which would also help solve my second point.

2.  The interpretation of "overwritten attribute" is not entirely correct. What we're really interested in is all `lang` attributes that affect text, not `lang` attributes specified on elements without descendants that don't also have a `lang` attribute. The latter, and current interpretation, will behave incorrectly for a case like this:

    ```html
    <div lang="invalid">
      Hello <span lang="en">world</span>
    </div>
    ```

    `lang="invalid"` affects the text `Hello`, but would currently be considered inapplicable. An algorithm for finding these text nodes in linear time looks like this:

    1.  Let `ancestorLang` be `null`.
    2.  Given a `Node` `node`:
        1. If `node` is an `Element` and has a `lang` attribute `lang`, let `ancestorLang` be `lang`.
        2. If `node` is a `Text` and `ancestorLang` is not `null`, `ancestorLang` is an applicable attribute.
        3. For each child `child` of `node`, repeat step ii. with `child` as `node`.

    The remaining applicability conditions of course still apply.

packages/alfa-rules/src/sia-r7/rule.ts
```diff
@@ -1,37 +1,71 @@
...
-              .filter(isElement)
-              .filter(hasAttribute("lang", not(isEmpty)))
-              .map((element) => element.attribute("lang").get())
-          );
+        return visit(
+          document
+            .descendants()
+            .find(isElement)
+            .filter(and(hasNamespace(Namespace.HTML), hasName("body")))
+            .get(),
```
kasperisager, line 41: This will cause a panic when used for documents whose first descendant element is not a `<body>` element. Something along the lines of the following avoids that:

```ts
document
  // Get all descendants of the document, including those in shadow roots
  // and embedded documents
  .descendants({
    composed: true,
    nested: true,
  })
  // Remove all descendants that aren't elements
  .filter(isElement)
  // Find all `<body>` elements among the descendants
  .filter(and(hasNamespace(Namespace.HTML), hasName("body")))
  // Visit each `<body>` element and collect applicable `lang` attributes
  .flatMap(element => Sequence.from(visit(element, None)))
  // Remove `lang` attributes that may have been yielded multiple times
  .distinct();
```

packages/alfa-rules/src/sia-r7/rule.ts
```diff
@@ -1,37 +1,71 @@
...
+            test(
+              and(isText, or(isVisible(device), not(isIgnored(device)))),
+              node
+            ) &&
+            lang !== None
+          ) {
+            yield lang.get();
+          }
+
+          const children = node.children({ flattened: true, nested: true });
```
kasperisager, line 63:
```suggestion
          const children = node.children({ flattened: true });
```

With the adjustments to the applicability above, we just need to traverse the flattened subtree of every `<body>` element.

kasperisager (changes requested):

packages/alfa-rules/src/sia-r7/rule.ts
```diff
@@ -1,37 +1,70 @@
...
+
+          if (
+            test(
+              and(isText, or(isVisible(device), not(isIgnored(device)))),
+              node
+            ) &&
+            lang !== None
+          ) {
+            yield lang.get();
+          }
```
kasperisager, line 53:
```suggestion
          if (
            test(
              and(isText, or(isVisible(device), not(isIgnored(device)))),
              node
            )
          ) {
            yield* lang;
          }
```

`Option<T>` implements `Iterable<T>` so the above will yield nothing if `lang` is `None` and will yield an `Attribute` if `lang` is `Some`.

kasperisager (changes requested):
Looks great! 👏 Just a small tweak and then I think this is good to go 👍

kasperisager (approved):
Beautiful! 👌

PR #676, opened 2021-01-29
Implement cannot-be-a-base-url flag

### LINKED ISSUE #459: `URL#toString()` is incorrect for URLs that cannot be used as base URLs
URL parsing and stringification do not handle the `cannot be a base URL` flag (as far as I understand).
This result in `{ scheme: "about", path: ["blank"] }` to be stringified as `about:/blank` instead of `about:blank`.

vemolista:
Resolves #459 

Spec:
https://url.spec.whatwg.org/#concept-base-url
https://url.spec.whatwg.org/#url-serializing


kasperisager:
I dug through the spec again and sorted out the (hopefully) last quirks so I'll punt this back to you, @vemolista. If there's nothing else you'd like to add then feel to send this off for review.

packages/alfa-url/src/url.ts
```diff
@@ -34,10 +35,25 @@ export class URL implements Equatable, Hashable, Serializable<URL.JSON> {
...
+    return URL.parse("https://example.com").get();
+  }
+
+  /**
+   * @see https://tools.ietf.org/html/rfc6694#section-3
+   */
+  public static blank(): URL {
+    return URL.parse("about:blank").get();
+  }
+
```
Jym77, line 56: These two seem to be unused 🤔

packages/alfa-url/src/url.ts
```diff
@@ -354,11 +400,33 @@ export namespace URL {
           // "#" which we need to remove.
           Option.of(hash)
             .reject(isEmpty)
-            .map((hash) => hash.replace(/^#/, ""))
+            .map((hash) => hash.replace(/^#/, "")),
+
+          // The URL cannot be used as a base URL when the scheme either isn't
+          // special or the pathname doesn't start with a leading "/".
+          !(isSpecialScheme(scheme) || pathname[0] === "/")
```
Jym77, line 407: Comment and code don't match.
Code test "scheme not special **and** path doesn't start with /" but comment says **or**.
Given that tests are working, I guess that code is correct. Thus said, I also suggest to De Morgan-ise it to better match the comment.
```suggestion
          // The URL cannot be used as a base URL when the scheme isn't
          // special and the pathname doesn't start with a leading "/".
          !isSpecialScheme(scheme) && pathname[0] !== "/"
```

Jym77 (changes requested):
Small editorial cleanup.

packages/alfa-url/src/url.ts
kasperisager: `URL.blank()` currently is, but `URL.example()` isn't.

packages/alfa-url/src/url.ts
kasperisager: Good catch!

Jym77 (dismissed):

kasperisager (approved):

kasperisager (approved):

Issue #298, opened 2020-07-28
Elements whose implicit role depends on the presence of an accessible name are not correctly handled


kasperisager:
An example is the `<section>` element which should have an implicit role of `region` _if_ the element also has an accessible name: https://w3c.github.io/html-aam/#el-section. Alfa does not yet handle cases like these correctly though:

https://github.com/Siteimprove/alfa/blob/45e77ac69e6547c4f13273ac36761318108d0574/packages/alfa-aria/src/feature.ts#L511-L514

The reason is the circularity between the role and accessible name computations; to determine the role of an element we might need to first determine its accessible name and to determine its accessible name we might need to first determine its role. The best way to deal with this seems to be to make it possible to break the accessible name computation into configurable steps to allow callers to only enable a subset of the steps taken in determining the accessible name.

kasperisager:
Surprisingly, Chrome also fails to handle this case correctly and exposes `<section>` elements with a role of `region` regardless of whether or not they have an accessible name. Safari never assigns implicit roles to `<section>` elements while Firefox acts according to spec. Browsers... 🤦

kasperisager:
This _may_ be influenced by #307.

kasperisager:
It appears that Chrome might actually just be dealing with this at a different layer, possibly to get around the same issue as we're facing? There's a method `BrowserAccessibility::GetLocalizedStringForLandmarkType()`, which I imagine gets the localised string to announce for landmark types, that does a name check for the `region` and `section` (equivalent to `region` in Chromium it seems) roles: [`content/browser/accessibility/browser_accessibility.cc#L1892-L1895`](https://github.com/chromium/chromium/blob/4d3e3190a79f45699aed9c02cb65c1776368fb7f/content/browser/accessibility/browser_accessibility.cc#L1892-L1895)

kasperisager:
Here are some test cases to try out:

1.  ```html
    <section>Hello world</section>
    ```

2.  ```html
    <section aria-label>Hello world</section>
    ```

3.  ```html
    <section aria-label=" ">Hello world</section>
    ```

4.  ```html
    <section aria-label="Label">Hello world</section>
    ```

Results on my end:

1.  ```html
    <section>Hello world</section>
    ```

    **Chrome + VoiceOver**: "hello world"
    **Safari + VoiceOver**: "hello world"
    **Firefox + VoiceOver**: "hello world"

2.  ```html
    <section aria-label>Hello world</section>
    ```

[...]

PR #678, opened 2021-02-01
Limit SIA-R40 to explicit `region` roles

kasperisager:
See https://github.com/Siteimprove/alfa/issues/298#issuecomment-770720526.

- [ ] Follow up on DEV-11187.


Jym77 (approved):
I agree that this does not really solve the problem but is good enough for now…

PR #687, opened 2021-02-04
Treat content clipped by masking as invisible.

### LINKED ISSUE #467: `isVisible()` does not account for text hidden using the `clip` property

```html
<style>
  .hidden {
    position: absolute;
    clip: rect(1px, 1px, 1px, 1px);
  }
</style>

```

#533 is needed for this.

- [ ] Follow up on DEV-10711.
- [ ] Follow up on DEV-11075.

Jym77:
Resolves #467.

Probably not the most complete solution, but it should handle our main use case…


kasperisager:
@Jym77 This looks like it might need a little rebase 🙈

Jym77:
> @Jym77 This looks like it might need a little rebase 🙈

🙈

PR #688, opened 2021-02-04
Treat content clipped by masking as invisible

### LINKED ISSUE #467: `isVisible()` does not account for text hidden using the `clip` property

Jym77:
Resolves #467.

Probably not the most complete solution, but it should handle our main use case…


packages/alfa-rules/src/common/predicate/is-clipped-by-size.ts
```diff
@@ -11,7 +11,7 @@ const { isText } = Text;
 
 const cache = Cache.empty<Device, Cache<Context, Cache<Node, boolean>>>();
 
-export function isClipped(
+export function isClippedBySize(
```
kasperisager, line 14: Splitting these in two opens up the possibility to use one without the other, which really shouldn't ever happen. You can put them in the same module as two functions and then export an `or()` of the two.

kasperisager (changes requested):

kasperisager (approved):

PR #693, opened 2021-02-08
Support `font` shorthand property

### LINKED ISSUE #682: Implement `font` CSS property
https://drafts.csswg.org/css-fonts/#propdef-font

- [ ] Follow up on DEV-11198.

Jym77:
Resolves #682


Jym77:
Arf, need more heavy handed work to handle `font-style`… (or maybe just ignore the extra angle due to poor browser support).

kasperisager (dismissed):

packages/alfa-css/src/value/keyword.ts
```diff
@@ -72,7 +74,7 @@ export namespace Keyword {
         // `Keyword<"foo"> | Keyword<"bar">`, not `Keyword<"foo" | "bar">`. Why?
         // Because the former is assignable to the latter, but the latter isn't
         // assignable to the former.
-        Keyword.of(ident.value.toLowerCase()) as { [K in T]: Keyword<K> }[T]
+        Keyword.of(ident.value.toLowerCase()) as ToKeyword<T>
```
kasperisager, line 77: I'm not too keen on this as the difference between `Keyword<"foo" | "bar">` and `Keyword.ToKeyword<"foo" | "bar">` is too small, especially when broken over several lines. The former type is incorrect and the latter will result in a correct type being produced, but all it takes is forgetting `.ToKeyword` for the type to end up wrong. It's much easier to tell `Keyword<"foo"> | Keyword<"bar">` from `Keyword<"foo" | "bar">`, even if it does cost some additional keystrokes. Correctness over conciseness though 😄

kasperisager (changes requested):

packages/alfa-css/src/value/percentage.ts
```diff
@@ -33,7 +33,7 @@ export class Percentage extends Numeric<"percentage"> {
   }
 
   public toString(): string {
-    return `${this._value}%`;
+    return `${this._value * 100}%`;
```
kasperisager, line 36: 👏

packages/alfa-css/src/value/keyword.ts
```diff
@@ -61,7 +61,7 @@ export namespace Keyword {
     return value instanceof Keyword;
   }
 
-  export function parse<T extends string>(...keywords: Array<T>) {
+  export function parse<T extends string>(...keywords: ReadonlyArray<T>) {
```
kasperisager, line 64:
```suggestion
  export function parse<T extends string>(...keywords: Array<T>) {
```

The arrays created for argument spreads are internal to the function and so there's no risk of the caller passing a mutable array that is also mutable by the function:

```js
const f = (...a) => a;
const a = [1, 2, 3];

f(...a) !== a;
f.apply(this, a) !== a;
```

kasperisager (dismissed):

kasperisager (approved):
Looks great! 🎉

PR #696, opened 2021-02-08
Add @siteimprove/alfa-network package

kasperisager:
Needed for some internal Siteimprove shenanigans.


packages/alfa-network/src/network.ts
```diff
@@ -0,0 +1,317 @@
...
+            )
+          )
+          .get()
+      )
+    );
+  }
+
+  public disconnect(from: N, to: N): Network<N, E>;
+
+  public disconnect(from: N, to: N, edge: E): Network<N, E>;
```
Jym77, line 100: This could also take a set of edge to disconnect several but not all in one go.

packages/alfa-network/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "$schema": "http://json.schemastore.org/package",
+  "name": "@siteimprove/alfa-network",
+  "homepage": "https://siteimprove.com",
+  "version": "0.10.0",
+  "license": "MIT",
+  "description": "An implementation of an immutable, directed graph that allows for multiple, unique edges",
```
Jym77, line 7: Isn't that a (directed) multigraph?
(which is apparently also called a quiver 🤷 )
https://en.wikipedia.org/wiki/Multigraph#Directed_multigraph_(edges_with_own_identity)
https://en.wikipedia.org/wiki/Quiver_(mathematics)

packages/alfa-network/test/network.spec.ts
```diff
@@ -0,0 +1,241 @@
...
+test("#disconnect() disconnects two nodes in a network", (t) => {
+  // foo
+  // |- 2 => baz
+  //    |- 3 => foo
+  //       |- ...
+  // bar
+  t.deepEqual(network.disconnect("foo", "bar").toArray(), [
+    ["baz", [["foo", [3]]]],
+    ["foo", [["baz", [2]]]],
+    ["bar", []],
```
Jym77, line 78: It's a bit weird that `.from` does not allow to create disjoint graphs, but `.disconnect` is. But I guess it wouldn't be clear which bit to drop.

Jym77 (approved):

packages/alfa-network/package.json
kasperisager: That's exactly what it is, yes 👍 I named it "network" to align with Google Guava (https://guava.dev/releases/snapshot/api/docs/com/google/common/graph/Network.html). Didn't know about the name "quiver" though, that's neat 😄

packages/alfa-network/test/network.spec.ts
kasperisager: `Network.from()` can do that as well:

```ts
Network.from([
  ["foo", []],
  ["bar", []],
]);
```

packages/alfa-network/src/network.ts
kasperisager: Good idea! 👍

Jym77 (approved):

PR #702, opened 2021-02-10
Add @siteimprove/alfa-selective package

### LINKED ISSUE #410: Investigate structured conditional computations

kasperisager:
Resolves #410.

### DIFF
--- a/packages/alfa-selective/src/selective.ts
+++ b/packages/alfa-selective/src/selective.ts
@@ -1,0 +1,103 @@
+import { Either, Left, Right } from "@siteimprove/alfa-either";
+import { Equatable } from "@siteimprove/alfa-equatable";
+import { Functor } from "@siteimprove/alfa-functor";
+import { Hash, Hashable } from "@siteimprove/alfa-hash";
+import { Serializable } from "@siteimprove/alfa-json";
+import { Mapper } from "@siteimprove/alfa-mapper";
+import { Predicate } from "@siteimprove/alfa-predicate";
+import { Refinement } from "@siteimprove/alfa-refinement";
+
+export class Selective<S, T = never>
+  implements
+    Functor<T>,
+    Iterable<S | T>,
+    Equatable,
+    Hashable,
+    Serializable<Selective.JSON<S, T>> {
+  public static of<T>(value: T): Selective<T> {
+    return new Selective(Left.of(value));
+  }
+
+  private readonly _value: Either<S, T>;
+
+  private constructor(value: Either<S, T>) {
+    this._value = value;
+  }
+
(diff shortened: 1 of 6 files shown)


packages/alfa-selective/src/selective.ts
```diff
@@ -0,0 +1,103 @@
...
+  private readonly _value: Either<S, T>;
+
+  private constructor(value: Either<S, T>) {
+    this._value = value;
+  }
+
+  public map<U>(mapper: Mapper<T, U>): Selective<S, U> {
+    return new Selective(
+      this._value.either(
+        (value) => Left.of(value) as Either<S, U>,
```
Jym77, line 30: I'm wondering if `Left` (and `Left.of`, `Right`, …) should take a second type parameter (defaulting to `never`) so that this could be `Left.of<S, U>(value)` and remove the need for type coercion which is always a dent in type safety…

Jym77 (approved):

packages/alfa-selective/src/selective.ts
kasperisager: `Left` has no knowledge of the right branch of an `Either` so even if we added an additional type parameter to `Left.of`, we'd have nothing to use it for.

This really isn't a coercion though; it's an assertion and TypeScript actually performs assignability checks when encountering assertions to ensure that there's a reasonable overlap between the inferred type and the asserted type. In this case, it's therefore just a way of overruling the inference of TypeScript.

PR #707, opened 2021-02-11
Rework media query

Jym77:


kasperisager:
@Jym77 I'll port some of the things you've implemented here and then put together a draft design for range values. This has become a little too involved without yet progressing on solving range values so we need that sorted first and foremost.

kasperisager:
@Jym77 A draft design is now available here: https://github.com/Siteimprove/alfa/tree/media-range-values. The core additions are the `Value<T>` interface and the `Value.Discrete<T>` and `Valute.Range<T>` classes which implement this interface. `Value.Discrete<T>` models discrete values and matches based on equality. `Value.Range<T>` models range values and matches based on ordering. To make dimensional CSS values work in context of `Value.Range<T>`, `Dimension` now implements `Comparable<T>` to allow comparing dimensions. To make things work across units, `Dimension#canonicalUnit` must now be implemented by subclasses such that dimensions can be converted to their canonical unit before comparing them.

kasperisager:
Closing in favour of #722.

PR #722, opened 2021-02-24
Rework media query and handle media ranges

### LINKED ISSUE #109: Implement remaining media query parser productions
Specifically, `<mf-range>` is currently missing: https://www.w3.org/TR/mediaqueries/#typedef-mf-range

Jym77:
Resolves #109.
Supercedes #707.


kasperisager:
I dialled down the complexity of this a tad. First and foremost, I'll take a single ~1.1 KLoC module any day over the following ~1.1 KLoC module graph that we ended up with:

module graph

That's simply not a helpful separation of concerns, also because the individual modules then started exposing internal functionality that was needed by the other modules. Previously, all these internals were simply hidden inside the same namespace.

Some of the abstractions that were introduced also made it pretty difficult to map the code to specification and the only real purpose these indirections seemed to serve was to avoid some additional keystrokes for things that would otherwise have to be copy/pasted and then shuffled around a bit. I did precisely that and ended up with a total of 37 _fewer_ LoC and no confusing indirections.

As I've said previously, DRY is a worthy pursuit up to a point. Often times, a little copy/paste is all you need though, and abstractions need to serve a different purpose than just cutting down on keystrokes. If an abstraction helps with comprehension then, by all means, abstract away! If it doesn't then additional keystrokes really aren't an evil to avoided at all costs.

Things look great otherwise so if we're happy with the test coverage let's land this 👍

packages/alfa-media/src/media.ts
```diff
@@ -232,20 +202,213 @@ export namespace Media {
...
+    ): Result<Feature, string>;
+
+    export function tryFrom(
+      name: string,
+      value?: Value
+    ): Result<Feature, string> {
+      if (value) {
+        switch (name) {
+          case "width":
+            if (value.hasValue(Length.isLength)) {
```
Jym77, line 219: I'd rather keep the validations inside the classes themselves for better locality.
Essentially, it's `Width` who needs to know what it accepts as values. `tryFrom` shouldn't need to know what each of the individual feature accepts but simply ask them to validate the value.

packages/alfa-media/src/media.ts
```diff
@@ -312,15 +675,238 @@ export namespace Media {
...
+    }
+
+    export const { of: bound } = Bound;
+  }
+
   /**
    * @see https://drafts.csswg.org/mediaqueries/#media-condition
    */
-  export class Condition implements Equatable, Serializable {
+  export class Condition
```
Jym77, line 908: I'm not too keen on calling this class `Condition` given that it does not corresponds to what CSS calls "media condition" nor to how we use it.
The link just before the class actually points to a list whose first item is `Negation` (hence, a CSS media condition, but not a `Condition`), so not really explaining what the class does.
Similarly, `Negation` has a `condition` field which is not a `Condition` but a `Condition | Feature | Negation` hinting that the name is not good.

packages/alfa-media/src/media.ts
```diff
@@ -523,18 +1129,17 @@ export namespace Media {
...
   );
 
   /**
    * @see https://drafts.csswg.org/mediaqueries/#media-query
    */
-  export class Query implements Equatable, Serializable {
+  export class Query implements Queryable {
     public static of(
       modifier: Option<Modifier>,
       type: Option<Type>,
```
Jym77, line 1145: I'm really reluctant in keeping this as `type` because we do have a fairly consistent use of `type` as a discriminator field almost everywhere.
Especially, since unknown media types are accepted, it would mean that for example a media query `"length"`  would result in an object whose `type` is `length` but is absolutely not a `Length` 🙈 

This is somewhat tempered by typing itself (since stuff that are `Query | Length` are fairly rare), but this could happen with any other object and feels way too brittle for my confort zone.

kasperisager: I don't agree. The term "type" has a very particular meaning for this structure and since we have no need to discriminate it as part of a union, "type" does seem like the best fit. The fact that the term "type" is always used as a discriminator when we deal with union types doesn't really seem relevant here as we're not dealing with a union. Additionally, the `Type` structure doesn't serialise to a `string` so there's never any doubt the the `Query#type` property _isn't_ a discriminator because you'd see this:

```js
{
  type: {
    name: "length"
  }
}
```

packages/alfa-media/src/media.ts
kasperisager: A "media condition" is any combination of features, negations, and other conditions. Yes, `Negation` should likely fall under "media condition", but the point is that a "media condition" is a _combination_ of multiple "media features", as per the prose in the specification.

I'd be happy to use the term "media expression" to cover over "media condition, feature, and negation" though, which would also mean that the `Negation#condition` would become `Negation#expression` but "media condition" should remain the combination of _multiple_ "media expressions".

packages/alfa-media/src/media.ts
kasperisager: That's fine by me as well; I'll branch out `Feature.tryFrom()` into the different specific features. That's just good ol' divide and conquer 😄

packages/alfa-media/src/media.ts
kasperisager: I think I found a better approach that aligns more closely with the specification prose: `Condition` is now `And | Or | Not`, which matches the description of conditions being features combined with boolean algebra.

kasperisager (approved):

PR #725, opened 2021-02-26
Accept spaces around CSS variable names in `var()` functions

Jym77:


kasperisager (changes requested):
We just need a change log and then this is good to go 👍

kasperisager (approved):

PR #729, opened 2021-03-01
Unbranch ARIA types

### LINKED ISSUE #153: Implement browsers discrepancies in handling `alt` attribute
Chrome is keeping `` in the accessibility tree with a role of `img`.

Firefox is putting all ``, including ``, in the accessibility tree with a role of `graphic`.

Edge is (correctly) removing `` from the accessibility tree.

### LINKED ISSUE #154: Implement browsers discrepancies in computing accessible name
Browser computation of accessible names has several weirdness.

* Chrome is trimming out spaces, but keeping non-breaking spaces.
* Firefox also keeps non-breaking spaces when they are surrounded by regular spaces.
* Firefox handles name from content, `aria-label` and `alt` attributes on images in three different ways and do not perform the same trimming on each.

See https://github.com/act-rules/act-rules.github.io/issues/976 for more details.

kasperisager:
While looking into CPU profiles over the past couple of months, one thing that consistently came out near the top of our cycle spend was handling of branched values during accessibility tree construction. The only thing we currently account for, however, is a bug in Firefox related to case sensitive `role` attributes. 

The plan has for a long time been to expand our coverage of these browser differences, but for lack of actually presenting them in the Siteimprove platform this work has kept getting pushed. As the release of the new Accessibility product has also shown, many customers are already finding it difficult enough to understand some of the more advanced rules that are now available. Adding an additional layer of complexity that is results dependant on possibly branched values will not be an improvement on this front.

For the reasons outlined above, I've therefore decided to pull out what current handling we have in @siteimprove/alfa-aria of branched values, which here and now serves two purposes: Improve performance and simplify the code.

That of course leaves a bit of an open question: What to do about the stuff that branched values were supposed to solve, which is browser differences that could lead to accessibility issues being present in one browser, but not another? My immediate thought is to encode these as separate rules, which will hopefully make it _much_ easier for end users to understand what the issue is. Consider, for example, the following two statements:

> The accessible name of the element is empty in "Browser XYZ".

vs.

> The accessible name of the element is based on its `title` attribute and might therefore not be correctly computed in "Browser XYZ".

The former is what we'd be able to determine with branched values. The latter is what we'd be able to determine with a rule that specifically calls out cases where an accessible name has a source that is based on the `title` attribute, thanks to us already tracking the sources of accessible names.

Resolves #153.
Resolves #154.

### DIFF
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -1,6 +1,4 @@
-import { Branched } from "@siteimprove/alfa-branched";
 import { Cache } from "@siteimprove/alfa-cache";
-import { Browser } from "@siteimprove/alfa-compatibility";
 import { Device } from "@siteimprove/alfa-device";
 import { Node, Element, Namespace } from "@siteimprove/alfa-dom";
 import { Iterable } from "@siteimprove/alfa-iterable";
@@ -27,7 +25,7 @@ export class Feature {
   public static of(
     role: Feature.Aspect.Role = () => [],
     attributes: Feature.Aspect.Attributes = () => [],
-    name: Feature.Aspect.Name = () => Branched.of(None)
+    name: Feature.Aspect.Name = () => None
   ): Feature {
     return new Feature(role, attributes, name);
   }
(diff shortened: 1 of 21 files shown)


Jym77 (approved):
Should we gather knowns discrepancies somewhere, especially if we plan to write rules for them later? (so, maybe in Sanshikan)

kasperisager:
Yes, let's put those as rule ideas in Sanshikan.

Issue #731, opened 2021-03-02
`isClippedBySize()` does not account for cases where the grandparent of a text node fully indents and clips text


Jym77:
`isClippedBySize()` currently returns `false` for the following text node:

```html
<style>
  div {
    overflow-x: hidden;
    text-indent: 100%;
    white-space: nowrap;
  }
</style>

<div>
  <span>This is invisible</span>
</div>
```

- [ ] Follow up on DEV-11338.

kasperisager:
Adjusted the title and description to clarify the _issue_ at hand, not the proposed solution. The issue with simply looking further up the tree for a computed `overflow-x` is that, for some reason, many pages declare `overflow-x: hidden` near the document root (`<body>` or `<html>`).

For lack of a layout system, I'd rather we assume that `text-indent: -100%` or `text-indent: 100%; white-space: nowrap` means that the text will be invisible, regardless of `overflow-x`. There might be other pitfalls to that so let's examine those a little further at a refinement.

PR #734, opened 2021-03-03
Limit SIA-R83 to non-inline, non-font relative heights

### LINKED ISSUE #546: SIA-R83 should not fail containers that leave enough space for more than 1 line of text without clipping
SIA-R83 currently passes containers that hide overflowing text, but leave exactly enough room for 1 line of text. That is, if the line height is equal to the height of the container, the rule passes. However, it fails containers that leave exactly enough room for _more than_ 1 line of text, such as by setting the height of the container to a multiple of the line height:

```html
<style>
  div {
    overflow: hidden;
    height: calc(3 * 1.5em);
    line-height: 1.5em;
  }
</style>

<div>
  Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi a dapibus
  dolor. In hac habitasse platea dictumst. Maecenas ornare efficitur finibus. In
  eget nisi sed nisi vehicula lacinia quis vel tellus. Nullam tempus consectetur
  imperdiet. Nunc a nisl a elit consectetur eleifend. Praesent ut nibh
  scelerisque, dictum ipsum vel, ultricies arcu. Phasellus sed mauris at augue
  ullamcorper venenatis volutpat ac libero. Vivamus quis finibus libero.
</div>

```

- [ ] Follow up on CE-4948.

### LINKED ISSUE #622: SIA-R83 should not fail when an innermost ancestor handles the overflow
SIA R83 fails on https://act-rules.github.io/rules/59br37#passed-example-4 :
```html
<div style="overflow: hidden; height: 2em; font-size: 16px;">
	<div style="overflow: auto; height: 2em;">
		Once upon a midnight dreary, while I pondered, weak and weary, Over many a quaint and curious volume of forgotten
		lore. While I nodded, nearly napping, suddenly there came a tapping.
	</div>
</div>
```
This passes because the innermost `<div>` handles the overflow and thus does **not** itself overflows, and the outermost `<div>` thus has nothing to clip.

Alfa fails this because `wrapsText` doesn't look at the innermost `<div>` at all and only see the outermost one.

Slightly tricky/verbose fix since each dimension has to be handled individually.

### LINKED ISSUE #649: SIA-R83 should accept `min-height` as a way to avoid clipped text
SIA-R83 currently fails a case like this:

```html
<div style="overflow: hidden; height: 10px; line-height: 1.5em; min-height: 1.5em">
  Hello world
</div>
```

- [ ] Follow up on DEV-11016.

### LINKED ISSUE #712: SIA-R83 should not flag possible text clipping based on a `height` set via `style`
I'm seeing a increasing number of false positives for SIA-R83 related to code like the following:

```html
<style>
  .card {
    overflow: hidden;
  }
</style>

<div class="card-list">
  <div class="card" style="height: 20px">
    First card
  </div>
  <div class="card" style="height: 20px">
    Second card
  </div>
</div>
```

On its own, that code is indeed problematic as the absolute height does not scale with an increased root font size and the text will therefore clip when resized to 200%. What most often happens in these cases, however, is that JavaScript is used to adjust the `height` so that the text fits within the cards and ensure that all cards have the same height. While a pattern like this is possible, and arguably preferred, to implement in plain CSS with either grid or flexbox layout, it's often implemented like this for sites where compatibility is a concern.

- [ ] Follow up on DEV-11310.

kasperisager:
Resolves #546.
Resolves #622.
Resolves #649.
Resolves #712.

### DIFF
--- a/packages/alfa-css/src/value/unit.ts
+++ b/packages/alfa-css/src/value/unit.ts
@@ -30,12 +30,12 @@ export namespace Unit {
       export type Viewport = "vw" | "vh" | "vmin" | "vmax";
     }
 
-    export function isRelative(input: string): input is Relative {
-      return isFontRelative(input) || isViewportRelative(input);
+    export function isRelative(unit: string): unit is Relative {
+      return isFontRelative(unit) || isViewportRelative(unit);
     }
 
-    export function isFontRelative(input: string): input is Relative.Font {
-      switch (input) {
+    export function isFontRelative(unit: string): unit is Relative.Font {
+      switch (unit) {
         case "em":
         case "ex":
         case "ch":
@@ -47,9 +47,9 @@ export namespace Unit {
     }
 
     export function isViewportRelative(
-      input: string
-    ): input is Relative.Viewport {
-      switch (input) {
+      unit: string
+    ): unit is Relative.Viewport {
+      switch (unit) {
         case "vw":
         case "vh":
         case "vmin":
@@ -65,8 +65,8 @@ export namespace Unit {
      */
     export type Absolute = "cm" | "mm" | "Q" | "in" | "pc" | "pt" | "px";
 
-    export function isAbsolute(input: string): input is Absolute {
-      switch (input) {
+    export function isAbsolute(unit: string): unit is Absolute {
+      switch (unit) {
         case "cm":
         case "mm":
         case "Q":
@@ -93,8 +93,8 @@ export namespace Unit {
    */
   export type Length = Length.Relative | Length.Absolute;
 
-  export function isLength(input: string): input is Length {
-    return isRelativeLength(input) || isAbsoluteLength(input);
+  export function isLength(unit: string): unit is Length {
+    return isRelativeLength(unit) || isAbsoluteLength(unit);
   }
 
   /**
@@ -102,8 +102,8 @@ export namespace Unit {
    */
   export type Angle = "deg" | "grad" | "rad" | "turn";
 
-  export function isAngle(input: string): input is Angle {
-    switch (input) {
+  export function isAngle(unit: string): unit is Angle {
+    switch (unit) {
       case "deg":
       case "grad":
       case "rad":
@@ -119,8 +119,8 @@ export namespace Unit {
    */
   export type Time = "s" | "ms";
 
-  export function isTime(input: string): input is Time {
-    switch (input) {
+  export function isTime(unit: string): unit is Time {
+    switch (unit) {
       case "s":
       case "ms":
         return true;
@@ -134,8 +134,8 @@ export namespace Unit {
    */
   export type Frequency = "hz" | "kHz";
 
-  export function isFrequency(input: string): input is Frequency {
-    switch (input) {
+  export function isFrequency(unit: string): unit is Frequency {
+    switch (unit) {
       case "hz":
       case "kHz":
         return true;
@@ -149,8 +149,8 @@ export namespace Unit {
    */
   export type Resolution = "dpi" | "dpcm" | "dppx";
 
-  export function isResolution(input: string): input is Resolution {
-    switch (input) {
+  export function isResolution(unit: string): unit is Resolution {
+    switch (unit) {
       case "dpi":
       case "dpcm":
       case "dppx":
--- a/packages/alfa-css/src/value/unit.ts
+++ b/packages/alfa-css/src/value/unit.ts
@@ -16,22 +16,40 @@ export namespace Unit {
     /**
      * {@link https://drafts.csswg.org/css-values/#relative-lengths}
      */
-    export type Relative =
-      | "em"
-      | "ex"
-      | "ch"
-      | "rem"
-      | "vw"
-      | "vh"
-      | "vmin"
-      | "vmax";
+    export type Relative = Relative.Font | Relative.Viewport;
+
+    export namespace Relative {
+      /**
+       * {@link https://drafts.csswg.org/css-values/#font-relative-lengths}
+       */
+      export type Font = "em" | "ex" | "ch" | "rem";
+
+      /**
+       * {@link https://drafts.csswg.org/css-values/#viewport-relative-lengths}
+       */
+      export type Viewport = "vw" | "vh" | "vmin" | "vmax";
+    }
 
     export function isRelative(input: string): input is Relative {
+      return isFontRelative(input) || isViewportRelative(input);
+    }
+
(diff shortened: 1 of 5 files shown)


packages/alfa-css/src/value/unit.ts
```diff
@@ -16,22 +16,40 @@ export namespace Unit {
...
+      export type Font = "em" | "ex" | "ch" | "rem";
+
+      /**
+       * {@link https://drafts.csswg.org/css-values/#viewport-relative-lengths}
+       */
+      export type Viewport = "vw" | "vh" | "vmin" | "vmax";
+    }
 
     export function isRelative(input: string): input is Relative {
+      return isFontRelative(input) || isViewportRelative(input);
```
Jym77, line 34: I think in this (and the others), `unit` would be a more descriptive variable name than `input`.

packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -93,32 +92,41 @@ function isPossiblyClipped(device: Device): Predicate<Element> {
...
     }
 
-    // Case 2: The height of the element has been restricted. In this case, text
-    // might clip if overflow of the y-axis is hidden.
+    // Case 2: The height of the element has been restricted using an non-font
+    // relative length not set via the `style` attribute. In this case,
+    // text might clip if overflow of the y-axis is hidden.
+    //
+    // For heights set via the `style` attribute, we assume that its value is
+    // controlled by JavaScript and is adjusted as the content scales.
```
Jym77, line 109: As far as I understand, font relative height are assumed to leave enough room for several lines (#546) 🤔 
That should also be recorded here (especially since it is the bit requiring to use the cascaded value rather than the computed one).

packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -93,32 +92,41 @@ function isPossiblyClipped(device: Device): Predicate<Element> {
         .some((whiteSpace) => whiteSpace.value === "nowrap")
     ) {
       return style
-        .computed("overflow-x")
-        .some(
-          (overflow) => overflow.value === "hidden" || overflow.value === "clip"
+        .cascaded("overflow-x")
```
Jym77, line 95: We should explain that why we need cascaded value rather than computed, since it is a bit unusual.

packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -93,32 +92,41 @@ function isPossiblyClipped(device: Device): Predicate<Element> {
...
+              !height.isFontRelative() &&
+              source.some((declaration) => declaration.parent.isSome())
+          )
+        )
     ) {
       return style
-        .computed("overflow-y")
-        .some(
-          (overflow) => overflow.value === "hidden" || overflow.value === "clip"
+        .cascaded("overflow-y")
```
Jym77, line 124: Here also, we should explain why we need the cascaded value.

packages/alfa-rules/test/sia-r83/rule.spec.tsx
```diff
@@ -125,70 +130,74 @@ test(`evaluate() fails a text node that clips overflow by not wrapping text
...
-    </div>,
-  ]);
+test(`evaluate() is inapplicable to a text node with a fixed relative height`, async (t) => {
+  const document = Document.of(
+    [<div>Hello world</div>],
+    [
+      h.sheet([
+        h.rule.style("div", {
+          overflow: "hidden",
+          height: "1.2em",
```
Jym77, line 196: This is actually clipping the second line of text.
(Well, sort of, because there is top and bottom white space in the `<div>`, but depending on the height, this is clipped when there are several lines of text).
`height: 1.5em`:
clip

packages/alfa-rules/test/sia-r83/rule.spec.tsx
```diff
@@ -125,70 +130,74 @@ test(`evaluate() fails a text node that clips overflow by not wrapping text
...
+        }),
+      ]),
+    ]
+  );
```
Jym77, line 201: We should add an example for overflow clipped by an ancestor.

Jym77 (changes requested):

packages/alfa-rules/src/sia-r83/rule.ts
kasperisager: `overflow-x` doesn't inherit and computes as specified so this is just a micro optimisation. That's also worth pointing out though.

packages/alfa-rules/test/sia-r83/rule.spec.tsx
kasperisager: Yes and that case is immediately obvious without scaling the text as that will _always_ cause clipping. Without a layout system, however, we can't accurately call out these cases without a high risk of false positives. The rule will therefore only call out the less obvious cases where content might look alright at normal scale, but will likely cause text clipping at increased scales.

packages/alfa-rules/test/sia-r83/rule.spec.tsx
kasperisager: The rule doesn't support those cases and can't reliably do so without a layout system. As mentioned, there has been several cases where fixed heights have been used in combination with `overflow-{x,y}: hidden` on an ancestor, sometimes near the root, and I've yet to see a case where this actually lead to clipped text. There has always some other thing going on in the layout that ensured that the text wouldn't clip. The rule therefore only deals with `overflow` and `height` properties specified for the same element.

packages/alfa-rules/src/sia-r83/rule.ts
kasperisager: I forgot that computed values are pulled directly from cache just as cascaded values are pulled directly from the property map so we might as well grab the computed value and save two lines of code 😄

packages/alfa-rules/test/sia-r83/rule.spec.tsx
Jym77: My bad 🙈 I thought this was also solving #731 (actually confused #622 with it when reading the names…)

packages/alfa-rules/test/sia-r83/rule.spec.tsx
Jym77: It could be displayed on a single (unclipped) line when not zoomed and need another (clipped) line when zoomed .

I agree that it is too involved without a layout system.

Jym77 (dismissed):

PR #737, opened 2021-03-05
Add viewport fitting and animation awaiting to scraper and CLI

kasperisager:
This pull request introduces two new features:

1. A `fit: boolean` option in `Scraper#scrape()` and a corresponding `--[no-]fit` flag in the `alfa scrape` command. When set, which it is by default, the viewport will expand to fit the contents of the page. This is needed to accurately scrape pages that toggle content depending on scroll position.

2. An `Awaiter.animations()` function and a corresponding `--await-animations` flag in the `alfa scrape` command. This awaiter polls the document for active animations and resolves when no animations are active. It does not yet deal with infinitely repeating animations, however, and so this is left as a future exercise.

The changes are breaking as the `Awaiter` type has been adjusted. The previously available awaited value, `T`, was never used and made it difficult to compose awaiters. This has therefore been removed and instead of `Result<T, string>` awaiters are now instead expected to return `Option<string>` indicating whether or not an error occurred.


kasperisager:
I've still got some change logging to do but this should otherwise be ready for review.

Jym77 (approved):

PR #739, opened 2021-03-08
Correct handling of `opacity` when computing foreground color of elements

### LINKED ISSUE #726: SIA-R69 does not correctly composite background and foreground colors when encountering `opacity`
Consider the following case:

```html
<style>
  div {
    background-color: black;
  }

  p {
    background-color: red;
    color: white;
    opacity: 0.5;
  }
</style>

<div>
  <p>Hello world</p>
</div>
```

SIA-R69 currently assumes that for the above case it's sufficient to apply a 50% opacity to the `color` and `background-color` properties of the `p` declaration, effectively reducing it to the following:

```html
<style>
  div {
    background-color: black;
  }

  p {
    background-color: rgba(255, 0, 0, 0.5);
    color: rgba(255, 255, 255, 0.5);
  }
</style>

<div>
  <p>Hello world</p>
</div>
```

That's incorrect, however. The `opacity` property specifies the degree to which the element _as a whole_ should be blended with its background. So, when `opacity` is set, all colors of the _current_ layer should be blended with all colors of the _previous_ layer.

- [x] Follow up on DEV-11324.

Jym77:
Resolves #726 

Main changes:
* The `opacity` is not handled any more in `resolveColor` or `getLayers`.
* The `opacity` is now handled in `getBackground` when flattening the layers together.
* `getLayers` thus return a list of `{colors, opacity}`, keeping the opacity of each layer for `getBackground` to flatten.
* `composite` now requires an `opacity` argument to handle that bit.

* In `getForeground`, when the color is not opaque, it is first merged with its own background according to its alpha channel, and then merged with the **parent** background according to the `opacity`.
* When merging with its own background, `opacity` of the element must be disregarded, but not the alpha channel of the text nor the alpha channel of the background, nor the alpha channel **and** `opacity` of each of the **previous** layers.
* For this, `getBackground` (and `getLayers`) take an extra fake opacity parameter that can override the opacity **of the last layer** only.

Results seem OK. I've checked the tests by comparing the actual and expected elements in Chrome next to each other and not seeing any visual difference… Not ideal, but seems OK…
Still feels a bit hacky and brittle… 😕


packages/alfa-rules/test/common/expectation/get-colors.spec.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+  t.deepEqual(Iterable.first(getForeground(target).get()).get().toJSON(), {
+    type: "color",
+    format: "rgb",
+    red: { type: "percentage", value: 0.75 },
+    green: { type: "percentage", value: 0.25 },
+    blue: { type: "percentage", value: 0.5 },
+    alpha: { type: "percentage", value: 1 },
+  });
+});
```
kasperisager, line 114: Rather than `Document.of()`, the tests should use `h.document()`. The unused variables can also be removed and `h.document()` simply left as a side effect.

kasperisager (dismissed):
Looks good! Just a small editoral for the tests.

kasperisager (approved):

PR #740, opened 2021-03-10
Add Chai assertions for Cypress integration

kasperisager:
While working through the examples in https://github.com/Siteimprove/alfa-examples, it struck me that I really don't like the way we currently integrate with Cypress. It was originally necessitated by the odd command model used by Cypress which meant that we couldn't do things like writing to files in the middle of an assertion, which is how our assertion handlers usually operate. A custom command was therefore introduced rather than the usual assertions, but this feels clunky and is not terribly flexible. Among other things, one can't assert that nodes _aren't_ accessible.

This pull request fixes that by effectively splitting the persistence handler in two: The first part registers an `after()` hook that will write files after the test run and the second part registers files to be written during assertions. The end result is an integration that aligns with the rest of the current set of integrations both in terms of look, feel, and flexibility.

The changes are breaking as `Cypress.createPlugin()` now returns a Chai plugin rather than registering a custom Cypress command. Also, where one would previously write something like this...

```ts
it("should be accessible", () => {
  cy.visit("/fixtures/page.html");
  cy.document().audit();
});

```

...one now needs to write this:

```ts
it("should be accessible", () => {
  cy.visit("/fixtures/page.html");
  cy.document().should("be.accessible");
});
```

Typings are also provided for the normal `expect()` API:

```ts
it("should be accessible", () => {
  cy.visit("/fixtures/page.html");
  cy.document().should(async (document) => {
    await expect(document).to.be.accessible();
  });
});
```


kasperisager:
@scttjhnsn This might be of interest to you, if only as a heads up that I'm expecting this to land in v0.15.0.

packages/alfa-cypress/src/cypress.ts
```diff
@@ -52,21 +63,31 @@ export namespace Cypress {
...
 
-    const command = (value: Type) =>
-      cy.wrap(value, { log: false }).should(() => {
-        const input = toPage(value);
+    return (chai) => {
+      chai.Assertion.addMethod("accessible", async function () {
+        const input = toPage(this._obj);
 
-        const error = asserter.expect(input).to.be.accessible().get();
+        const error = await asserter.expect(input).to.be.accessible();
```
Jym77, line 73: Shouldn't that rather be called `result`?
All the `error.isOk` tests are confusing to read 😕

Jym77 (approved):

packages/alfa-cypress/src/cypress.ts
kasperisager: Good point!

Jym77 (approved):

PR #746, opened 2021-03-15
SIA-R84 should only consider elements scrollable along the x-axis if they specify `white-space: nowrap`

kasperisager:


packages/alfa-rules/src/sia-r84/rule.ts
```diff
@@ -78,22 +78,31 @@ export namespace Outcomes {
...
+              // For the x-axis, the content is likely only scrollable if
+              // `white-space: nowrap` is used. Otherwise, the content is likely
+              // to instead break along the y-axis.
+              if (axis === "x") {
+                return style
+                  .computed("white-space")
+                  .some((whitespace) => whitespace.value === "nowrap");
+              }
+
               return true;
```
Jym77, line 106: Always feels wonky to return booleans inside a test, but then
```suggestion
              return axis === "y") ||
                style
                  .computed("white-space")
                  .some((whitespace) => whitespace.value === "nowrap");
```
does obfuscate a bit what we test… 🤔

Jym77 (approved):

packages/alfa-rules/src/sia-r84/rule.ts
kasperisager: Aye, I do prefer the explicit branch for this.

PR #760, opened 2021-04-07
Rework SIA-R62 as a fully automated rule

### LINKED ISSUE #668: Update SIA-R62 according to latest ACT-R draft
The ACT-R rule that SIA-R62 builds on (https://github.com/act-rules/act-rules.github.io/pull/1010) has changed quite substantially since we first implemented it. In fact, it should now be possible to provide a fully automated implementation of this rule. While the rule hasn't been published yet, we should consider updating our implementation according to the latest draft as soon as time allows.

#710 and #709 are needed for this, though we might be able to do with just of subset of those properties.

- [ ] Follow up on DEV-11431.

kasperisager:
Resolves #668.

### DIFF
--- a/packages/alfa-rules/src/common/predicate/has-border.ts
+++ b/packages/alfa-rules/src/common/predicate/has-border.ts
@@ -1,0 +1,28 @@
+import { Color, Length } from "@siteimprove/alfa-css";
+import { Device } from "@siteimprove/alfa-device";
+import { Element } from "@siteimprove/alfa-dom";
+import { Predicate } from "@siteimprove/alfa-predicate";
+import { Context } from "@siteimprove/alfa-selector";
+import { Style } from "@siteimprove/alfa-style";
+
+const sides = ["top", "right", "bottom", "left"] as const;
+
+export function hasBorder(
+  device: Device,
+  context?: Context
+): Predicate<Element> {
+  return (element) => {
+    const style = Style.from(element, device, context);
+
+    return sides.some(
+      (side) =>
+        style.computed(`border-${side}-width` as const).none(Length.isZero) &&
+        style
+          .computed(`border-${side}-style` as const)
+          .none((style) => style.value === "none") &&
+        style
+          .computed(`border-${side}-color` as const)
+          .none((color) => color.type === "color" && Color.isTransparent(color))
+    );
+  };
+}
--- a/packages/alfa-rules/src/sia-r62/rule.ts
+++ b/packages/alfa-rules/src/sia-r62/rule.ts
@@ -1,15 +1,18 @@
 import { Rule, Diagnostic } from "@siteimprove/alfa-act";
+import { Color } from "@siteimprove/alfa-css";
 import { Device } from "@siteimprove/alfa-device";
 import { Element, Node, Text } from "@siteimprove/alfa-dom";
 import { Predicate } from "@siteimprove/alfa-predicate";
 import { Err, Ok } from "@siteimprove/alfa-result";
 import { Context } from "@siteimprove/alfa-selector";
+import { Style } from "@siteimprove/alfa-style";
 import { Criterion } from "@siteimprove/alfa-wcag";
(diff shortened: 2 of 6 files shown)


kasperisager:
I decided against implementing support for `box-shadow` for now as the same argument as for `background-image` applies; box shadows can layer and it's not yet clear how to best handle such cases. Plus, I'd be surprised to find `box-shadow` used as the _only_ distinguishing feature, but of course it's likely used as such _somewhere_. Let's revisit when concrete cases pop up.

packages/alfa-rules/src/common/predicate/has-border.ts
```diff
@@ -0,0 +1,28 @@
+import { Color, Length } from "@siteimprove/alfa-css";
+import { Device } from "@siteimprove/alfa-device";
+import { Element } from "@siteimprove/alfa-dom";
+import { Predicate } from "@siteimprove/alfa-predicate";
+import { Context } from "@siteimprove/alfa-selector";
+import { Style } from "@siteimprove/alfa-style";
+
+const sides = ["top", "right", "bottom", "left"] as const;
```
Jym77, line 8: I think we should also check the logical `border-[block,inline]-[start,end]` in the same way since this won't cost much more and could be used.

packages/alfa-rules/src/sia-r62/rule.ts
```diff
@@ -230,21 +169,57 @@ function isDistinguishable(
...
-      return true;
-    }
+function hasDistinguishableTextDecoration(
+  container: Element,
+  device: Device,
+  context?: Context
+): Predicate<Element> {
+  return (element) =>
+    test(not(hasTextDecoration(device, context)), container) &&
+    test(hasTextDecoration(device, context), element);
```
Jym77, line 196: We could have the reverse situation (container with text decoration and element without)
```suggestion
    test((hasTextDecoration(device, context)), container) !==
    test(hasTextDecoration(device, context), element);
```

packages/alfa-rules/test/sia-r62/rule.spec.tsx
```diff
@@ -41,45 +40,79 @@ test(`evaluate() passes an <a> element with a <p> parent element with non-link
...
+test(`evaluate() fails an <a> element that removes the default text decoration
+      without replacing it with another distinguishing feature`, async (t) => {
   const target = <a href="#">Link</a>;
 
   const document = Document.of(
     [<p>Hello {target}</p>],
     [
       h.sheet([
-        h.rule.style("a:hover, a:focus", {
+        h.rule.style("a", {
```
Jym77, line 51: We should keep some cases (likely not this one) where the "bad" styling is only here in one of the 6 states. It is a design choice to require distinguishability in each of these states (rather than only on focus/hover), so having test cases will avoid regressions on this choice.

Jym77 (changes requested):

packages/alfa-rules/src/sia-r62/rule.ts
kasperisager: Text decoration is applied to all descendant text so the reverse situation won't result in the link being distinguishable:

```html
<!DOCTYPE html>

<style>
  p {
    text-decoration: underline;
  }

  a {
    text-decoration: none;
  }
</style>

<p>Hello <a href="#">world</a></p>
```

rendered output

packages/alfa-rules/src/common/predicate/has-border.ts
kasperisager: That should be handled by logical property mapping when implemented. Until then, I'd rather not have seemingly partial support for logical properties.

packages/alfa-rules/test/sia-r62/rule.spec.tsx
kasperisager: Good point 👍

Jym77 (approved):

PR #763, opened 2021-04-13
Implement remaining `display` grammar

### LINKED ISSUE #761: `display` property is only partially implemented and does not accept several valid values
The `display` property currently only accepts the following values in Alfa:

- `content`
- `none`
- `block`
- `inline`
- `run-in`
- `flow`
- `flow-root`
- `table`
- `flex`
- `grid`
- `ruby`
- `inline-block`
- `inline-table`
- `inline-flex`
- `inline-grid`

That's obviously not the full grammar: https://drafts.csswg.org/css-display/#propdef-display

- [ ] Follow up on DEV-11517.

kasperisager:
These changes are breaking as the specified and computed type of the `display` property has changed from a plain tuple type to the `Tuple` class.

Resolves #761.

### DIFF
--- a/packages/alfa-style/src/property/display.ts
+++ b/packages/alfa-style/src/property/display.ts
@@ -1,7 +1,10 @@
-import { Keyword } from "@siteimprove/alfa-css";
+import { Keyword, Token } from "@siteimprove/alfa-css";
 import { Parser } from "@siteimprove/alfa-parser";
+import { Err, Result } from "@siteimprove/alfa-result";
+import { Slice } from "@siteimprove/alfa-slice";
 
 import { Property } from "../property";
+import { Tuple } from "./value/tuple";
 
 const { map, either } = Parser;
 
@@ -14,11 +17,15 @@ declare module "../property" {
 /**
  * @internal
  */
-export type Specified =
-  | readonly [Specified.Internal]
-  | readonly [Specified.Box]
-  | readonly [Specified.Outside, Specified.Inside]
-  | readonly [Specified.Outside, Specified.Inside, Specified.ListItem];
+export type Specified = Tuple<
+  | [
+      outside: Specified.Outside,
+      inside: Specified.Inside,
+      listitem?: Specified.ListItem
+    ]
+  | [outside: Specified.Internal, inside: Specified.Internal | Specified.Inside]
+  | [Specified.Box]
+>;
 
 /**
  * @internal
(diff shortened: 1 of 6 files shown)


packages/alfa-style/src/property/display.ts
```diff
@@ -76,54 +83,172 @@ export namespace Specified {
  */
 export type Computed = Specified;
 
+/**
+ * {@link https://drafts.csswg.org/css-display/#typedef-display-outside}
+ */
+const parseDisplayOutside = Keyword.parse("block", "inline", "run-in");
+
+/**
+ * {@link https://drafts.csswg.org/css-display/#typedef-display-inside}
```
Jym77, line 92: We have several properties for which we link to CSS Drafts, and other to MDN. Part of the MDN switch seems to trace back to
https://github.com/Siteimprove/alfa/commit/ba1c4743e9c79c9f38988721f079c8911d41dc2e

I'm not sure which is best, but we should stick to one and not mix both.

packages/alfa-style/src/property/display.ts
```diff
@@ -14,11 +17,15 @@ declare module "../property" {
...
-  | readonly [Specified.Outside, Specified.Inside, Specified.ListItem];
+export type Specified = Tuple<
+  | [
+      outside: Specified.Outside,
+      inside: Specified.Inside,
+      listitem?: Specified.ListItem
+    ]
+  | [outside: Specified.Internal, inside: Specified.Inside]
+  | [Specified.Box]
+>;
```
Jym77, line 28: The alternate on the parameter is less precise and I find it more confusing than putting it outside of `Tuple`.
```suggestion
export type Specified = Tuple<[
      outside: Specified.Outside,
      inside: Specified.Inside,
      listitem?: Specified.ListItem
    ]>
  | Tuple<[outside: Specified.Internal, inside: Specified.Inside]>
  | Tuple<[Specified.Box]>;
```

packages/alfa-style/src/property/display.ts
```diff
@@ -76,54 +83,172 @@ export namespace Specified {
...
+  "table",
+  "flex",
+  "grid",
+  "ruby"
+);
+
+/**
+ * {@link https://drafts.csswg.org/css-display/#typedef-display-listitem}
+ */
+const parseDisplayListItem = Keyword.parse("list-item");
```
Jym77, line 106: Do we need to repeat `Display` on the names of these functions? They're already namespaced and anyway not exported.

Jym77 (approved):

packages/alfa-style/src/property/display.ts
kasperisager: The properties themselves always link to MDN as it provides very broad information aimed at callers. Only the internal bits link to the CSS specifications themselves.

packages/alfa-style/src/property/display.ts
kasperisager: I don't agree. Splitting the type like this is only needed if it should be possible to later narrow the union to a subset of its individual members, which isn't the case here. It's variations of the _same_ `display` value.

packages/alfa-style/src/property/display.ts
kasperisager: Nope, I'll remove them 👍

packages/alfa-style/src/property/display.ts
Jym77: In that very case, for me it's more of a legibility thing.

I immediately read `Tuple<[A, B]> | Tuple<[C, D]>` as "either (a, b) or (c, d)" while `Tuple<[A, B] | [C, D]>` makes me think in order to get that it cannot be (a, d) nor (c, b) nor some other weird combination. The outer alternative feels clearer to indicate that it's "a tuple like this or a tuple like that" while the inner alternative looks like it could possibly be a way to mix the arguments. Of course, it is different from `Tuple<[A | C, B | D]>` but I need more thinking to reach the same conclusion (and it also looks a bit like `Tuple<[A | B, C | D]>` with just slightly different separators).

Jym77 (approved):

Issue #751, opened 2021-03-23
Security restrictions for the `:visited` pseudo-class are not accounted for


kasperisager:
This was split from #708. See https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector.

Jym77:
These security restrictions are also triggered without the presence of a `:visited` selector in rules:
c

```html
<style>
a {
  color: red;
  text-decoration: none;
}  
a:link {
  color: blue;
  text-decoration: underline;
}
</style>

This is <a href="https://example.org">a link</a> to test.

```
When the link is visited, the `text-decoration` property is read as if the link was not visited (because `text-decoration` is not allowed on visited links) and therefore the `a:link` rule **does apply** and the decoration is kept 🤯

Jym77:
As a consequence, the second expectation of R62 can almost never fail (once this is fixed) (if the first one passes):
* `hasTextDecoration`, `hasOutline` and `hasBorder` => only the color may change between `:link` and `:visited`, and not the transparency. So result will be the same.
* `hasDistinguishableBackground` => only color may change, so we can make it non-distinguishable by choosing the same as the container's background color, but that's it.

PR #771, opened 2021-04-20
Develop branch

### LINKED ISSUE #709: Implement `font-variant-*` CSS properties
**Shorthands:**

- [x] [`font-variant`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant)

**Longhands:**

- [x] [`font-variant-ligatures`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-ligatures)
- [x] [`font-variant-caps`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-caps)
- [x] [`font-variant-numeric`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric)
- ~~[ ] [`font-variant-alternates`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-alternates)~~ (only supported by Firefox, not present in CSS Fonts Level 3, super annoying and depending on `@font-feature`…)
- [x] [`font-variant-east-asian`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-east-asian)
- [x] [`font-variant-position`](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-position) (not a longhand of `font-variant`)

### LINKED ISSUE #738: Account for additional focus indicator types in SIA-R65
In continuation of #658, this issue will track additional focus indicator types that SIA-R65 should support.

- [x] **The `border-color` property**

  This property is often used as the single means of providing a focus indicator for widgets with a solid background that doesn't itself change on focus, such as outlined buttons and inputs:

  ```html
  <style>
    button {
      background-color: #fff;
      border: 2px solid #eee;
    }

    button:focus {
      outline: none;
      border-color: #aaa;
    }
  </style>

  <button>Hello world</button>
  ```

  Depends on #718.

  - [x] Follow up on DEV-11365.

### LINKED ISSUE #748: SIA-R66 and SIA-R69 fail to account for background images not used as background images
Confused? So was I 🙈 This bit of HTML and CSS currently fails both rules due to seemingly poor contrast:

```html
<style>
  h1 {
    color: #333;
    background-image: linear-gradient(
      to right,
      rgb(0, 0, 0) 0%,
      rgb(0, 0, 0) 100%
    );
    background-repeat: repeat-x;
    background-position: 0px 100%;
    background-size: 100% 2px;
  }
</style>

<h1>Hello world</h1>
```

This is what it renders:

Rendered output

Yep, that's a background image being used to create a border. Because, why not? 🤷 We need to deal with that.

- [x] Follow up on DEV-11379.

### LINKED ISSUE #768: Functional pseudo-classes selectors need to reimplement `equals`
Parametric pseudo-classes like `:nth-of-type` inherit their `equals` (and `toString`) from parent class: https://github.com/Siteimprove/alfa/blob/main/packages/alfa-selector/src/selector.ts#L1573-L1610

This result in `equals` ignoring the parameter and thus, for example, `:nth-of-type(1)` to `equals` `:nth-of-type(2)`!

In turn, this cause the styling rule tree to be incorrect since, for example, a rule matching `:nth-of-type(2)` will be children of a rule matching `:nth-of-type(1)` and properties not defined in the former will (incorrectly) be searched in the latest.

So, we need to re-implement `equals` in all the inherited selectors that depend on more than their name. We should also reimplement `toString` for easier reporting (currently, both `:nth-of-type(1)` and `:nth-of-type(2)` are stringified as simply `":nth-of-type"`.

- [x] Follow up on [DEV-11605](https://siteimprove-wgs.atlassian.net/browse/DEV-11605)

### LINKED ISSUE #769: Text inside element with role="presentation" is omitted from Accessible Name
It seems that Alfa completely disregards contents of an element with `role="presentation"` when it comes to calculating an Accessible Name.

`<button><span role="presentation">Text</span></button>` is flagged as having no Accessible Name, even though `role="presentation"` is only supposed to remove the semantics (if any) from the node. `role="presentation"` isn't inherited to child nodes. Chrome's DevTools also seems to tell me that the button has the Accessible Name "Text".

### LINKED ISSUE #776: Step 1 of name computation should be skipped when descending.
Step 2F iii b jumps to step 2, thus skipping Step 1 and the "role prohibiting naming" bit. So step 1 should be skipped if `state.isDescending` is true.
Currently, we give no name to
```
<a href="#">
      <p>Hello world</p>
</a>
```
because role of `paragraph` prevents name from content.

### LINKED ISSUE #784: Parsing of background-size is wrong
It doesn't parse percentages… We also need to add some tests…

- [x] Follow up on [DEV-11704](https://siteimprove-wgs.atlassian.net/browse/DEV-11704)

### LINKED ISSUE #787: hasRole should use the accessible node
`hasRole` grabs the role with `Role.from`. This does, among other, ignores stuff like presentational children.
`hasRole` should do like `hasAccessibleName` and grabs the role with `Node.from(…).role`

- [x] Follow up on DEV-11705.

### LINKED ISSUE #797: Provide extended diagnostic for R62
mockup: https://siteimprove-wgs.atlassian.net/browse/CE-5395

- Build extended diagnostic, sending the computed style for each state.
  - the full computed style needs to be filtered and only the properties relevant for the rule (plus `color`) should be sent, to keep size small.

### LINKED ISSUE #798: Provide extended diagnostic for R16
mockup: https://siteimprove-wgs.atlassian.net/browse/CE-5396

- Send the missing required attributes.
- Send the present required attributes.
- Send the role.

Look at R14 for example extended diagnostic.

### LINKED ISSUE #815: Support `text-decoration-thickness`
- [x] Follow up on DEV-11787

### LINKED ISSUE #830: `Name.fromDescendant()` should reject empty spaces names
```html
<a href="#" title="Hello world">
      <span> </span>
</a>
```
Receives ` ` (space) instead of `Hello world` as a name, because the content takes precedence and `Name.fromDescendant()` returns `Some " "` rather than `None`.
Both Chrome and Firefox give `Hello World` as name.

- [ ] Follow up on DEV-11839
- [ ] Follow up on DEV-11885

Jym77:
Resolves #709, resolves #738, resolves #748, resolves #768, resolves #769, resolves #776, resolves #784, resolves #787, resolves #797, resolves #798, resolves #815, resolves #830

* Add some tests for R3 (#792), R4 (#793), R5 (#794), R6 (#795), R8 (#806), R11 (#770), R12 (#800), R17 (#791), R18 (#803), R19 (#807), R42 (#786), `background-size` (#788).
* R62: removes second expectation, as it mostly causes false positives due to #751 (#772).
* R62: accept differences in `font-weight` as good (dirty patch, need discussion) (#779).
* Implement `#equals` and `#toString` for functional pseudo-classes and elements (#773).
* Accessible name: Skip step 1 of name computation when descending (#778).
* R14: add extended diagnostic (#786).
* R16: add extended diagnostic (#804).
* R62: add extended diagnostic (#801).
* `background-size`: accept percentages, accept 2 values, use `Tuple` to store value (#788).
* R66, R69: if a background has a `background-size`, do not try to resolve its colour and return `cantTell`; we need layout info to figure out if the background actually goes under the text or not (#789).
* Fix `hasRole` to use the role from the accessible node rather than `Role.from()` (#805).
* Support `text-decoration-thickness` (#817)
* Increase nodejs heap size in actions (was often running out of memory) (#818, #822)
* R65: accept difference in `border` as a valid focus indicator (#819)
* Support `font-variant` and its longhands, plus `font-variant-position` (#821)
* Implement R56 (#829)
* Remove whitespace only descendants from accessible name computation (#831)


packages/alfa-aria/src/role.ts
```diff
@@ -241,7 +241,6 @@ export class Role<N extends Role.Name = Role.Name>
 
     return false;
   }
-
```
kasperisager, line 244: This should not have been removed.

.github/workflows/integrate.yml
```diff
@@ -11,9 +11,11 @@ jobs:
     name: Build
     strategy:
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ ubuntu-latest, windows-latest ]
         node: [12, 14]
     runs-on: ${{ matrix.os }}
+    env:
+      NODE_OPTIONS: "--max-old-space-size=4096"
```
kasperisager, line 18: I'd be _very_ interested in figuring out why we need 4 GB of memory for running our test suite as that seems... excessive 🙈 Did the OOM errors only occur on Windows?

README.md
```diff
@@ -131,7 +131,7 @@ Alfa will run in any [ECMAScript 2018](https://www.ecma-international.org/ecma-2
 In order to build Alfa, a recent version (>= 12) of [Node.js](https://nodejs.org/) is required in addition to the [Yarn](https://yarnpkg.com/) package manager. Once Node.js and Yarn are installed, go ahead and install the Alfa development dependencies:
 
 ```console
-$ yarn install
+$ yarn install --immutable
```
kasperisager, line 134: Immutable installs should not be used when developing Alfa as we want lockfile changes to be committed.

packages/alfa-rules/src/common/predicate/has-border.ts
```diff
@@ -17,9 +17,6 @@ export function hasBorder(
     return sides.some(
       (side) =>
         style.computed(`border-${side}-width` as const).none(Length.isZero) &&
-        style
-          .computed(`border-${side}-style` as const)
-          .none((style) => style.value === "none") &&
```
kasperisager, line 22: Why was this removed? `border-left-style: none` will result in there being no left border regardless of the other `border-left-*` properties, no?

packages/alfa-rules/src/sia-r56/rule.ts
```diff
@@ -0,0 +1,166 @@
...
+          .filter(Element.isElement)
+          .filter(
+            and(
+              hasNamespace(equals(Namespace.HTML)),
+              not(isIgnored(device)),
+              hasRole(device, (role) => role.is("landmark"))
+            )
+          )
+          .reduce((groups, landmark) => {
+            // since we already have filtered by having a landmark role, we can
```
kasperisager, line 38:
```suggestion
            // Since we already have filtered by having a landmark role, we can
```

packages/alfa-rules/src/sia-r56/rule.ts
```diff
@@ -0,0 +1,166 @@
...
+            // safely get the role.
+            const role = Node.from(landmark, device).role.get()!;
```
kasperisager, line 40:
```suggestion
            const role = Node.from(landmark, device).role.get();
```

Assuming that `Node#role` is of type `Option<Role>` then a non-null assertion isn't needed as `Option<Role>#get()` will return `Role`.

packages/alfa-rules/src/sia-r56/rule.ts
```diff
@@ -0,0 +1,166 @@
...
+
+            return groups;
+          }, Map.empty<Role, List<Element>>())
+          .filter((elements) => elements.size > 1)
+          .map(Group.of)
+          .values();
+      },
+
+      expectations(target) {
+        // empty groups have been filtered out already, so we can safely get the
```
kasperisager, line 58:
```suggestion
        // Empty groups have been filtered out already, so we can safely get the
```

packages/alfa-rules/src/sia-r56/rule.ts
```diff
@@ -0,0 +1,166 @@
...
+        // first element
+        const role = Node.from(
+          Iterable.first(target).get()!,
+          device
+        ).role.get()!.name;
```
kasperisager, line 63:
```suggestion
        ).role.get().name;
```

Same as above.

packages/alfa-rules/src/sia-r62/rule.ts
```diff
@@ -223,3 +240,222 @@ function hasDistinguishableBackground(
       .none((color) => Color.isTransparent(color) || color.equals(reference));
   };
 }
+
+/**
+ * Check if an element has a different font weight than its container.
+ *
+ * This is brittle and imperfect but removes a strong pain point until we find
+ * a better solution…
```
kasperisager, line 248:
```suggestion
 * a better solution.
```

packages/alfa-rules/test/sia-r11/rule.spec.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import R11, { Outcomes } from "../../src/sia-r11/rule";
+
+import { evaluate } from "../common/evaluate";
+import { passed, failed, inapplicable } from "../common/outcome";
+
+test(`evaluate() passes a link with a name given by content`, async (t) => {
+  const target = <a href="#">Foo</a>;
+
+  const document = Document.of([target]);
```
kasperisager, line 12: This should use the `h.document()` API instead. We should actually go through all existing tests at some point and change this. The HyperScript-like `h` API is designed for convenience, unlike the DOM constructor functions, and is therefore favoured for tests.

packages/alfa-style/src/property/text-decoration-thickness.ts
```diff
@@ -0,0 +1,63 @@
...
+ * @internal
+ */
+export const parse = either<Slice<Token>, Specified, string>(
+  Keyword.parse("auto", "from-font"),
+  Length.parse,
+  Percentage.parse
+);
+
+/**
+ * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/outline-width}
```
kasperisager, line 40:
```suggestion
 * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-thickness}
```

kasperisager (changes requested):

.github/workflows/integrate.yml
kasperisager: Oh, it actually seems to be the build failing 🤔

Jym77: 4GB are probably too much. I've just used the ~rule of thumb~ Bayes' estimation with power law prior of doubling the current value 🤷 https://github.com/Siteimprove/alfa/runs/2728515736

We may want to tighten that amount, of course…

README.md
Jym77: My reasoning here was that we mostly want to update lockfile when we want to update lockfile, not when we work on unrelated stuff.

packages/alfa-rules/src/common/predicate/has-border.ts
Jym77: It's handled by https://github.com/Siteimprove/alfa/blob/main/packages/alfa-style/src/property/border-top-width.ts#L48-L54 (and the previous test).

README.md
kasperisager: CI will fail anyway if the lockfile isn't up to date on push so there's no reason at all to do immutable installs when developing. It might actually be harmful as we'll then have to keep in mind that developers should specifically do mutable installs, i.e. omit `--immutable`, when modifying package manifests. Omitting `--immutable` will always do the right when developing.

packages/alfa-rules/src/common/predicate/has-border.ts
kasperisager: Good point!

packages/alfa-rules/test/sia-r11/rule.spec.tsx
Jym77: Feels like this updated all the tests 😅 https://github.com/Siteimprove/alfa/pull/771/commits/ce61ed613795fbe131cffa8a462fffd141efa171

kasperisager (changes requested):
See https://github.com/Siteimprove/alfa/pull/771#discussion_r649974449.

kasperisager (approved):
💯👏

PR #783, opened 2021-05-03
Handle arrays in Equatable.equals

Jym77:
`Equatable.equals([1, 2], [1,2])` is currently `false` while `Equatable.equals(List.of([1, 2]), List.of([1,2]))` is `true` (same for other iterables AFAICT), which is a bit confusing given that `List` is mostly a wrapper for arrays.

This handle arrays in `Equatable.equals`

### DIFF
--- a/packages/alfa-equatable/src/equatable.ts
+++ b/packages/alfa-equatable/src/equatable.ts
@@ -35,6 +35,12 @@ export namespace Equatable {
       return b.equals(a);
     }
 
+    if (Array.isArray(a) && Array.isArray(b)) {
+      return (
+        a.length === b.length && a.every((value, i) => equals(value, b[i]))
+      );
+    }
+
     return false;
   }
 }


kasperisager:
`Equatable.equals()` and `Hash#writeUnknown()`, both of which are the main entry points for our hashing based data structures, must maintain symmetry. I therefore think the better approach would be to make both of those behave "intuitively" when passed iterable objects, that is test equality of and hash each item of the passed iterable(s). Either way, there are likely deeper implications that we'll have to consider carefully first, such as if any current code paths depend on arrays being treated as reference equal. Not that that wouldn't be silly, but you never known 🙈

Jym77:
Yes, I wasn't sure how much design would break with that one… At least, tests are passing, so we probably don't rely too much on the current behaviour 😁 
Also figured out that this wasn't the correct way to solve the problem I was looking at anyway…

PR #792, opened 2021-05-07
SIA R3: add test

elenamongelli:


packages/alfa-rules/test/sia-r3/rule.spec.tsx
```diff
@@ -0,0 +1,85 @@
...
+      passed(R3, target2, {
+        1: Outcomes.HasUniqueId,        
+      }),
+      passed(R3, target3, {
+        1: Outcomes.HasUniqueId,        
+      }),
+    ]);
+  });
+
+test("evaluate() fails unique id attributes in the document context.", async (t) => {
```
Jym77, line 42:
```suggestion
test("evaluate() fails duplicated id attributes in the document context.", async (t) => {
```

packages/alfa-rules/test/sia-r3/rule.spec.tsx
```diff
@@ -0,0 +1,85 @@
...
+        failed(R3, target1, {
+        1: Outcomes.HasNonUniqueId,        
+      }),
+        failed(R3, target2, {
+        1: Outcomes.HasNonUniqueId,        
+      })      
+    ]);
+  });
+
+  test("evaluate() fails unique id attributes in the document context.", async (t) => {
```
Jym77, line 58:
```suggestion
  test("evaluate() fails duplicated id attributes on SVG element.", async (t) => {
```
The difference between this test and the previous is that this one is checking an SVG element, not only HTML elements. And so the reason this test is here is to make sure that the rule doesn't ignore SVG elements. And so this should be what the description says 😀

Jym77 (changes requested):

PR #794, opened 2021-05-07
SIA R5: add test

elenamongelli:


packages/alfa-rules/test/sia-r5/rule.spec.tsx
```diff
@@ -0,0 +1,63 @@
+import { test } from "@siteimprove/alfa-test";
+
+import { Document } from "@siteimprove/alfa-dom";
+
+import R5, { Outcomes } from "../../src/sia-r5/rule";
+
+import { evaluate } from "../common/evaluate";
+import { passed, failed, inapplicable } from "../common/outcome";
+
+test(`evaluate() passes an html element with lang attribute whose value is a valid primary language subtag`, async (t) => {
```
Jym77, line 10:
```suggestion
test(`evaluate() passes a lang attribute with valid primary tag`, async (t) => {
```
We can be a bit more terse here, the full technically-precise language is needed on ACT rules but not really in the context of Alfa.
Plus, the rule (in Alfa) targeting the attribute rather than the element, we can match that in the description.

packages/alfa-rules/test/sia-r5/rule.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const target = <html lang="en"></html>;
```
Jym77, line 11: Technically, this is not the target, so it should have another name. `html` is likely a good name.
```suggestion
    const html = <html lang="en"></html>;
```
The name change must be propagated.
The same change should be done on all tests.

packages/alfa-rules/test/sia-r5/rule.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const document = Document.of([target]);
+  
+    t.deepEqual(await evaluate(R5, { document }), [
+      passed(R5, target.attribute("lang").get(), {
+        1: Outcomes.HasValidLanguage,
+      }),
+    ]);
+  });
+
+  test(`evaluate() passes an html element with lang attribute whose value is a  valid language tag even though the region subtag is not`, async (t) => {
```
Jym77, line 22:
```suggestion
  test(`evaluate() passes a lang attribute with valid primary tag and invalid region subtag`, async (t) => {
```

packages/alfa-rules/test/sia-r5/rule.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const document = Document.of([target]);
+  
+    t.deepEqual(await evaluate(R5, { document }), [
+      passed(R5, target.attribute("lang").get(), {
+        1: Outcomes.HasValidLanguage,
+      }),
+    ]);
+  });
+
+  test(`evaluate() fails an html element with lang attribute whose value is not a valid language tag.`, async (t) => {
```
Jym77, line 34:
```suggestion
  test(`evaluate() fails a lang attribute with invalid primary tag.`, async (t) => {
```

packages/alfa-rules/test/sia-r5/rule.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const target = <html lang="em-US"></html>;
```
Jym77, line 35:
```suggestion
    const target = <html lang="invalid"></html>;
```
Here, the ACT rule was trying to have a "realistic" example (typo `em` instead of `en`). I'd prefer for us to have a "blatantly wrong" example to avoid spending time reading it and trying to understand what is wrong.

packages/alfa-rules/test/sia-r5/rule.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const target = <html lang="#1"></html>;
+  
+    const document = Document.of([target]);
+  
+    t.deepEqual(await evaluate(R5, { document }), [
+      failed(R5, target.attribute("lang").get(), {
+        1: Outcomes.HasNoValidLanguage,
+      }),
+    ]);
+  });
```
Jym77, line 56:
```suggestion
```
This is fairly redundant with the previous case. We don't need to check every possible way that a language tag can be invalid (ACT rule does want to showcase it, so there it make more sense to have both).

packages/alfa-rules/test/sia-r5/rule.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  test(`evaluate() is inapplicable to svg elements.`, async (t) => {
+    const target = <svg xmlns="http://www.w3.org/2000/svg" lang="fr"></svg>;
+    const document = Document.of([target]);
+
+    t.deepEqual(await evaluate(R5, { document }), [inapplicable(R5)]);
+});
```
Jym77, line 63:
```suggestion
});

```
(missing newline at end of file).

Jym77 (changes requested):

PR #795, opened 2021-05-07
SIA R6: Added test

elenamongelli:


packages/alfa-rules/test/sia-r6/rule.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+    t.deepEqual(await evaluate(R6, { document }), [
+      passed(R6, target, {
+        1: Outcomes.HasMatchingLanguages,
+      }),
+    ]);
+  });
+
+  test("evaluate() passes an html element which has identical primary language subtags for its lang and xml:lang attributes. The extended language subtags also match.", async (t) => {
+    const target = <html lang="en-GB" xml:lang="en-GB"></html>;
+
```
Jym77, line 24:
```suggestion
  test("evaluate() passes an html element which has identical primary language subtags for its lang and xml:lang attributes. The extended language subtags don't match.", async (t) => {
    const target = <html lang="en-GB" xml:lang="en-US"></html>;
  
```
That's a case where ACT rules want to show "good code" in its passing example (so perfect match between the attributes), but we want to test the border conditions of the code.
The rule is only checking for the primary tag, not the region subtag. To make sure this is correctly done, we need a case where only the primary tags match.

packages/alfa-rules/test/sia-r6/rule.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+  test(`evaluate() is inapplicable to svg elements.`, async (t) => {
+        const target = <svg xmlns="http://www.w3.org/2000/svg" lang="en" xml:lang="en"></svg>;
+  
+        const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R6, { document }), [inapplicable(R6)]);
+});
+
+test(`evaluate() is inapplicable to html elements whose lang attribute is not a valid language tag.`, async (t) => {
+    const target = <html lang="em" xml:lang="en"></html>;
```
Jym77, line 67:
```suggestion
    const target = <html lang="invalid" xml:lang="en"></html>;
```
Le's make it obvious that it is invalid.

packages/alfa-rules/test/sia-r6/rule.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+    const document = Document.of([target]);
+
+t.deepEqual(await evaluate(R6, { document }), [inapplicable(R6)]);
+});
+
+test(`evaluate() is inapplicable to html elements with an empty ("") xml:lang attribute..`, async (t) => {
```
Jym77, line 74:
```suggestion
test(`evaluate() is inapplicable to html elements with an empty ("") xml:lang attribute.`, async (t) => {
```

packages/alfa-rules/test/sia-r6/rule.spec.tsx
```diff
@@ -0,0 +1,80 @@
...
+    const target = <html lang="fr" xml:lang=""></html>;
+
+    const document = Document.of([target]);
+
+t.deepEqual(await evaluate(R6, { document }), [inapplicable(R6)]);
+});
```
Jym77, line 80:
```suggestion
});

```

Jym77 (changes requested):

PR #800, opened 2021-05-11
SIA R12: Add Test

elenamongelli:

### DIFF
--- a/packages/alfa-rules/test/sia-r12/rule.spec.tsx
+++ b/packages/alfa-rules/test/sia-r12/rule.spec.tsx
@@ -90,3 +90,35 @@ test(`evaluates() fails element with <button> role because it doesn't have a con
     }),
   ]);
 });
+
+test(`evaluate() is inapplicabile to input element with image attribute`, async (t) => {
+  const target = <input type="image" value="download" alt="Download" />;
+
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R12, { document }), [inapplicable(R12)]);
+});
+
+test(`evaluate() is inapplicabile to input element with image attribute`, async (t) => {
+  const target = <input type="image" value="download" alt="Download" />;
+
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R12, { document }), [inapplicable(R12)]);
+});
+
+test(`evaluate() is inapplicabile to element with no button role`, async (t) => {
+  const target = <div>Press Here</div>;
+
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R12, { document }), [inapplicable(R12)]);
+});
+
+test(`evaluate() is inapplicabile to button element with none role`, async (t) => {
+  const target = <button role="none" disabled></button>;
+
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R12, { document }), [inapplicable(R12)]);
+});
--- a/packages/alfa-rules/test/sia-r12/rule.spec.tsx
+++ b/packages/alfa-rules/test/sia-r12/rule.spec.tsx
@@ -1,0 +1,56 @@
+import { test } from "@siteimprove/alfa-test";
+
+import { Document } from "@siteimprove/alfa-dom";
+
+import R12, { Outcomes } from "../../src/sia-r12/rule";
+
+import { evaluate } from "../common/evaluate";
+import { passed, failed, inapplicable } from "../common/outcome";
+
+test(`evaluates() passes a <button> with accessible name `, async (t) => {
+  const target = <button>My button</button>;
+
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R12, { document }), [
+    passed(R12, target, {
+      1: Outcomes.HasName,
+    }),
+  ]);
+});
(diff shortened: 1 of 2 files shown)


packages/alfa-rules/test/sia-r12/rule.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R12, { document }), [
+    passed(R12, target, {
+      1: Outcomes.HasName,
+    }),
+  ]);
+});
+
+test(`evaluates() passes an input element with accessible value attribute `, async (t) => {
```
Jym77, line 22:
```suggestion
test(`evaluates() passes a submit button element with an accessible name given by the value attribute `, async (t) => {
```
* the rule is checking name of buttons, this `<input>` has a type of `submit` which makes it a "submit button" 🤷 We want the description to point out it is a button since that explains why this rule is testing it.
* "accessible name" is the technically correct term, so we also want to use it in descriptions.

packages/alfa-rules/test/sia-r12/rule.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const document = Document.of([input]);
+
+  t.deepEqual(await evaluate(R12, { document }), [
+    passed(R12, input, {
+      1: Outcomes.HasName,
+    }),
+  ]);
+});
+
+test(`evaluates() passes a <button> with aria label attribute`, async (t) => {
```
Jym77, line 34:
```suggestion
test(`evaluates() passes a <button> with a name given by the \`aria-label\` attribute`, async (t) => {
```

packages/alfa-rules/test/sia-r12/rule.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R12, { document }), [
+    passed(R12, target, {
+      1: Outcomes.HasName,
+    }),
+  ]);
+});
+
+test(`evaluates() fails <button> with no attribute`, async (t) => {
```
Jym77, line 46:
```suggestion
test(`evaluates() fails a button with no accessible name`, async (t) => {
```

packages/alfa-rules/test/sia-r12/rule.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const target = <button type="button" value="read more"></button>;
+
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R12, { document }), [
+    failed(R12, target, {
+      1: Outcomes.HasNoName,
+    }),
+  ]);
+});
```
Jym77, line 68:
```suggestion
```
I think that test is not needed. It's only checking that we correctly ignore `type` and `value` on `<button>` elements, which is not specific to this rule.

packages/alfa-rules/test/sia-r12/rule.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+test(`evaluates() fails element with <button> role because it doesn't have a content with accessible name`, async (t) => {
```
Jym77, line 70:
```suggestion
test(`evaluates() fails an element with \`button\` role because and without an accessible name`, async (t) => {
```

packages/alfa-rules/test/sia-r12/rule.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const target = <span role="button"></span>;
+
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R12, { document }), [
+    failed(R12, target, {
+      1: Outcomes.HasNoName,
+    }),
+  ]);
+});
```
Jym77, line 92:
```suggestion
```
Seems to be duplicated from previous test.

packages/alfa-rules/test/sia-r12/rule.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+test(`evaluate() is inapplicabile to input element with image attribute`, async (t) => {
```
Jym77, line 94:
```suggestion
test(`evaluate() is inapplicable to image buttons`, async (t) => {
```

packages/alfa-rules/test/sia-r12/rule.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+  t.deepEqual(await evaluate(R12, { document }), [inapplicable(R12)]);
+});
+
+test(`evaluate() is inapplicabile to input element with image attribute`, async (t) => {
+  const target = <input type="image" value="download" alt="Download" />;
+
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R12, { document }), [inapplicable(R12)]);
+});
```
Jym77, line 108:
```suggestion
```
Duplicated test.

Jym77 (changes requested):

elenamongelli:
changes implements

Jym77 (approved):

PR #804, opened 2021-05-17
R16 extended diagnostic

### LINKED ISSUE #798: Provide extended diagnostic for R16

elenamongelli:
Resolves #798


packages/alfa-rules/src/sia-r16/rule.ts
```diff
@@ -32,53 +35,175 @@ export default Rule.Atomic.of<Page, Element>({
...
+
+    for (const attribute of required) {
+      if (node.attribute(attribute).every(property("value", isEmpty))) {
+        missing.push(attribute);
+        result = false;
+      } else {
+        found.push(attribute);
       }
     }
+  }
```
Jym77, line 87: Since we end up using `role.requiredAttributes` only here (rather than twice: here and in the expectation), I think it's better to roll back to the original version so that we're not leaking the stuff about `separator` being a mess in two places.
```suggestion
    for (const attribute of role.attributes) {
      if (role.isAttributeRequired(attribute)) {
        if (node.attribute(attribute).every(property("value", isEmpty))) {
          missing.push(attribute);
          result = false;
        } else {
          found.push(attribute);
        }
      }
    }
  }
```

packages/alfa-aria/src/role.ts
```diff
@@ -241,7 +241,19 @@ export class Role<N extends Role.Name = Role.Name>
...
+   * The `separator` role is poorly architected in the sense that its
+   * inheritance and attribute requirements depend on aspects of the element
+   * carrying the role. If the element is not focusable, the `separator`
+   * role has no required attributes.
+   *
+   * It is up to the caller to make that distinction.
+   */
+  public requiredAttributes(): ReadonlyArray<Attribute.Name> {
+    return this.attributes.filter((attr) => this.isAttributeRequired(attr));
+  }
```
Jym77, line 256: Let's keep the `separator` mess only in the rule.
```suggestion
```

Jym77 (changes requested):

PR #806, opened 2021-05-17
SIA-R8: Added tests

elenamongelli:
Resolves #

### DIFF
--- a/packages/alfa-rules/test/sia-r8/rule.spec.tsx
+++ b/packages/alfa-rules/test/sia-r8/rule.spec.tsx
@@ -1,0 +1,200 @@
+import { test } from "@siteimprove/alfa-test";
+
+import { Document } from "@siteimprove/alfa-dom";
+
+import R8, { Outcomes } from "../../src/sia-r8/rule";
+
+import { evaluate } from "../common/evaluate";
+import { passed, failed, inapplicable } from "../common/outcome";
+
+test("evaluate() passes an input element with accessible name", async (t) => {
+  const target = <input />;
+
+  const label = (
+    <label>
+      first name
+      {target}
+    </label>
+  );
+
+  const document = Document.of([label]);
+
+  t.deepEqual(await evaluate(R8, { document }), [
+    passed(R8, target, {
+      1: Outcomes.HasName,
+    }),
+  ]);
+});
+
+test("evaluate() passes an input element with aria label", async (t) => {
+  const target = <input aria-label="last name" disabled />;
+
+  const label = <div>last name</div>;
+
+  const document = Document.of([label, target]);
+
+  t.deepEqual(await evaluate(R8, { document }), [
+    passed(R8, target, {
+      1: Outcomes.HasName,
+    }),
+  ]);
+});
+
+test("evaluate() passes a select element with accessible options", async (t) => {
+  const target = (
+    <select id="country">
+      <option>England</option>
+      <option>Scotland</option>
+      <option>Wales</option>
+      <option>Northern Ireland</option>
+    </select>
+  );
+
+  const label = <label for="country">Country</label>;
+
+  const document = Document.of([label, target]);
+
+  t.deepEqual(await evaluate(R8, { document }), [
+    passed(R8, target, {
+      1: Outcomes.HasName,
+    }),
+  ]);
+});
+
+test("evaluate() passes a textarea element with aria labelledby", async (t) => {
+  const target = <textarea aria-labelledby="country"></textarea>;
+
+  const label = <div id="country">Country</div>;
+
+  const document = Document.of([label, target]);
+
+  t.deepEqual(await evaluate(R8, { document }), [
+    passed(R8, target, {
+      1: Outcomes.HasName,
+    }),
+  ]);
+});
+
+test("evaluate() passes a input element with placeholder attribute", async (t) => {
+  const target = <input placeholder="Your search query" />;
+
+  const label = <button type="submit">search</button>;
+
+  const document = Document.of([label, target]);
+
+  t.deepEqual(await evaluate(R8, { document }), [
+    passed(R8, target, {
+      1: Outcomes.HasName,
+    }),
+  ]);
+});
+
+test("evaluate() passes a input element with combobox role, whose has an aria-label attribute", async (t) => {
+  const target = (
+    <div aria-label="country" role="combobox" aria-disabled="true">
+      England
+    </div>
+  );
+
+  const label = <div>Country</div>;
+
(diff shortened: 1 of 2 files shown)


packages/alfa-rules/test/sia-r8/rule.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+    }),
+  ]);
+});
+
+test("evaluate() passes an input element with aria label", async (t) => {
+  const target = <input aria-label="last name" disabled />;
+
+  const label = <div>last name</div>;
+
+  const document = Document.of([label, target]);
```
Jym77, line 34: That's a case where the `<div>` is added in ACT rule to provide "good" code (the `<input>` needs a visual label in addition to the accessible name), and we don't care as part of a test suite, so we can keep it small.
```suggestion
  const document = Document.of([target]);
```

packages/alfa-rules/test/sia-r8/rule.spec.tsx
```diff
@@ -0,0 +1,200 @@
+import { test } from "@siteimprove/alfa-test";
+
+import { Document } from "@siteimprove/alfa-dom";
+
+import R8, { Outcomes } from "../../src/sia-r8/rule";
+
+import { evaluate } from "../common/evaluate";
+import { passed, failed, inapplicable } from "../common/outcome";
+
+test("evaluate() passes an input element with accessible name", async (t) => {
```
Jym77, line 10: Let's be more precise on how it gets its name ("implicit" because it is a surrounding element).
```suggestion
test("evaluate() passes an input element with implicit label", async (t) => {
```

packages/alfa-rules/test/sia-r8/rule.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const document = Document.of([label, target]);
+
+  t.deepEqual(await evaluate(R8, { document }), [
+    passed(R8, target, {
+      1: Outcomes.HasName,
+    }),
+  ]);
+});
+
+test("evaluate() passes a select element with accessible options", async (t) => {
```
Jym77, line 43:
```suggestion
test("evaluate() passes a select element with explicit label", async (t) => {
```
("explicit" because there is a `for` attribute that explicitly makes the connection)

packages/alfa-rules/test/sia-r8/rule.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+    }),
+  ]);
+});
+
+test("evaluate() fails a input element with not attribute", async (t) => {
+  const target = <input />;
+
+  const label = <div>last name</div>;
+
+  const document = Document.of([label, target]);
```
Jym77, line 115:
```suggestion
  const document = Document.of([target]);
```

packages/alfa-rules/test/sia-r8/rule.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+  t.deepEqual(await evaluate(R8, { document }), [
+    passed(R8, target, {
+      1: Outcomes.HasName,
+    }),
+  ]);
+});
+
+test("evaluate() fails a input element with not attribute", async (t) => {
```
Jym77, line 110:
```suggestion
test("evaluate() fails a input element without accessible name", async (t) => {
```

packages/alfa-rules/test/sia-r8/rule.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+test("evaluate() passes a input element with combobox role, whose has an aria-label attribute", async (t) => {
+  const target = (
+    <div aria-label="country" role="combobox" aria-disabled="true">
+      England
+    </div>
+  );
+
+  const label = <div>Country</div>;
+
+  const document = Document.of([label, target]);
```
Jym77, line 101:
```suggestion
  const document = Document.of([target]);
```

packages/alfa-rules/test/sia-r8/rule.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+  t.deepEqual(await evaluate(R8, { document }), [
+    failed(R8, target, {
+      1: Outcomes.HasNoName,
+    }),
+  ]);
+});
+
+test("evaluate() fails a input element with trimmed aria label", async (t) => {
```
Jym77, line 124:
```suggestion
test("evaluate() fails a input element with empty aria-label", async (t) => {
```

packages/alfa-rules/test/sia-r8/rule.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R8, { document }), [
+    failed(R8, target, {
+      1: Outcomes.HasNoName,
+    }),
+  ]);
+});
+
+test("evaluate() fails a select element with div, whose has no text", async (t) => {
```
Jym77, line 136:
```suggestion
test("evaluate() fails a select element with aria-labelledby pointing to an element with no text", async (t) => {
```

packages/alfa-rules/test/sia-r8/rule.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+
+test("evaluate() is inapplicable for an input element with aria-hidden", async (t) => {
+  const target = <input disabled aria-hidden="true" aria-label="firstname" />;
+
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R8, { document }), [inapplicable(R8)]);
+});
+
+test("evaluate() is inapplicable for a select element because it's disabled ", async (t) => {
```
Jym77, line 180:
```suggestion
test("evaluate() is inapplicable for a disabled element", async (t) => {
```

packages/alfa-rules/test/sia-r8/rule.spec.tsx
```diff
@@ -0,0 +1,200 @@
...
+      <option value="opel">Opel</option>
+    </select>
+  );
+
+  const document = Document.of([target]);
+
+  t.deepEqual(await evaluate(R8, { document }), [inapplicable(R8)]);
+});
+
+test("evaluate() is inapplicable for an input element with style element set to none ", async (t) => {
```
Jym77, line 194:
```suggestion
test("evaluate() is inapplicable for an input element which is not displayed", async (t) => {
```

Jym77 (changes requested):

PR #820, opened 2021-05-31
Experience with Inventory

Jym77:
Trying to see what we can do to support https://siteimprove-wgs.atlassian.net/wiki/spaces/CE/pages/2533720392/Feature+idea+Accessibility+Inventory

Adding a new kind of outcome and rule, "inventory" + some examples of what inventory rules could look like.

Code would need to be desentangled from the existing one and moved out of `alfa-act`.
Plus consider questions in inventory (could be used with site-wide answers, as a light-weight Policy).


Jym77:
Closing in favour of #904.

PR #827, opened 2021-06-08
Consider element with `height` or `width` of 0 as invisible unless the text overflows

### LINKED ISSUE #774: Elements whose height is 0 should be considered invisible even with scroll/auto overflow
```html
<style>
  div {
    overflow-y: scroll;
    height: 0px;
  }
</style>

<div>Foo</div>
```
This is currently considered visible because the height check is perform only on `overflow: hidden` in `isClippedBySize`.
But this is actually invisible since the scrollbar itself has height 0 and therefore is not visible…

elenamongelli:
Resolves #774


packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -64,6 +64,52 @@ function isClippedBySize(
...
+          switch (dimension.type) {
+            case "percentage":
+              if (dimension.value <= 0) {
+                return true;
+              } else {
+                break;
+              }
+
+            case "length":
+              if (dimension.value <= 1) {
```
Jym77, line 82: If this case, we need <=0. If the height/width is 1 pixel big, it is visible (not much, but still…)
```suggestion
              if (dimension.value <= 0) {
```

packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -64,6 +64,52 @@ function isClippedBySize(
...
+                return true;
+              } else {
+                break;
+              }
+          }
+        }
+      }
```
Jym77, line 112: We can probably merge both cases and avoid duplicating the rest of the logic:
`if (x.value === "auto" || x.value === "scroll" || y.value === "auto" || y.value === "scroll")`

packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -64,6 +64,52 @@ function isClippedBySize(
...
+
+      if (x.value === "scroll" || y.value === "scroll") {
+        const { value: height } = style.computed("height");
+        const { value: width } = style.computed("width");
```
Jym77, line 93: Also, since `height` and `width` are used in several places (also in the `"hidden"` case), we can refactor the definitions out of the `if`.

packages/alfa-rules/test/common/predicate/is-visible.spec.tsx
```diff
@@ -61,6 +61,37 @@ test(`isVisible() returns false when an element is hidden by reducing its size
   t.equal(isVisible(element), false);
 });
 
+test(`isVisible() returns false when an element is hidden by reducing its size
+      to 1x1 pixels and clipping overflow`, (t) => {
+  const element = (
+    <div style={{ width: "0px", height: "0px", overflow: "hidden" }}>
+      Hello World
+    </div>
+  );
```
Jym77, line 70: That one is indeed not needed since it is already covered by the previous tests.
```suggestion
```

packages/alfa-rules/test/common/predicate/is-visible.spec.tsx
```diff
@@ -61,6 +61,37 @@ test(`isVisible() returns false when an element is hidden by reducing its size
...
+
+  t.equal(isVisible(element), false);
+});
+
+test(`isVisible() returns false when an element is scroll and its size is 0x0 px`, (t) => {
```
Jym77, line 75:
```suggestion
test(`isVisible() returns false when an element scrolls its overflow and its size is reduced to 0 pixel, thus hidding the scrollbar`, (t) => {
```
(similar for the `auto` case)

packages/alfa-rules/test/common/predicate/is-visible.spec.tsx
```diff
@@ -61,6 +61,37 @@ test(`isVisible() returns false when an element is hidden by reducing its size
...
+  const element = (
+    <div style={{ width: "0px", height: "0px", overflow: "scroll" }}>
```
Jym77, line 77: That should probably be split in several cases for each dimension just like https://github.com/Siteimprove/alfa/blob/develop/packages/alfa-rules/test/common/predicate/is-visible.spec.tsx#L34-L51 to show that individual dimensions of 0 are enough (it's not about having **both** height and width at 0)

(same for the `auto` case)

packages/alfa-rules/test/common/predicate/is-visible.spec.tsx
```diff
@@ -382,3 +413,13 @@ test(`isVisible() returns false for a text node with a parent element with
...
+      Hello World
+    </div>
+  );
+
+  t.equal(isVisible(element), false);
+});
```
Jym77, line 425:
```suggestion
```
Looks like a duplicate.

Jym77 (changes requested):

Jym77 (dismissed):

kasperisager:
@Jym77 or @elenamongelli: Could one of you update the title to be a little more descriptive of what the changes involve?

kasperisager (approved):

PR #831, opened 2021-06-10
Accessible name: reject whitespace descendants

### LINKED ISSUE #830: `Name.fromDescendant()` should reject empty spaces names

Jym77:
Resolves #830

### DIFF
--- a/packages/alfa-aria/src/name.ts
+++ b/packages/alfa-aria/src/name.ts
@@ -742,7 +742,7 @@ export namespace Name {
         )
       );
 
-    const name = flatten(names.map((name) => name.value).join(" "));
+    const name = flatten(names.map((name) => name.value).join(" ")).trim();
 
     if (name === "") {
       return None;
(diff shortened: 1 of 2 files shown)


kasperisager:
Have we tested how this behaves with HTML like below?

```html
<button>Hello<span> </span>world</button>
```

Jym77:
That gives the name `Hello world`, but I guess that thanks to the `join(" ")` in the middle.
Chrome gives it a name `Helloworld` 😞  Firefox agrees with `Hello world`…

Current code does the same 😄

PR #835, opened 2021-06-15
Implement SIA-R55

Jym77:


packages/alfa-rules/src/sia-r55/rule.ts
```diff
@@ -0,0 +1,99 @@
...
+      },
+
+      expectations(target) {
+        // Empty groups have been filtered out already, so we can safely get the
+        // first element
+        const role = Node.from(Iterable.first(target).get(), device).role.get()
+          .name;
+
+        const sameResource = Question.of(
+          "reference-equivalent-resources",
```
Jym77, line 65: I'm not fully sure I should reuse that question URI here 🤔

kasperisager: Nope, we should use a new one appropriate for this case. Something along the lines of `is-content-equivalent` maybe?

kasperisager (dismissed):

kasperisager (approved):
👍

kasperisager:
Oh, there are actually some unused imports in the rule: https://github.com/Siteimprove/alfa/security/code-scanning

PR #838, opened 2021-06-21
Pass roles as part of diagnostics in SIA-R55 and SIA-R56

Jym77:
We actually want to send the role as part of the diagnostic to avoid frontend trying to recomputing it…

### DIFF
--- a/packages/alfa-rules/src/sia-r55/rule.ts
+++ b/packages/alfa-rules/src/sia-r55/rule.ts
@@ -1,7 +1,9 @@
 import { Diagnostic, Rule } from "@siteimprove/alfa-act";
 import { Node, Role } from "@siteimprove/alfa-aria";
+import { Array } from "@siteimprove/alfa-array";
 import { Element, Namespace } from "@siteimprove/alfa-dom";
 import { Iterable } from "@siteimprove/alfa-iterable";
+import { List } from "@siteimprove/alfa-list";
 import { Predicate } from "@siteimprove/alfa-predicate";
 import { Err, Ok } from "@siteimprove/alfa-result";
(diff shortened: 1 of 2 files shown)


packages/alfa-rules/src/sia-r55/rule.ts
```diff
@@ -82,15 +82,63 @@ export default Rule.Atomic.of<Page, Group<Element>, Question>({
...
-      Diagnostic.of(
-        `Some \`${role}\` have the same name and different content.`
+      WithRole.of(
+        `Some \`${role}\` have the same name and different content.`,
+        role
       )
     );
 }
+
+class WithRole extends Diagnostic {
```
Jym77, line 100: Not the most inspired name for a class 🙈 Couldn't find a better idea…

packages/alfa-rules/src/sia-r56/rule.ts
```diff
@@ -66,7 +66,7 @@ export default Rule.Atomic.of<Page, Group<Element>>({
 
 export namespace Outcomes {
   export const differentNames = (role: Role.Name) =>
-    Ok.of(Diagnostic.of(`No two \`${role}\` have the same name.`));
+    Ok.of(SameNames.of(`No two \`${role}\` have the same name.`, role, []));
```
Jym77, line 69: I think it's just easier to keep the same structure in both case, even if the passed ones are always going to have an empty list of errors.

packages/alfa-rules/src/sia-r55/rule.ts
kasperisager: I like it! 😄

kasperisager (approved):

PR #839, opened 2021-06-21
Implement SIA-R70

### LINKED ISSUE #826: Implement SIA-R70
https://github.com/Siteimprove/sanshikan/blob/df450e27db11c4d7b36eb1d53fa0aecdb0f729ba/rules/sia-r70.md

Applicability:
* take inspiration from, e.g., R59 or R87 (both do the same thing in different ways…)

Expectation:
* Use `descendants()`, flat tree is obtained with options `flattened` and `nested`. See, e.g. Applicability of R2.
* use `isVisible` and `not(isIgnored)`
* Get inspiration from R1 (expectation) for grabbing elements by their name. `hasName` can take several arguments.

Extended diagnostic:
* We need to send the list of offending elements we've found.

elenamongelli:
Closes #826


packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,147 @@
+/// <reference lib="dom" />
+
```
Jym77, line 2:
```suggestion
```

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,147 @@
...
+import { List } from "@siteimprove/alfa-list";
+import { Iterable } from "@siteimprove/alfa-iterable";
+
+import { expectation } from "../common/expectation";
+import { isIgnored, hasChild } from "../common/predicate";
+import { Group } from "../common/group";
+
+import { isRendered } from "../common/predicate/is-rendered";
+
+import { isDocumentElement } from "../common/predicate/is-document-element";
```
Jym77, line 17: These can be imported together with the other predicates.

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,147 @@
...
+
+const { isElement, hasName, hasNamespace } = Element;
+const { and, or, test, not } = Predicate;
+const deprecated = [
```
Jym77, line 21:
```suggestion

const deprecated = [
```

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,147 @@
...
+  "noframes",
+  "plaintext",
+  "rb",
+  "rtc",
+  "shadow",
+  "spacer",
+  "strike",
+  "tt",
+  "xmp",
+];
```
Jym77, line 49:
```suggestion
];

```

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,147 @@
...
+export default Rule.Atomic.of<Page, Document, Group<Element>>({
```
Jym77, line 50:
```suggestion
export default Rule.Atomic.of<Page, Document>({
```
The third type parameter is for questions and shouldn't be needed here.

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,147 @@
...
+          .filter(
+            and(
+              hasNamespace(Namespace.HTML),
+              hasName("acronym", ...deprecated),
+              isRendered(device)
+            )
+          );
+
+        //console.dir(deprecatedElements.toJSON(), { depth: null });
+
```
Jym77, line 71:
```suggestion
```

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,147 @@
...
+        `The document contains deprecated elements: `,
+        errors
+      )
+    );
+}
+
+class DeprecatedElements extends Diagnostic implements Iterable<Element> {
+  public static of(
+    message: string,
+    errors: Iterable<Element> = []
```
Jym77, line 101:
```suggestion
    errors: Iterable<Element> = List.empty()
```
Since we're using `List`, we can as well initialise with one.

packages/alfa-rules/test/sia-r70/rule.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const document = h.document([target]);
+
+  t.deepEqual(await evaluate(R70, { document }), [
+    passed(R70, document, {
+      1: Outcomes.IsNotDeprecated,
+    }),
+  ]);
+});
+
+test("evaluate() passes a page with deprecated non visible element not into accessibility tree", async (t) => {
```
Jym77, line 25:
```suggestion
test("evaluate() passes a page with a deprecated but not rendered element", async (t) => {
```

packages/alfa-rules/test/sia-r70/rule.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const document = h.document([target]);
+
+  t.deepEqual(await evaluate(R70, { document }), [
+    passed(R70, document, {
+      1: Outcomes.IsNotDeprecated,
+    }),
+  ]);
+});
+
+test("evaluate() fails a page with deprecated visible element in accessibility tree", async (t) => {
```
Jym77, line 43:
```suggestion
test("evaluate() fails a page with deprecated and rendered element", async (t) => {
```

packages/alfa-rules/test/sia-r70/rule.spec.tsx
```diff
@@ -0,0 +1,107 @@
...
+  const document = h.document([target]);
+
+  t.deepEqual(await evaluate(R70, { document }), [
+    failed(R70, document, {
+      1: Outcomes.IsDeprecated([menuitem1, menuitem2]),
+    }),
+  ]);
+});
+
+test("evaluate() is inapplicable to text nodes in disabled groups", async (t) => {
```
Jym77, line 98:
```suggestion
test("evaluate() is inapplicable to non-HTML documents", async (t) => {
```

Jym77 (dismissed):

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+import { Document, Element, Namespace } from "@siteimprove/alfa-dom";
+import { Predicate } from "@siteimprove/alfa-predicate";
+import { Err, Ok } from "@siteimprove/alfa-result";
+import { Page } from "@siteimprove/alfa-web";
+import { List } from "@siteimprove/alfa-list";
+import { Iterable } from "@siteimprove/alfa-iterable";
+
+import { expectation } from "../common/expectation";
+import { hasChild, isRendered, isDocumentElement } from "../common/predicate";
+import { Group } from "../common/group";
```
Jym77, line 11:
```suggestion
```

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,141 @@
+import { Rule, Diagnostic } from "@siteimprove/alfa-act";
+import { Document, Element, Namespace } from "@siteimprove/alfa-dom";
+import { Predicate } from "@siteimprove/alfa-predicate";
+import { Err, Ok } from "@siteimprove/alfa-result";
+import { Page } from "@siteimprove/alfa-web";
+import { List } from "@siteimprove/alfa-list";
+import { Iterable } from "@siteimprove/alfa-iterable";
```
Jym77, line 7:
```suggestion
import { Rule, Diagnostic } from "@siteimprove/alfa-act";
import { Document, Element, Namespace } from "@siteimprove/alfa-dom";
import { Iterable } from "@siteimprove/alfa-iterable";
import { List } from "@siteimprove/alfa-list";
import { Predicate } from "@siteimprove/alfa-predicate";
import { Err, Ok } from "@siteimprove/alfa-result";
import { Page } from "@siteimprove/alfa-web";
```

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+          ),
+        };
+      },
+    };
+  },
+});
+
+export namespace Outcomes {
+  export const IsNotDeprecated = Ok.of(
+    Diagnostic.of(`The document doesn't contain deprecated elements`)
```
Jym77, line 80:
```suggestion
    Diagnostic.of(`The document doesn't contain any deprecated element`)
```

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+    };
+  }
+}
+
+namespace DeprecatedElements {
+  export interface JSON extends Diagnostic.JSON {
+    errors: List.JSON<Element>;
+  }
+
+  export function areDeprecatedElements(
```
Jym77, line 136:
```suggestion
  export function isDeprecatedElements(
```
Even if there may be several deprecated element, this function checks whether a single object is an instance of the class `DeprecatedElements`.

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+            () => Outcomes.IsDeprecated(deprecatedElements)
+          ),
+        };
+      },
+    };
+  },
+});
+
+export namespace Outcomes {
+  export const IsNotDeprecated = Ok.of(
```
Jym77, line 79: Let's change the name to `HasNoDeprecatedElement`.
It's not about the target (the page) being or not deprecated, but having or not deprecated elements.

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,141 @@
...
+    Diagnostic.of(`The document doesn't contain deprecated elements`)
+  );
+
+  export const IsDeprecated = (errors: Iterable<Element>) =>
```
Jym77, line 83: And that should become `HasDeprecatedElements`.

Jym77 (changes requested):

Jym77:
Also, the rule needs to be activated in `packages/alfa-rules/src/rules.ts`. See #850

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,140 @@
...
+    );
+  }
+
+  public toJSON(): DeprecatedElements.JSON {
+    return {
+      ...super.toJSON(),
+      errors: this._errors.toJSON(),
+    };
+  }
+}
```
kasperisager, line 128: The class should expose a getter for the `_errors` property.

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,140 @@
...
+      },
+
+      expectations(target) {
+        const deprecatedElements = target
+          .descendants({ flattened: true, nested: true })
+          .filter(isElement)
+          .filter(
+            and(
+              hasNamespace(Namespace.HTML),
+              hasName("acronym", ...deprecated),
```
kasperisager, line 60: Is there a reason for `"acronym"` being spelled out on its own?

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,140 @@
...
+  "plaintext",
+  "rb",
+  "rtc",
+  "shadow",
+  "spacer",
+  "strike",
+  "tt",
+  "xmp",
+];
+
```
kasperisager, line 44: Pretty composition is pretty powerful:

```suggestion
/**
 * {@link https://developer.mozilla.org/en-US/docs/Web/HTML/Element#obsolete_and_deprecated_elements}
 */
const isDeprecated = hasName(
  "acronym",
  "applet",
  "basefont",
  "bgsound",
  "big",
  "blink",
  "center",
  "content",
  "dir",
  "font",
  "frame",
  "frameset",
  "hgroup",
  "image",
  "keygen",
  "marquee",
  "menuitem",
  "nobr",
  "noembed",
  "noframes",
  "plaintext",
  "rb",
  "rtc",
  "shadow",
  "spacer",
  "strike",
  "tt",
  "xmp",
);

```

kasperisager (changes requested):

packages/alfa-rules/src/sia-r70/rule.ts
elenamongelli: hasName (https://github.com/Siteimprove/alfa/blob/01db7c61462ef8bdd0badcb2405119088acd463d/packages/alfa-dom/src/node/element/predicate/has-name.ts#L20-L22) asks for `string `and `array<String>`

kasperisager: Ah, `deprecated` should be asserted `const` to fix that:

```ts
const deprecated = [
  "acronym",
  ...
] as const;
```

`hasName()` indeed requires _at least 1_ name, so TypeScript needs to know that the list of names has more than one value. A `const` assertion will make TypeScript infer the type of the array as a fixed-size tuple `[string, string, ...]` rather than `string[]`.

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,138 @@
...
+    return {
+      ...super.toJSON(),
+      errors: Array.toJSON(this._errors),
+    };
+  }
+
+  public get errors(): ReadonlyArray<Element> {
+    return this._errors;
+  }
+}
```
kasperisager, line 126:
```suggestion
class DeprecatedElements extends Diagnostic implements Iterable<Element> {
  public static of(
    message: string,
    errors: Iterable<Element> = []
  ): DeprecatedElements {
    return new DeprecatedElements(message, Array.from(errors));
  }
  
  private _errors: ReadonlyArray<Element>;

  private constructor(message: string, errors: Array<Element>) {
    super(message);
    this._errors = errors;
  }
  
  public get errors(): ReadonlyArray<Element> {
    return this._errors;
  }

  public equals(value: DeprecatedElements): boolean;

  public equals(value: unknown): value is this;

  public equals(value: unknown): boolean {
    return (
      value instanceof DeprecatedElements &&
      value._message === this._message &&
      Array.equals(value._errors, this._errors)
    );
  }
  
  public *[Symbol.iterator](): Iterator<Element> {
    yield* this._errors;
  }

  public toJSON(): DeprecatedElements.JSON {
    return {
      ...super.toJSON(),
      errors: this._errors.map((element) => element.path()),
    };
  }
}
```

Some reorganisation to align with the rest of the classes in Alfa. I also changed the way elements are represented in JSON. Now, they're represented by their path, which is easier for people to grok than their structure.

packages/alfa-rules/src/sia-r70/rule.ts
```diff
@@ -0,0 +1,138 @@
...
+
+namespace DeprecatedElements {
+  export interface JSON extends Diagnostic.JSON {
+    errors: List.JSON<Element>;
```
kasperisager, line 130:
```suggestion
    errors: Array<string>;
```

kasperisager (changes requested):

packages/alfa-rules/src/sia-r70/rule.ts
Jym77: > I also changed the way elements are represented in JSON. Now, they're represented by their path, which is easier for people to grok than their structure.

https://siteimprove.slack.com/archives/G0E9EH2AJ/p1623060925002600
Our main consumer seems to prefer element to their path 😀

kasperisager: That relates to the structure of the _class instance_, not its JSON representation. The JSON representation of elements on its own is essentially useless since two structurally equivalent elements sitting different places in the document really aren't the same.

Jym77:
PR got botched during rebase (?). Closing in favour of #870.

PR #840, opened 2021-06-24
Add `archive` option in `Scraper#scrape()`

kasperisager:
This pull request adds a new option `archive` in the `Scraper#scrape()` method that currently supports capturing an [MHTML](https://en.wikipedia.org/wiki/MHTML) archive of the page being scraped, similar to how screenshots can be captured with the `screenshot` option. It also adds two new flags to the CLI: `--archive` and `--archive-format`.

Archives can be a useful addition to screenshots for verifying the state of the page when scraped. It also provides a useful tool for performing additional manual inspection of the page.

### DIFF
--- a/packages/alfa-cli/bin/alfa/command/scrape/flags.ts
+++ b/packages/alfa-cli/bin/alfa/command/scrape/flags.ts
@@ -144,14 +144,14 @@ export const Flags = {
 
   screenshot: Flag.string(
     "screenshot",
-    "The path to write a screenshot to. If not provided, no screenshot is taken."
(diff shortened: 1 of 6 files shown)


Jym77 (approved):
Should we also have an option to **read** from an archive?
That will definitely be needed if we want to have an open list of archives and test toward it regularly.

But maybe topic for later…

kasperisager:
That's already possible by virtue of the scraper being based on Chromium:

```sh
alfa scrape --archive example.com.mht https://example.com
alfa scrape example.com.mht
```

PR #843, opened 2021-06-25
SIA-R83: jump to offset parent when searching for clipping ancestor

### LINKED ISSUE #824: Absolutely positioned elements are able to escape their clipping ancestor
```html
<style>
  .clipping {
    overflow: hidden;
    height: 28px;
    border: red solid;
    margin: 20px;
  }
  .absolute {
    position: absolute;
  }
</style>

<div class="clipping">Lorem ipsum…</div>

<div class="clipping">
  <div class="absolute">Lorem ipsum…</div>
</div>
```

R83 flags both text, but the absolute positioning on the second actually let it escape the clipping.

clipped

As far as I understand, the absolutely positioned element has a 0-sized box in the normal flow (and that box is inside the clipping ancestor), and then another box, in another stacking context, is created for it (and placed where it should). The box which actually has content is not part of the same stacking context as the clipping box, so it can overflow it freely…

elenamongelli:
Resolves #824 
Adds tests


packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -60,11 +61,13 @@ export default Rule.Atomic.of<Page, Text>({
...
 
       expectations(target) {
+        const parent = target.parent();
+
         return {
           1: expectation(
-            wrapsText(device)(target),
-            () => Outcomes.WrapsText,
-            () => Outcomes.ClipsText
+            parent.every(and(isElement, not(wrapsText(device)))),
```
Jym77, line 68:
```suggestion
            target.parent({ flattened: true, nested: true }).every(and(isElement, not(wrapsText(device)))),
```
Since `parent` is only used once, no need to create a variable for it. Plus, we should go up with the same options (flattened, nested) that we went down in the applicability.

Jym77: And I'm actually thinking we messed up the quantifier here 😕 
If there is no parent (which is not really possible due to the applicability, but whatever), the rule should pass (nothing clips). But `parent.every` is `true` on `None` which ends up returning the error…
So it should be `parent.some` instead (I think we messed up when we moved the negations around and forgot to negate the quantifier).

packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -134,29 +137,55 @@ function isPossiblyClipped(device: Device): Predicate<Element> {
...
+    return relevantParent.every(wrapsText(device));
+  };
+}
+
+function isPositioned(device: Device, position: string): Predicate<Element> {
+  return (element) =>
+    Style.from(element, device).computed("position").value.value === position;
+}
+
+const isActuallyClipping = (element: Element, device: Device) => {
```
Jym77, line 159: I think this is actually checking that the element is actually **not** clipping and should be renamed `isNotActuallyClipping`. (if there is an `overflow` of `clip`, then it returns `false`…)
(or we keep the name and negate the returns, and use a wrapping `not()` at call site)

Jym77 (dismissed):

kasperisager:
@elenamongelli and @Jym77: Could either of you adjust the title to reflect the actual change made? The changes fortunately don't involve adjusting the way positioning works in CSS, which is what the title currently reads like 😄

packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -134,29 +139,55 @@ function isPossiblyClipped(device: Device): Predicate<Element> {
...
+      : offsetParent(element, device);
+
+    return relevantParent.every(wrapsText(device));
+  };
+}
+
+function isPositioned(device: Device, position: string): Predicate<Element> {
+  return (element) =>
+    Style.from(element, device).computed("position").value.value === position;
+}
```
Jym77, line 159: We just got a brand new `hasComputedStyle` 😀 (https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/common/predicate/has-computed-style.ts)
```suggestion
function isPositioned(device: Device, position: string): Predicate<Element> {
  return hasComputedStyle("position", value => value.value === position, device);
}
```
should do the trick.

packages/alfa-rules/test/sia-r83/rule.spec.tsx
```diff
@@ -126,6 +153,36 @@ test(`evaluate() fails a text node that clips overflow and sets a fixed height
   ]);
 });
 
+test("evaluate() fails a text node thet is relatively positioned with a clipping absolutely positioned ancestor ", async (t) => {
```
Jym77, line 156:
```suggestion
test("evaluate() fails a text node that is non-statically positioned with a clipping offset parent", async (t) => {
```

packages/alfa-rules/test/sia-r83/rule.spec.tsx
```diff
@@ -49,7 +49,34 @@ test(`evaluate() passes a child text node of an element whose parent truncates
       ]),
     ]
   );
+  test("evaluate() passes a text node that is absolutely positioned with a clipping ancestor", async (t) => {
```
Jym77, line 52:
```suggestion
  test("evaluate() passes a text node that is non-statically positioned with a clipping ancestor which is not the offset parent", async (t) => {
```

Jym77 (changes requested):

packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -60,11 +61,15 @@ export default Rule.Atomic.of<Page, Text>({
       },
 
       expectations(target) {
+        const parent = target.parent();
+
```
kasperisager, line 65: This is unused.

packages/alfa-rules/src/sia-r83/rule.ts
kasperisager: Also, `isActuallyClipping` should be a `function` declaration.

packages/alfa-rules/test/sia-r83/rule.spec.tsx
```diff
@@ -49,7 +49,34 @@ test(`evaluate() passes a child text node of an element whose parent truncates
       ]),
     ]
   );
```
kasperisager, line 51:
```suggestion
  );

```

kasperisager (changes requested):

packages/alfa-rules/src/sia-r83/rule.ts
Jym77: These haven't been addressed.
Should be 
```javascript
function isNotActuallyClipping(element: Element, device: Device): boolean {
```
Or maybe, since it's only used once and negated:
```javascript
function isActuallyClipping(element: Element, device: Device): boolean {
```
plus negating the `return` + removing the `!` when calling it.

Jym77 (dismissed):

Jym77 (commented):
Looks good 👍

kasperisager (approved):

PR #846, opened 2021-06-29
Aria attributes depending on element

### LINKED ISSUE #799: ARIA attributes that do not depend on HTML attributes are not set
As far as I can tell, in `feature.ts` we set the ARIA attributes that are defined in [attribute mappings](https://w3c.github.io/html-aam/#html-attribute-state-and-property-mappings) (as being mapped from HTML attributes).
However, the AAM is also setting some ARIA attributes as part of [role mappings](https://w3c.github.io/html-aam/#html-element-role-mappings) for those who do not depend directly on HTML attributes…

Quick look gives the following list of HTML elements that need to be considered:
- [ ] `<datalist>`
- [ ] `<input type="checkbox">` (we already have something around `aria-checked` not fully sure how this changes that or not)
- [ ] `<input type="radio">` (same comment)
- [X] `<input>` with type text, search, telephone, URL or email
- [x] `<li>`
- [X] `<option>`
- [x] `<textarea>`

Some of these might actually be tricky to build…

Jym77:
Resolves #799 

* Ignoring `<datalist>` because 
    1. it's hard and costly to find if there is a correct `<input list="id">`
    2. it's never rendered (I think), so always ignored 🤔 
    3. let's keep it simple until it proves problematic…
* Ignoring checkboxes and radio buttons because:
    1. the bit around the `checkedness` is mostly covered by the `checked` attribute (https://html.spec.whatwg.org/multipage/input.html#dom-input-checked) not fully, since it is the IDL attribute, but should be good enough.
    2. the bit around the `indeterminate` IDL attribute looks like something we just can't handle currently…

### DIFF
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -13,10 +13,10 @@ import { Cell, Table } from "@siteimprove/alfa-table";
 import { Attribute } from "./attribute";
 import { Name } from "./name";
 import { Role } from "./role";
-import and = Refinement.and;
 
 const { hasInputType, hasName, isElement } = Element;
 const { or, test } = Predicate;
+const { and } = Refinement;
 
 /**
  * @internal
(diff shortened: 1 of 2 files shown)


kasperisager (approved):
👏

PR #848, opened 2021-07-01
Make descendents of fully clipped nodes invisible

### LINKED ISSUE #731: `isClippedBySize()` does not account for cases where the grandparent of a text node fully indents and clips text

Jym77:
Resolves #731 

Plus a bit a refactor/streamlining of `isVisible` and `isClipped`.

### DIFF
--- a/packages/alfa-rules/src/common/predicate/has-computed-style.ts
+++ b/packages/alfa-rules/src/common/predicate/has-computed-style.ts
@@ -1,15 +1,28 @@
 import { Device } from "@siteimprove/alfa-device";
-import { Element } from "@siteimprove/alfa-dom";
+import { Element, Text } from "@siteimprove/alfa-dom";
 import { Predicate } from "@siteimprove/alfa-predicate";
 import { Context } from "@siteimprove/alfa-selector";
 import { Property, Style } from "@siteimprove/alfa-style";
 
+const { isElement } = Element;
+
 export function hasComputedStyle<N extends Property.Name>(
-  name: N,
-  predicate: Predicate<Style.Computed<N>>,
   device: Device,
   context?: Context
-): Predicate<Element> {
-  return (element) =>
-    Style.from(element, device, context).computed(name).some(predicate);
+): (
+  name: N,
+  predicate: Predicate<Style.Computed<N>>
+) => Predicate<Element | Text> {
+  return function hasComputedStyle(
+    name: N,
+    predicate: Predicate<Style.Computed<N>>
+  ): Predicate<Element | Text> {
+    return (node) =>
+      isElement(node)
+        ? Style.from(node, device, context).computed(name).some(predicate)
+        : node
+            .parent({ flattened: true })
+            .filter(isElement)
+            .some(hasComputedStyle(name, predicate));
+  };
 }
--- a/packages/alfa-rules/src/common/predicate/has-computed-style.ts
+++ b/packages/alfa-rules/src/common/predicate/has-computed-style.ts
@@ -1,0 +1,15 @@
+import { Device } from "@siteimprove/alfa-device";
+import { Element } from "@siteimprove/alfa-dom";
+import { Predicate } from "@siteimprove/alfa-predicate";
+import { Context } from "@siteimprove/alfa-selector";
+import { Property, Style } from "@siteimprove/alfa-style";
+
+export function hasComputedStyle<N extends Property.Name>(
+  name: N,
+  predicate: Predicate<Style.Computed<N>>,
+  device: Device,
+  context?: Context
+): Predicate<Element> {
+  return (element) =>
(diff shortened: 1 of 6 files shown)


packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -23,132 +27,141 @@ export function isClipped(
...
+                })
+                .some(isClipped(device, context))
           ),
           node
         )
       );
 }
 
+/**
+ * @internal
```
kasperisager, line 54:
```suggestion
```

Release tags should only be used for package exported API.

packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -23,132 +27,141 @@ export function isClipped(
...
+}
 
-          if (indent.value < 0 || whitespace.value === "nowrap") {
-            switch (indent.type) {
-              case "percentage":
-                if (abs(indent.value) >= 1) {
-                  return true;
-                }
+/**
+ * @internal
```
kasperisager, line 106:
```suggestion
```

packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -1,83 +1,69 @@
...
-                .computed("font-size")
-                .some((size) => size.value === 0)
-            )
-        ) {
-          return false;
-        }
-      }
+/**
+ * @internal
+ */
```
kasperisager, line 31:
```suggestion
```

packages/alfa-rules/src/common/predicate/has-computed-style.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const { isElement } = Element;
+
+export function hasComputedStyle<N extends Property.Name>(
+  device: Device,
+  context?: Context
+): (
+  name: N,
+  predicate: Predicate<Style.Computed<N>>
+) => Predicate<Element | Text> {
```
kasperisager, line 15:
```suggestion
export function hasComputedStyle<N extends Property.Name>(
  name: N, 
  predicate: Predicate<Style.Computed<N>>,
  device: Device,
  context?: Context
): Predicate<Element | Text> {
```

Is there a compelling reason to optimise for a partially applied function? If not, let's not wrap things in an additional closure.

packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -23,132 +27,141 @@ export function isClipped(
...
-              }
-
-            case "length":
-              if (dimension.value <= 0) {
-                return true;
-              } else {
-                break;
-              }
-          }
+          // technically, 1×1 elements are (possibly) visible since they
```
kasperisager, line 87:
```suggestion
          // Technically, 1×1 elements are (possibly) visible since they
```

packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -1,83 +1,69 @@
...
+    isClipped(device, context),
+    isOffscreen(device, context),
+    // Empty text
+    and(isText, (text) => text.data.trim() === ""),
+    // Text of size 0
+    and(
+      isText,
+      hasStyle("font-size", (size) => size.value === 0)
+    ),
+    // Element with visibility ≠ "visible"
```
kasperisager, line 50:
```suggestion
    // Element with visibility != "visible"
```

kasperisager (changes requested):

packages/alfa-rules/src/common/predicate/has-computed-style.ts
Jym77: Yep, wasn't sure about that one… My first version was uncurryfied (https://github.com/Siteimprove/alfa/pull/848/commits/0b65f47453948011a9489a2cf5f124536fd27a73) but then I thought the call site would be much better if we can partially apply the common stuff (https://github.com/Siteimprove/alfa/pull/848/files#diff-cddeee40baf90b48e519a73729f0571a0544d5e7b80fa7fabf6350e2b8702f49R51-R54) but then again, this does result in an ugly partial application due to eta-expansion of type variables (https://github.com/Siteimprove/alfa/pull/848/files#diff-cddeee40baf90b48e519a73729f0571a0544d5e7b80fa7fabf6350e2b8702f49R33-R36)

So I wasn't sure about it and I guess I'll just uncurryfy it again 😁

kasperisager (approved):
Looks great! 👏

PR #851, opened 2021-07-05
SIA-R75: Only report elements whose font size affect some text

### LINKED ISSUE #810: SIA-R75 should ignore elements with no text depending on their font size
```html
<div style="font-size: 1px">
  <span style="font-size: 16px">Hello World</span>
</div>
```

Alfa currently fails the `<div>` but shouldn't since it doesn't have text content that depends on its `font-size`…

- [x] Depends on Siteimprove/sanshikan#176
- [ ] Follow up on DEV-11898

elenamongelli:
Resolves #810


packages/alfa-rules/src/sia-r75/rule.ts
```diff
@@ -46,13 +47,32 @@ export default Rule.Atomic.of<Page, Element>({
...
-        );
+        const texts = target
+          .descendants({
+            flattened: true,
+            nested: true,
+          })
+          .filter(isText)
+          .map((text) => text.parent().filter(isElement))
+          .filter((option) => option.isSome())
+          .map((option) => option.get())
```
kasperisager, line 58: This sequence of operations can be replaced by `Sequence#collect()`:

```suggestion
          .collect((text) => text.parent().filter(isElement))
```

packages/alfa-rules/test/sia-r75/test.html
```diff
@@ -0,0 +1,2 @@
+<small p id="output">Hello world<small>
+ <10px></10px>
```
kasperisager, line 2: This should be included as an accompanying unit test.

kasperisager (dismissed):

Jym77 (approved):

kasperisager (approved):

PR #853, opened 2021-07-06
Refactor `Hash` and add additional methods

kasperisager:


Jym77 (approved):

kasperisager:
Changing to minor as I really consider the breaking changes bug fixes more than anything.

PR #858, opened 2021-07-09
Accept arbitrarily deeply nested children in `jsx()`

kasperisager:
This was actually already accepted, but the types weren't accounting for it and so neither was the code 🙈


Jym77 (approved):

kasperisager:
TypeScript is acting up so I've got some investigation to do before requesting another review.

Jym77 (approved):
To infinity and beyond!

PR #859, opened 2021-07-09
Fix naming from subtree and implicitly referenced `<label>` elements

### LINKED ISSUE #808: `Node.from()` is thrown off guard by `aria-hidden` around implicit labels
…yeah, that is a very precise case…

This test is currently passing:
```javascript
test(`.from() is thrown off guard by aria-hidden around implicit label`, (t) => {
  const input = <input />;
  const label = <label>Foo {input}</label>;
  const div = <div aria-hidden="false">{label}</div>;

  h.document([div]);

  t.deepEqual(Node.from(input, device).toJSON(), {
    type: "element",
    node: "/div[1]/label[1]/input[1]",
    role: "textbox",
    name: null,
    attributes: [{ name: "aria-checked", value: "false" }],
    children: [],
  });
});
```
The name of the input is wrong.
`Name.from(input)`, however, returns the correct value (unless cache stuff happens).

```html
<div aria-hidden="false">
  <label>
  Foo
  <input />
  </label>
</div>
```

The code is thrown off guard at https://github.com/Siteimprove/alfa/blob/main/packages/alfa-aria/src/node.ts#L512-L530 where `aria-hidden` is a global attribute making the `<div>` into an element (with no role) rather than a container. Not 100% sure about what happens next, the differences seems to be the call to `Name.from(node)` at https://github.com/Siteimprove/alfa/blob/main/packages/alfa-aria/src/node.ts#L541 which descend into the subtree too early and somehow ends up giving no name to the `<input>`…

Somehow, this doesn't affects explicit labels (`<label for="…">`) or other cases I've tried…

- [ ] Follow up to DEV-11744

kasperisager:
A quantifier was incorrect and a referrer was missing when computing names of implicit `<label>` references.

Resolves #808

### DIFF
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -124,19 +124,18 @@ const nameFromAttribute = (element: Element, ...attributes: Array<string>) => {
   return None;
 };
 
-const nameFromChild = (predicate: Predicate<Element>) => (
-  element: Element,
-  device: Device,
-  state: Name.State
-) => {
-  for (const child of element.children().filter(isElement).find(predicate)) {
-    return Name.fromDescendants(child, device, state.visit(child)).map((name) =>
-      Name.of(name.value, [Name.Source.descendant(element, name)])
-    );
-  }
-
-  return None;
-};
+const nameFromChild =
+  (predicate: Predicate<Element>) =>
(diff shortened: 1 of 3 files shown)


Jym77 (approved):
Well done!

PR #860, opened 2021-07-09
SIA-R62: accept decoration on ancestors and descendants of links.

### LINKED ISSUE #777: SIA-R62 should accept decoration on the parent or child of the link
Currently, failing both these links:
```html
<style>
  a {
    text-decoration: none;
  }
  span {
    text-decoration: underline;
  }
</style>

<p>
  Lorem <a href="#"><span>Foo</span></a> Ipsum <span><a href="#">Bar</a></span>
</p>
```
The ACT rule draft (and Sanshikan) accepts both.
https://github.com/act-rules/act-rules.github.io/blob/4b64bba6cb77a8d4dc0649c83c55372f513d979f/_rules/inline-link-in-paragraph-is-distinguishable-be4d0c.md#expectation

We can probably have a good approximation by looking at the properties on the (parent of the) text node(s) inside the link. Might get a bit complex if there are more than one text node, with different properties…

Similarly, we might want to look at style on (parents of) text nodes inside the paragraph since we currently pass
```html
<p style="font-weight: 400">
  <span style="font-weight: 700">Lorem</span>
  <a href="#" style="font-weight: 700; text-decoration: none">Ipsum</a>
</p>
```
because the `<a>` has a different style than the `<p>`… But every text inside the `<p>` actually has the same style as the `<a> 😞 

- [ ] Follow up on DEV-11653.
- [ ] Follow up on DEV-11636.
- [ ] Follow up on DEV-11724.
- [ ] Follow up on DEV-11987.

Jym77:
Resolves #777

Plus make `takeWhile` and `takeLastWhile` accept refinements and narrow the iterable type.


kasperisager (dismissed):

packages/alfa-rules/src/sia-r62/rule.ts
```diff
@@ -1,14 +1,20 @@
 import { Rule, Diagnostic } from "@siteimprove/alfa-act";
+import { Array } from "@siteimprove/alfa-array";
```
kasperisager, line 2:
```suggestion
```

kasperisager (dismissed):

kasperisager (approved):
👌

PR #862, opened 2021-07-13
SIA R66 and R69: return `cantTell` when encountering a text shadow

### LINKED ISSUE #743: SIA-R66 and SIA-R69 do not account for `text-shadow`
The following case currently fails both rules:

```html
<!DOCTYPE html>

<style>
  p {
    color: white;
    text-shadow: 1px 1px 2px black, -1px 1px 2px black, 1px -1px 2px black,
      -1px -1px 2px black;
  }
</style>

<p>Hello world</p>
```

The result is, however, perfectly readable with white text on a black background:

Rendered output of the HTML snippet

A common use for this seems to be text overlaid on top of images in which case the text shadow ensures proper contrast regardless of where the text is positioned.

- [ ] Follow up on DEV-11367.

Jym77:
Resolves #743 
Plus add support for `text-shadow`.

### DIFF
--- a/packages/alfa-style/src/property/text-shadow.ts
+++ b/packages/alfa-style/src/property/text-shadow.ts
@@ -1,0 +1,109 @@
+import {
+  Color,
+  Current,
+  Keyword,
+  Length,
+  Percentage,
+  RGB,
+  System,
+  Token,
+} from "@siteimprove/alfa-css";
+import { Option } from "@siteimprove/alfa-option";
+import { Parser } from "@siteimprove/alfa-parser";
+import { Slice } from "@siteimprove/alfa-slice";
+
+import { Property } from "../property";
+import { Resolver } from "../resolver";
+import { Tuple } from "./value/tuple";
+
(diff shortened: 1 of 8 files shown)


packages/alfa-style/src/property/text-shadow.ts
```diff
@@ -0,0 +1,190 @@
...
+  > extends Value.JSON<"text-shadow"> {
+    color: Option.JSON<C>;
+    offset: Tuple.JSON<[Length<U>, Length<U>]>;
+    blur: Length.JSON<U>;
+  }
+
+  export function isTextShadow(value: unknown): value is TextShadow {
+    return value instanceof TextShadow;
+  }
+}
```
kasperisager, line 190: Is there a reason not to use the existing `Shadow` class from `@siteimprove/alfa-css`?

kasperisager (changes requested):

kasperisager (dismissed):

PR #865, opened 2021-07-14
Refactor collection types and associates

kasperisager:
This pull request is a refactor of the collection types and some associated types, such as `Monad<T>`, `Result<T, E>`, and `Option<T>`. Notably, the way functionality that depends on details of `T` is implemented has changed from functions with argument constraints to methods with `this` constraints, such as sorting:

```ts
interface Collection<T> {
  sort<T extends Comparable<T>>(this: Collection<T>): Collection<T>
}

class Foo implements Comparable<Foo> {
  compare(other: Foo): Comparison {
    // ...
  }
}

const collection: Collection<Foo>;

collection.sort();
```

### DIFF
--- a/packages/alfa-list/src/list.ts
+++ b/packages/alfa-list/src/list.ts
@@ -90,6 +90,10 @@ export class List<T> implements Collection.Indexed<T> {
     );
   }
 
+  public flatten<T>(this: List<List<T>>): List<T> {
+    return this.flatMap((list) => list);
+  }
+
   public reduce<U>(reducer: Reducer<T, U, [index: number]>, accumulator: U): U {
     return Iterable.reduce(this, reducer, accumulator);
   }
@@ -395,11 +399,22 @@ export class List<T> implements Collection.Indexed<T> {
     return Iterable.join(this, separator);
   }
 
+  public sort<T extends Comparable<T>>(this: List<T>): List<T> {
+    return this.sortWith(compareComparable);
+  }
+
   public sortWith(comparer: Comparer<T>): List<T> {
     return List.from(Iterable.sortWith(this, comparer));
   }
 
-  public compareWith(iterable: Iterable<T>, comparer: Comparer<T>): Comparison {
+  public compare(this: List<Comparable<T>>, iterable: Iterable<T>): Comparison {
+    return this.compareWith(iterable, compareComparable);
+  }
+
+  public compareWith<U = T>(
+    iterable: Iterable<U>,
+    comparer: Comparer<T, U, [index: number]>
+  ): Comparison {
     return Iterable.compareWith(this, iterable, comparer);
   }
 
@@ -686,16 +701,5 @@ export namespace List {
       (list, value) => list.append(value),
       List.empty()
     );
-  }
-
-  export function sort<T extends Comparable<T>>(list: List<T>): List<T> {
-    return list.sortWith(compareComparable);
-  }
-
-  export function compare<T extends Comparable<T>>(
-    a: List<T>,
-    b: Iterable<T>
-  ): Comparison {
-    return a.compareWith(b, compareComparable);
   }
 }
--- a/packages/alfa-list/src/list.ts
+++ b/packages/alfa-list/src/list.ts
@@ -83,6 +83,10 @@ export class List<T> implements Collection.Indexed<T> {
     );
   }
 
+  public apply<U>(mapper: List<Mapper<T, U>>): List<U> {
+    return mapper.flatMap((mapper) => this.map(mapper));
+  }
+
   public flatMap<U>(mapper: Mapper<T, List<U>, [index: number]>): List<U> {
     return this.reduce(
       (list, value, index) => list.concat(mapper(value, index)),
@@ -112,10 +116,6 @@ export class List<T> implements Collection.Indexed<T> {
     accumulator: U
   ): U {
     return Iterable.reduceUntil(this, predicate, reducer, accumulator);
-  }
-
-  public apply<U>(mapper: List<Mapper<T, U>>): List<U> {
-    return this.flatMap((value) => mapper.map((mapper) => mapper(value)));
   }
 
   public filter<U extends T>(
--- a/packages/alfa-map/src/map.ts
+++ b/packages/alfa-map/src/map.ts
@@ -56,10 +56,32 @@ export class Map<K, V> implements Collection.Keyed<K, V> {
     return new Map(this._root.map(mapper), this._size);
   }
 
+  /**
+   * Apply a map of functions to each corresponding value of this map.
+   *
+   * @remarks
+   * Keys without a corresponding function or value are dropped from the
+   * resulting map.
+   *
+   * @example
+   * ```ts
+   * Map.of(["a", 1], ["b", 2])
+   *   .apply(Map.of(["a", (x) => x + 1], ["b", (x) => x * 2]))
+   *   .toArray();
+   * // => [["a", 2], ["b", 4]]
+   * ```
+   */
   public apply<U>(mapper: Map<K, Mapper<V, U>>): Map<K, U> {
-    return mapper.flatMap((mapper) => this.map(mapper));
+    return this.collect((value, key) =>
+      mapper.get(key).map((mapper) => mapper(value))
+    );
   }
 
+  /**
+   * @remarks
+   * As the order of maps is undefined, it is also undefined which keys are
+   * kept when duplicate keys are encountered.
+   */
   public flatMap<L, U>(mapper: Mapper<V, Map<L, U>, [key: K]>): Map<L, U> {
     return this.reduce(
       (map, value, key) => map.concat(mapper(value, key)),
@@ -67,6 +89,11 @@ export class Map<K, V> implements Collection.Keyed<K, V> {
     );
   }
 
+  /**
+   * @remarks
+   * As the order of maps is undefined, it is also undefined which keys are
+   * kept when duplicate keys are encountered.
+   */
   public flatten<K, V>(this: Map<K, Map<K, V>>): Map<K, V> {
     return this.flatMap((map) => map);
   }
@@ -145,7 +172,7 @@ export class Map<K, V> implements Collection.Keyed<K, V> {
   /**
    * @remarks
    * As the order of maps is undefined, it is also undefined which keys are
-   * deleted when duplicate values are encountered.
+   * kept when duplicate values are encountered.
    */
   public distinct(): Map<K, V> {
     let seen = Map.empty<V, V>();
--- a/packages/alfa-map/src/map.ts
+++ b/packages/alfa-map/src/map.ts
@@ -48,22 +48,30 @@ export class Map<K, V> implements Collection.Keyed<K, V> {
     return this._size === 0;
   }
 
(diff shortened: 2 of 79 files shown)


packages/alfa-list/src/list.ts
```diff
@@ -90,6 +90,10 @@ export class List<T> implements Collection.Indexed<T> {
     );
   }
 
+  public flatten<T>(this: List<List<T>>): List<T> {
+    return this.flatMap((list) => list);
+  }
```
Jym77, line 95: Could it be 
```suggestion
  public flatten<T>(this: List<Iterable<T>>): List<T> {
    return this.flatMap((iterable) => List.from(iterable));
  }
```

packages/alfa-result/src/err.ts
```diff
@@ -36,30 +36,34 @@ export class Err<E> implements Result<never, E> {
...
   }
 
-  public reduce<U>(reducer: unknown, accumulator: U): U {
-    return accumulator;
+  public flatMap(): Err<E> {
+    return this;
   }
 
-  public apply(): this {
+  public flatten<T, E>(this: Err<E>): Result<T, E> {
```
Jym77, line 59: Should we also have a `flattenErr` doing the symmetric thing? (flatten `Err<Result<T, E>>` and keep `Ok<T>` intact)

packages/alfa-result/src/result.ts
```diff
@@ -32,9 +32,10 @@ export interface Result<T, E = T>
   map<U>(mapper: Mapper<T, U>): Result<U, E>;
   mapErr<F>(mapper: Mapper<E, F>): Result<T, F>;
   mapOrElse<U>(ok: Mapper<T, U>, err: Mapper<E, U>): U;
+  apply<U>(mapper: Result<Mapper<T, U>, E>): Result<U, E>;
   flatMap<U>(mapper: Mapper<T, Result<U, E>>): Result<U, E>;
+  flatten<T, E>(this: Result<Result<T, E>, E>): Result<T, E>;
```
Jym77, line 37: Could be
```suggestion
  flatten<T, E, F>(this: Result<Result<T, E>, F>): Result<T, E | F>;
```
to be a bit more generic.

Also, it is a bit confusing to re-use the names `T` and `E` which are already the global type parameter. All other signatures are using `U` or `F` when they need new parameters.

Jym77 (commented):
Nice trick 😃

packages/alfa-list/src/list.ts
kasperisager: That breaks the signature of a monadic join: `F<F<T>> => F<T>`.

packages/alfa-result/src/result.ts
kasperisager: "Fresh" type parameters are needed for properly typing `this` and so the names should align with the ones used for the class itself.

I'm not too keen on further complicating the generics for no current benefit so let's keep it simple 🙈

packages/alfa-list/src/list.ts
kasperisager: I should actually add the method to the `Monad<T>` interface.

packages/alfa-result/src/err.ts
kasperisager: Errors really shouldn't contain `Ok` values 😬

packages/alfa-map/src/map.ts
```diff
@@ -48,59 +48,63 @@ export class Map<K, V> implements Collection.Keyed<K, V> {
...
 
-  public map<U>(mapper: Mapper<V, U, [K]>): Map<K, U> {
+  public map<U>(mapper: Mapper<V, U, [key: K]>): Map<K, U> {
     return new Map(this._root.map(mapper), this._size);
   }
 
-  public flatMap<L, U>(mapper: Mapper<V, Map<L, U>, [K]>): Map<L, U> {
+  public apply<U>(mapper: Map<K, Mapper<V, U>>): Map<K, U> {
+    return mapper.flatMap((mapper) => this.map(mapper));
+  }
```
Jym77, line 61: I'm not sure I understand what it means to apply a map to another map.
Can we add some tests of that?

Trying
```javascript
const map = Map.of(["foo", 1], ["bar", 2], ["baz", 3], ["qux", 4]);
  const fmap = Map.of(
    ["a", (x: number) => x + 1],
    ["b", (x: number) => x + 2],
    ["c", (x: number) => x + 3],
    ["d", (x: number) => x + 4]
  );
```
gives `[ [ 'baz', 7 ], [ 'qux', 8 ], [ 'foo', 5 ], [ 'bar', 6 ] ]` (so only the fourth function was applied?)
But with
```javascript
const fmap = Map.of(
    ["foo", (x: number) => x + 1],
    ["bar", (x: number) => x + 2],
    ["baz", (x: number) => x + 3],
    ["qux", (x: number) => x + 4]
  );
```
I get `[ [ 'baz', 5 ], [ 'qux', 6 ], [ 'foo', 3 ], [ 'bar', 4 ] ]` (so it's another function that get applied).
And I'm not sure I understand the differences nor the impact that similar keys has (or doesn't have).


---
Intuitively, I would have said that if I get a M={ (k, v), k in K and v in V } and a F={ (k, f), k in K and f in V -> U} then applying F to M should give { (k, f(v)) / (k,v) in M and (k, f(v)) in F} but that is clearly not what running the tests do…

---
What I guess happens is that each f is applied to each v; then the resulting f(v) is associated with the key that was associated to v in the first place; so for a given (k, v) we now have several (k, f₁(v)), (k, f₂(v)), …; and then only one of those "wins" and stays in the resulting map but the key collision resolution depends on the keys of the f in a somewhat incontrollable way 🤔

packages/alfa-map/src/map.ts
```diff
@@ -48,59 +48,63 @@ export class Map<K, V> implements Collection.Keyed<K, V> {
...
+
+  public flatMap<L, U>(mapper: Mapper<V, Map<L, U>, [key: K]>): Map<L, U> {
     return this.reduce(
       (map, value, key) => map.concat(mapper(value, key)),
       Map.empty<L, U>()
     );
   }
 
-  public reduce<R>(reducer: Reducer<V, R, [K]>, accumulator: R): R {
+  public flatten<K, V>(this: Map<K, Map<K, V>>): Map<K, V> {
```
Jym77, line 70: `flatMap` and `flatten` would also benefit from some tests. Notably to see what happens with key collisions.

packages/alfa-selective/src/selective.ts
```diff
@@ -32,7 +35,13 @@ export class Selective<S, T = never>
   }
 
   public map<U>(mapper: Mapper<T, U>): Selective<S, U> {
-    return this.flatMap((value) => new Selective(Right.of(mapper(value))));
+    return this.flatMap(
+      (value) => new Selective<S, U>(Right.of(mapper(value)))
+    );
+  }
+
+  public apply<U>(mapper: Selective<S, Mapper<T, U>>): Selective<S, U> {
```
Jym77, line 43: That's also a type where `apply`, `flatMap`, `flatten` could use some tests.

Jym77 (changes requested):

packages/alfa-map/src/map.ts
kasperisager: Agreed! I've copied the remarks from `Map#distinct()` to both methods:

> As the order of maps is undefined, it is also undefined which keys are deleted when duplicate values are encountered.

packages/alfa-map/src/map.ts
kasperisager: Usually, `#apply()` will give you the Cartesian product between a collection of functions and a collection of values. For non-keyed types, this makes sense:

```
[1, 2].apply([x => x + 1, x => x * 2]) == [1 + 1, 2 + 1, 1 * 2, 2 * 2]
```

For keyed types like a map, however, I guess it doesn't really sense 🤔 Let's try:

```
[["a", 1], ["b", 2]].apply([["f", x => x + 1], ["g", x => x * 2]]) == [["a", 1 + 1], ["b", 2 + 1], ["a", 1 * 2], ["b", 2 * 2]]
```

Unless we're dealing with a multi map, this will result in keys colliding and the final result will either be `[["a", 1 + 1], ["b", 2 + 1]]` _or_ `[["a", 1 * 2], ["b", 2 * 2]]`.

Can we think of a more appropriate implementation of `#apply()` for keyed types?

kasperisager: One option is performing pairwise application similar to Haskell's [`ZipList`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Control-Applicative.html#t:ZipList):

```js
[["a", 1], ["b", 2]].apply([["a", x => x + 1], ["b", x => x * 2]]) == [["a", 1 + 1], ["b", 2 * 2]]
```

Jym77: Yes, the cartesian product makes sense for non-keyed collections. But for keyed collection we somehow need to combine keys or handle collisions in a somewhat intuitive way.

---
Getting a bit away from the code and the typing, on the mathematical side I would see 1 "natural" way of doing cartesian product of keyed values (i.e sets of pairs):
The pure cartesian product of `[ ["a", 1], ["b", 2]] × [[X, f], [Y, g]]` would be `[ ["a", 1, X, f], ["a", 1, Y, g], ["b", 2, X, f], ["b", 2, Y, g] ]`. Now, we are in a case where we want `f` and `g` to be functions, so we know how to combine them and can get `[ ["a", f(1), X], ["a", g(1), Y], ["b", f(2), X], ["b", g(2), Y] ]`. But from there, we need a way to somehow combine the keys, these could be:
1. Drop the `X/Y` key, only keep the `"a"/"b"` one and rely on some key collision resolution. This is the current implementation. I feel (as a dev) the collision resolution is too obscur for me to rely on it (as in, I'll never use this cartesian product so maybe we should just not have it).
2. Drop the `"a"/"b"` key and keep the `X/Y` one. Same problem.
3. Use an implicit but clear collision solving, typically only keep values where both keys are "equal".
4. Require an explicit way to combine keys, that is an extra parameter of type `comb: (K, K) => L` and then build `[ [comb("a", X), f(1)], [comb("a", Y), g(1)], [comb("b", X), f(2)], [comb("b", Y), g(2)] ]`. Collisions may still happen but are now the problem of the caller. Might be better to have `comb: (K, K) => Option<L>` and we drop the combinations that return `None` (so that 3. is just `comb = (k1, k2) => Equatable.equals(k1, k2) ? Some.of(k1) : None`, and `comb` might be optional defaulting to this). This may cause problem with typing, however…
5. Require that the keys of the second map are of type `K => L` (or `K => Option<L>`) so we can combine keys the same way we combine values: `[ [X("a"), f(1)], [Y("a"), g(1)], [X("b"), f(2)], [Y("b"), g(2)] ]`. That sounds like asking a bit too much and I'm not sure what would the use case be… (plus maybe also problematic type-wise). Again, possible collisions are now the problem of the caller.

I'm not sure which one makes more sense… It also depends on what use case we envision for that. I do feel that case 3. is the most natural cartesian product of keyed values (cartesian product of values whose key match). I like the added flexibility of 4. if it's possible…

Jym77 (approved):

Jym77 (approved):

PR #867, opened 2021-07-15
Add `context` property to `Question` class

kasperisager:


packages/alfa-rules/src/sia-r87/rule.ts
```diff
@@ -73,20 +73,23 @@ export default Rule.Atomic.of<Page, Document, Question>({
           "first-tabbable-reference-is-main",
           "boolean",
           target,
+          target,
```
Jym77, line 76: This question (and others in this rule) should know be able to be asked about the first tabbable rather than the document. But that is maybe a topic for another PR 🤔

packages/alfa-act/src/rule.ts
```diff
@@ -24,22 +24,23 @@ const { flatMap, flatten, reduce } = Iterable;
 /**
  * @public
  */
-export abstract class Rule<I = unknown, T = unknown, Q = never>
+export abstract class Rule<I = unknown, T = unknown, Q = never, S = never>
```
Jym77, line 27: It feels like `S` should default to `T` since this is the main case we have. Would simplify things quite a bit in removing the need to specify it all over the place.

packages/alfa-rules/test/common/outcome.ts
```diff
@@ -3,8 +3,8 @@ import { Record } from "@siteimprove/alfa-record";
 import { Result } from "@siteimprove/alfa-result";
 import { Page } from "@siteimprove/alfa-web";
 
-export function passed<T, Q>(
-  rule: Rule<Page, T, Q>,
+export function passed<T, Q, S>(
```
Jym77, line 6: Here also, `S` could default to `T`.

packages/alfa-act/src/question.ts
```diff
@@ -9,38 +9,49 @@ import * as json from "@siteimprove/alfa-json";
 /**
  * @public
  */
-export class Question<Q, S, A, T = A>
+export class Question<Q, S, C, A, T = A>
```
Jym77, line 12: Looks like in most case `C` and `S` are the same. Can't really default here due to parameter order, however 😞 Might be worth reordering parameters to avoid copying stuff around at call sites.

packages/alfa-act/src/question.ts
```diff
@@ -9,38 +9,49 @@ import * as json from "@siteimprove/alfa-json";
...
     Monad<T>,
-    Serializable<Question.JSON<Q, S>>
+    Serializable<Question.JSON<Q, S, C>>
 {
-  public static of<Q, A, S>(
+  public static of<Q, S, C, A>(
     uri: string,
     type: Q,
     subject: S,
+    context: C,
```
Jym77, line 23: Same here, main use case seems to be asking questions whose subject is the same as the context. Defaulting (or overloading) would make call sites cleaner.

packages/alfa-act/src/interview.ts
```diff
@@ -9,34 +9,35 @@ import { Rule } from "./rule";
...
+ * provides `D - 1 = Depths[D]`.
  */
 type Depths = [-1, 0, 1, 2];
 
 /**
  * @public
  */
-export type Interview<Q, S, T, D extends number = 3> =
-  | T
+export type Interview<Q, S, C, A, D extends number = 3> =
```
Jym77, line 21: Same remark here, `C` and `S` are usually the same, but we can't default without swapping parameter order 😞

Jym77 (changes requested):

packages/alfa-rules/src/sia-r87/rule.ts
kasperisager: Agreed! That's another, and slightly more involved, breaking change so let's deal with that in another pull request.

packages/alfa-rules/test/common/outcome.ts
kasperisager: The type parameters are here always inferred based on use so there's no need for defaults.

kasperisager:
While we can neither default the context type `C` to `S` in the general `Question` class, nor do constructor overloading to default `C` to `S`, we _can_ overload the common `Question.of()` constructor function used in `@siteimprove/alfa-rules`. That does lead to some really nasty type assertions though, as we have to convince the compiler that `C` and `S` are the same type when one and not the other overload is used:

```ts
export function of<Q extends keyof Question, S>(
  uri: string,
  type: Q,
  subject: S,
  message: string
): act.Question<Q, S, S, Question[Q]>;

export function of<Q extends keyof Question, S, C>(
  uri: string,
  type: Q,
  subject: S,
  context: C,
  message: string
): act.Question<Q, S, C, Question[Q]>;

export function of<Q extends keyof Question, S, C>(
  uri: string,
  type: Q,
  subject: S,
  contextOrMessage: C | string,
  message?: string
): act.Question<Q, S, C, Question[Q]> {
  let context: C;

  if (message === undefined) {
    context = subject as unknown as C;
    message = contextOrMessage as string;
  } else {
    context = contextOrMessage as C;
  }

  return act.Question.of(uri, type, subject, context, message);
}
```

I'm not totally convinced that's a worthwhile tradeoff 🤔

Jym77:
> While we can neither default the context type `C` to `S` in the general `Question` class, nor do constructor overloading to default `C` to `S`, we _can_ overload the common `Question.of()` constructor function used in `@siteimprove/alfa-rules`. That does lead to some really nasty type assertions though, as we have to convince the compiler that `C` and `S` are the same type when one and not the other overload is used:

> I'm not totally convinced that's a worthwhile tradeoff 🤔

Fair point, this is indeed a bit ugly 🤔 

It does however cleans the call sites a lot (essentially, if https://github.com/Siteimprove/alfa/pull/867#discussion_r689521004 is also done, then most of the individual rules don't need to be changed at all). The current multiplication of `Question.of("has-focus-indicator", "boolean", target, target, "Does the element have a visible focus indicator?")` like patterns with duplicated `target` in the middle is a bit annoying, I feel…

A cleaner solution might be to change parameters order (moving `message` early, since it is always required) and then `context` can easily be made optional, but it is a wider change (plus `message` is incidental to to the type, while the other parameters are the real meat, so it would maybe give a wrong feeling…)

---
I would argue to do the overload anyway, with the rational of "hide the ugly stuff deep to keep the tip of the iceberg clean". In that case, I'd argue that it's more important to make rule writing (and reading) simple by reducing the boiler plate needed at call site. 

---
Actually, we might need a different wrapper in `alfa-rules` instead 🤔 
As far as I understard, the context of questions is (almost) supposed to be the rule's target (while the subject of questions can be anything). This kind of means that any `Question.of(…)` will (almost) always use `target` as its context (and then the subject needs to be given) (this is not true for questions inside Applicability where no target exist).

Would it be possible to have a wrapper that can only used inside `expectations` and uses `target` as its context? 🤔 I don't really think it is 😢 

My point here is: Questions always need to provide their subject, since it can be anything, but most of the time (iirc all of our current use cases), the context is either not needed (same as subject), or inferable (the test target). Can we make use of that?

Feels weird to explain to a new dev "when asking questions, we need to provide a _subject_ (the thing about which we ask the question), and a _context_ (which is always the test target)"…

kasperisager:
I'm really thinking that Scala 3 is onto something with "givens" (https://docs.scala-lang.org/scala3/reference/contextual/givens.html#alias-givens). For lack of that, I'd be in favour of reordering the arguments to make overloading more natural. I really dislike argument juggling as it makes for confusing and brittle code 😕 

With regards to this...

> when asking questions, we need to provide a subject (the thing about which we ask the question), and a context (which is always the test target)

...I rather like the fact that the we're now very explicit about the context and subject of questions. As we've seen for several rules, trying to treat them as one and the same is just confusing and means there are things we simply cannot express. Previously, we couldn't ask questions about things that were related to the test target but weren't themselves the test target, such as each section of repeated content (the subject) within a document (the test target and question context).

Jym77 (approved):

PR #875, opened 2021-07-23
SIA-R62: accept different background image

### LINKED ISSUE #782: SIA-R62 fails to account for background images not used as background images
Feeling _déjà vu_? Me too…
```html
<style>
  a {
    text-decoration: none;

    background-image: linear-gradient(to right, #046B99 50%, transparent 50%);
    background-repeat: repeat-x;
    background-position: center bottom 4%;
    background-size: 2px 1px;
    background-origin: padding-box;
  }
</style>

<p>Hello world <a href="#">Foo</a></p>
```
🙈

We might be good with accepting any difference in `background` as good enough.

- [ ] Follow up on DEV-11733.

Jym77:
Resolves #782 

Still need to send something for `background` in the extended diagnostic, so that frontend can display the links as Alfa sees them.
Serialising `background` is a mess due do layers. Of course, browsers disagree…
* https://github.com/w3c/csswg-drafts/issues/418
* https://github.com/servo/servo/blob/0d0cfd030347ab0711b3c0607a9ee07ffe7124cf/components/style/properties/shorthands/background.mako.rs#L133-L193
* https://github.com/chromium/chromium/blob/d7da0240cae77824d1eda25745c4022757499131/third_party/blink/renderer/core/css/style_property_serializer.cc#L923-L1071

The (simpler) Servo way is likely more than enough for us.

We may also consider having a `toString` method directly on `Shorthand`. This is the only rule that uses them, but it could become a convenient thing in diagnostics in general. But architecture of `alfa-styles` doesn't seem to match that very well 🤔


kasperisager (dismissed):

Jym77:
Added serialisation of `background` + tests of it…

kasperisager (approved):
Looks fantastic 👏

PR #879, opened 2021-08-04
Test `white-space` property of text parents in SIA-R83

### LINKED ISSUE #747: SIA-R83 should look at the `white-space` property of the parent element of text nodes
SIA-R83 currently looks at the `white-space` property of the clipping ancestor of text nodes, but this property can change between the clipping ancestor and the text node. The rule therefore fails a case like the following:

```html
<style>
  p {
    overflow: hidden;
    white-space: nowrap;
  }

  span {
    white-space: normal;
  }
</style>

<p>
  <span>
    Lorem ipsum dolor sit amet, consectetur adipiscing elit. Fusce venenatis
    augue non placerat accumsan. Sed nulla libero, hendrerit a mi quis, placerat
    tempus dolor. Nunc id ligula nunc. Sed luctus fringilla tellus ut finibus.
    Pellentesque eget imperdiet ipsum, non vestibulum nunc. Nullam nec sem a
    velit ultrices dapibus. Maecenas quis mauris venenatis ipsum facilisis
    efficitur in ac arcu.
  </span>
</p>
```

- [ ] Follow up on DEV-11401.

kasperisager:
Resolves #747


packages/alfa-rules/src/common/predicate/has-non-wrapped-text.ts
```diff
@@ -0,0 +1,42 @@
...
+            )
+          ),
+          node
+        ) ||
+        node
+          .children({
+            flattened: true,
+          })
+          .filter(or(isElement, isText))
+          .some(hasNonWrappedText)
```
Jym77, line 39: Side tracking… 🤔 
We do have quite a bunch of predicates that check some atomic stuff for a node and (recursively) either its ancestors (the full `isVisible` family) or its descendant (as here).
Wondering whether we could have a wrapper in `alfa-dom` like `someAncestor(predicate: Predicate<Node>): Predicate<Node>)` (or `everyAncestor`, `someInclusiveDescendant`, …) Might get tricky to handle caching on these.
Not sure it's worth the effort… Definitely not a topic for this PR.

Jym77 (approved):
👌 I feel the rule is getting even more understandable that way 🎉

packages/alfa-rules/src/common/predicate/has-non-wrapped-text.ts
kasperisager: I feel that such a method mixes two concerns; ancestor traversal and predicate application on a sequence. To me, it seems like the composition `.ancestors(traversalOptions).<some | every | all>(predicate)` is favourable to `.<some | every | all>Ancestors(predicate, traversalOptions)`, especially since the latter requires three additional methods for something that can already be achieved with the four existing methods.

Jym77: You are perfectly right 😄 
Then I guess the question is why we have so many recursive predicates instead of using `.ancestors().some(predicate)` 😅

kasperisager: Caching 🙈

Jym77: 🤔 
```javascript
const cache = Cache.empty<Node, boolean>();

function foo(node: Node): boolean {
  return cache.get(node, () => !isElement(node) || node.id.isSome()
}

function bar(node: Node): boolean {
  return foo.inclusiveAncestors().some(foo);
}
```
should handle the cache and separate the actual test from the DOM tree walk 🤔 
`foo` could even be inlined if its test doesn't make sense in isolation (e.g. `isRendered` doesn't make sense if not testing ancestors).

kasperisager: That only caches the result of each invocation of `foo()`, but you still potentially walk far up the inclusive ancestor chain every time `bar()` is called. We want to cache the entire recursive computation, making each step along the inclusive ancestor chain reusable.

PR #894, opened 2021-08-20
Add `@siteimprove/alfa-tuple` package

kasperisager:


packages/alfa-tuple/src/tuple.ts
```diff
@@ -0,0 +1,390 @@
...
+
+  export type Size<T extends Tuple> = T["length"];
+
+  export function size<T extends Tuple>(tuple: T): Size<T> {
+    return Array.size(tuple);
+  }
+
+  export type Copy<T extends Tuple> = T extends readonly [...infer T]
+    ? T
+    : Empty;
```
Jym77, line 41: Isn't that the same thing as `T`?

Also, I feel it would make sense inside this namespace to use `Tuple<T>` instead of arrays to keep intention clearer:
```typescript
export type Copy<T extends Tuple> = T extends Tuple<[...infer T]>
    ? Tuple<T>
    : Empty;
```
should do the same 🤔 (same remark in various places). (notably if we ever change implementation so that `Tuple<T>` is not anymore a synonym for `T`, these notation wouldn't need to change)

Jym77 (changes requested):
Also thinking this might better as a class so we can chain the methods.
With the possibilities to constrain `this`, it should work out…

packages/alfa-tuple/src/tuple.ts
kasperisager: It's _usually_ the same as `T`, but will remove any `readonly` modifiers. That's also why `Tuple<T>` won't work as that re-introduces the `readonly` modifiers. Here's how it's intended to work:

```ts
const tuple: readonly [number, null, boolean] = [1, null, true];
const copy: [number, null, boolean] = copy(tuple);
```

kasperisager:
Wrapping tuples in an additional object has the major drawback of introducing a comparatively large overhead as tuples are meant to encode small, heterogeneous lists. It would also mean that we couldn't use tuple literal syntax, which currently means that `[1, null, true]` and `Tuple.of(1, null, true)` are the same thing and both will pass for `Tuple<[number, null, object]>`.

Jym77:
> Wrapping tuples in an additional object has the major drawback of introducing a comparatively large overhead as tuples are meant to encode small, heterogeneous lists. It would also mean that we couldn't use tuple literal syntax, which currently means that `[1, null, true]` and `Tuple.of(1, null, true)` are the same thing and both will pass for `Tuple<[number, null, object]>`.

Then I'm not fully sure of the value we are adding 🤔 
`Tuple.foo` just seems to be another name for `Array.foo` that only applies to read only arrays. Writing `Tuple.of(1, null, true)` definitely conveys more intention than `[1, null, true] as readonly` but it looks like that's the only benefit, and I'm not sure it's worth this much added code 🤔

kasperisager:
Arrays are homogenous while tuples are heterogeneous. Passing a tuple to `Array` helper functions will result in an array, but passing it to `Tuple` helpers functions will preserve it as a tuple. That's the main difference between `@siteimprove/alfa-array` and `@siteimprove/alfa-tuple`. For example:

```ts
const tuple = Tuple.of(1, {}, "foo");

Array.map(tuple, x => x.toString());
// Array<string>

Tuple.map(tuple, x => x.toString());
// [string, string, string]
```

kasperisager:
More fun stuff:

```ts
const tuple = Tuple.of(1, {}, "foo");

Tuple.reject(tuple, isNumber);
// [object, string]

Tuple.get(tuple, 2);
// Option<string>

Tuple.set(tuple, 2, true);
// [number, object, boolean]
```

Jym77 (approved):

PR #898, opened 2021-08-23
Don't count invisible text as possibly wrapped in SIA-R83

kasperisager:


kasperisager:
Merging a releasing a patch right away.

PR #900, opened 2021-08-25
SIA-R62: Add superscript and subscript as distinguishing features for links

### LINKED ISSUE #823: SIA-R62 should accept superscript and subscript as distinguishing features
I'm wondering what the team thinks of the following use case for distinguishable links in paragraph text. I'm leaning towards the treatment of a `<sup>` or `<sub>` tag being sufficient to distinguish the link but open to having a conversation. Also more generally wondering if positioning of the link should be considered as a distinguishing characteristic. That might be too variable and more suitable for a manual confirmation but curious of your thoughts.

Example: A link as a superscript footnote where underline might make this illegible. 
`<p>This is important paragraph text that requires a footnote. <sup><a href="#footnote">[1]</a><sup><p>`

elenamongelli:
Resolves #823


Jym77 (changes requested):
`vertical-align` needs to be sent in the Diagnostic, the same way that the other longhands (`color` and `font-weight`) are: https://github.com/Siteimprove/alfa/blob/b1b6b3b32b56a73d0af867969b85aed3f96fbf01/packages/alfa-rules/src/sia-r62/rule.ts#L448-L449

kasperisager (dismissed):
This could use a test case or two and then the UA style sheet also needs to be updated to add `vertical-align` to `<sup>` and `<sub>` elements: https://github.com/Siteimprove/alfa/issues/823#issuecomment-904426992

kasperisager (approved):

PR #904, opened 2021-08-31
Experiences with Inventory

Jym77:
Trying to see what we can do to support https://siteimprove-wgs.atlassian.net/wiki/spaces/CE/pages/2533720392/Feature+idea+Accessibility+Inventory

Adding a new kind of outcome and rule, "inventory" + some examples of what inventory rules could look like.


kasperisager (changes requested):
I'm still not quite convinced that stretching the interpretation of the ACT format to also encompass the Siteimprove specific concept of Inventory is ideal 🤔 ACT and Inventory are very distinct things in my book. The way we have to split the packages and concepts to make room for this new type of Inventory rule is a bit of a red flag in that regard.

It does seem like a completely separate abstraction would be a much better fit. First thought: A method of tagging content, such as elements, with structured metadata and then a way of collecting and organising all the tagged content to form the final Inventory.

Jym77:
> It does seem like a completely separate abstraction would be a much better fit. First thought: A method of tagging content, such as elements, with structured metadata and then a way of collecting and organising all the tagged content to form the final Inventory.

🤔 So the `dom.Node` would have a `metadata` free form field that can then be gathered however we want.
However, in that case, the metadata is only the accessible name. It seems a bit artificial to add a storage for it (since we already cache it in the accessibility tree and all that). I'm not foreseeing the "inventory" concept to have a lot of complex computation (other than selecting some nodes and some already computed properties on them)…

I do agree that the Applicability/Expectation split makes little sense for inventory. So just a "gather" thingie that go through the DOM to select some nodes and grab associated stuff. Then we need to some how export that in the global object sent back by Alfa 👍

kasperisager:
> 🤔 So the `dom.Node` would have a `metadata` free form field that can then be gathered however we want.

Not quite. The metadata would rather be associated with arbitrary objects from the outside through a `WeakMap`. I'm thinking that a "metadata collection flow" could be performed along the lines of:

```ts
interface Metadata {
  accessibleName: string;
}

const inventory = Inventory.empty<Element, Metadata>();

document
  .descendants()
  .filter(Element.isElement)
  .forEach((element) => {
    for (const name of ...) {
      inventory.add(element, name);
    }
  })
```

In the above, `Inventory` is really just a glorified `WeakMap` that could have additional methods for organising and exporting a structured representation of the collected data.

PR #907, opened 2021-09-02
Handle interposed backgrounds in `getForeground()` and `getBackground()`

### LINKED ISSUE #758: SIA-R66 and SIA-R69 do not account for backgrounds positioned underneath text and vice versa
The following code incorrectly fails both rules despite rendering white text on a black background:

```html
<!DOCTYPE html>

<style>
  div,
  p {
    position: relative;
  }

  span {
    position: absolute;
    top: 0;
    right: 0;
    bottom: 0;
    left: 0;
    background-color: black;
  }

  p {
    color: white;
  }
</style>

<div>
  <span></span>
  <p>Hello world</p>
</div>
```

```html
<!DOCTYPE html>

<style>
  div,
  p {
    position: relative;
  }

  div::before {
    content: "";
    position: absolute;
    top: 0;
    right: 0;
    bottom: 0;
    left: 0;
    background-color: black;
  }

  p {
    color: white;
  }
</style>

<div>
  <p>Hello world</p>
</div>
```

- [ ] Follow up on DEV-11398.
- [ ] Follow up on DEV-11445.
- [ ] Follow up on DEV-11472.
- [ ] Follow up on DEV-11520.
- [ ] Follow up on DEV-11615.
- [ ] Follow up on DEV-11716.
- [ ] Follow up on DEV-11704.
- [ ] Follow up on DEV-11757.
- [ ] Follow up on DEV-11770.
- [ ] Follow up on DEV-11933.
- [ ] Follow up on DEV-11963.
- [ ] Follow up on DEV-12121.
- [ ] Follow up on DEV-12244.

kasperisager:
Resolves #758


packages/alfa-rules/src/common/expectation/get-colors.ts
```diff
@@ -1,134 +1,143 @@
...
+import { Device } from "@siteimprove/alfa-device";
+import { Element } from "@siteimprove/alfa-dom";
+import { Option, None } from "@siteimprove/alfa-option";
+import { Result, Err } from "@siteimprove/alfa-result";
 import { Style } from "@siteimprove/alfa-style";
-import { Iterable } from "@siteimprove/alfa-iterable";
 
-const { flatMap, map } = Iterable;
+import { hasInterposedDescendant } from "../predicate/has-interposed-descendant";
+import { isPositioned } from "../predicate/is-positioned";
```
Jym77, line 9:
```suggestion
import { hasInterposedDescendant, isPositioned } from "../predicate";
```

packages/alfa-rules/src/common/expectation/get-colors.ts
```diff
@@ -150,140 +159,182 @@ function getLayers(
...
+    return Err.of(Layer.Error.HasNonStaticPosition);
+  }
+
+  // If the background layer does not have a lower layer that is fully opaque,
+  // we need to also locate the background layers sitting behind the current
+  // layer.
+
+  if (hasInterposedDescendant(device)(element)) {
+    return Err.of(Layer.Error.HasInterposedDescendant);
+  }
```
Jym77, line 206: That comment doesn't seem to match the following code.
Looks like it's comment for the previous step, and that comment from here is missing.

Jym77 (changes requested):

packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -70,3 +72,23 @@ function isInvisible(device: Device, context?: Context): Predicate<Node> {
...
+              return dimension.value > 0;
+            default:
+              return false;
+          }
+        },
+        device
+      )
+    )
+  );
+}
```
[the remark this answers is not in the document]
kasperisager: They're close, but not quite. `hasDimensions()` is meant to check if an element has explicit dimensions and so might be visible even though it doesn't have any content of its own. It should really also be looking at things like `min-<width | height>`, but `<width | height>` was enough for the current cases we're incorrectly failing.

packages/alfa-rules/src/common/expectation/get-colors.ts
kasperisager: Good catch, that was a brain fart 🙈

Jym77 (approved):

Jym77 (approved):

PR #915, opened 2021-09-07
Implements ´programmatically hidden´

elenamongelli:
Resolves #


packages/alfa-rules/src/common/predicate/is-programmatically-hidden.ts
```diff
@@ -0,0 +1,68 @@
...
+      "visibility",
+      (visibility) => visibility.value !== "visible",
+      device,
+      context
+    ),
+    hasHiddenAncestors(device, context)
+  );
+}
+
+function hasHiddenAncestors(
```
Jym77, line 31: Let's move the `cache` declaration here so it stays near to its usage.

Jym77 (dismissed):
The file needs to be added to `tsconfig`. Plus generating the documentation.

Jym77 (approved):

PR #916, opened 2021-09-07
SIA-R21: Target not "programmatically hidden" elements that are not in the accessibility tree

### LINKED ISSUE #895: SIA-R21 should use "programmatically hidden" rather than "included in the accessibility tree"
See https://github.com/act-rules/act-rules.github.io/pull/1631 and https://github.com/Siteimprove/sanshikan/issues/204

elenamongelli:
Resolves #895


packages/alfa-rules/test/sia-r21/rule.spec.tsx
```diff
@@ -1,3 +1,4 @@
+/// <reference lib="dom" />
```
Jym77, line 1:
```suggestion
```

packages/alfa-rules/test/sia-r21/rule.spec.tsx
```diff
@@ -54,6 +55,19 @@ test("evaluates() fails an element with both a valid and an invalid role", async
   ]);
 });
 
+test("evaluates() fails an element with both a valid and an invalid role", async (t) => {
```
Jym77, line 58:
```suggestion
test("evaluates() fails an element with an invalid role and not included in the accessibility tree by default", async (t) => {
```

Jym77 (changes requested):

Jym77 (approved):

PR #917, opened 2021-09-08
SIA R91, R92, and R93: handle properties changes

### LINKED ISSUE #897: SIA R91, R92 and R93 do not account for descendants changing their font-size or affected property.
See https://github.com/Siteimprove/sanshikan/issues/200 / https://github.com/Siteimprove/sanshikan/pull/208

Jym77:
Resolves #897 

Solving the problem on R91. Needs to be propagated and shared correctly between all rules. And polished a bit.


packages/alfa-rules/src/common/outcome/text-spacing.ts
```diff
@@ -16,6 +16,12 @@ export function TextSpacing(name: Property.Name) {
       Diagnostic.of(`The \`${name}\` property is cascaded from another element`)
     ),
 
+    GoodText: Ok.of(
```
Jym77, line 19: Let's call this `WideEnough` 🤔

packages/alfa-rules/src/common/predicate/has-cascaded-value-declared-in-inline-style.ts
```diff
@@ -1,8 +1,10 @@
 import { Element } from "@siteimprove/alfa-dom";
```
Jym77, line 1: Iirc, this predicate can now be deleted since it was only used in these 3 rules.
(plus remove from `tsconfig` and `predicate.ts`)

elenamongelli: That's right, I am deleting it

kasperisager (changes requested):
Looks like the formatting is off in a couple of places but it looks fine otherwise 👍

kasperisager (approved):
🚀

PR #935, opened 2021-09-30
Add additional `Thenable` helpers

kasperisager:
This pull request adds some additional helper functions for the `Thenable` type to maintain symmetry with `Promise` and `Future`. You might be wondering why we even have all these types as they can all seem like the same way to accomplish the same thing; asynchronous control flow.

The best way to reason about having 3 different types to achieve essentially the same thing is perhaps by comparing them to the `Iterable`, `Array`, and `Sequence` types as they have a lot in common:

- `Iterable` models sequences of values and `Thenable` models asynchronous control flow in a lazy, stack-unsafe manner by using the call stack. They provide excellent performance at the risk of blowing the stack.

- `Array` models sequences of values and `Promise` models asynchronous control flow in an eager, stack-safe manner by using the heap and task queue. They provide reasonable performance without risk of blowing the stack.

- `Sequence` models sequences of values and `Future` models asynchronous control flow in a lazy, stack-safe manner by using a trampoline and the task queue. They provide good performance without risk of blowing the stack.

Each type has its place and it's of course up to the programmer to pick the right tool for the job.

### DIFF
--- a/docs/review/api/alfa-thenable.api.md
+++ b/docs/review/api/alfa-thenable.api.md
@@ -4,12 +4,42 @@
 
 ```ts
 
+import { Array as Array_2 } from '@siteimprove/alfa-array';
 import { Callback } from '@siteimprove/alfa-callback';
+import { Continuation } from '@siteimprove/alfa-continuation';
+import { Mapper } from '@siteimprove/alfa-mapper';
 
 // @public (undocumented)
 export interface Thenable<T, E = unknown> {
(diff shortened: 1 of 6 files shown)


Jym77 (approved):

kasperisager:
Bonus info: `Array` and `Sequence` both implement `Iterable` just as `Promise` and `Future` both implement `Thenable`.

PR #964, opened 2021-10-29
`isClipped`: jump to offset parents of absolutely positioned elements

### LINKED ISSUE #887: `isClipped()` does not account for offset parents
Same situation as with #824, just for the `isClipped()` predicate.

elenamongelli:
Resolves #887


packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -6,22 +6,26 @@ import { Refinement } from "@siteimprove/alfa-refinement";
...
  * Checks if a node (or one of its ancestor) is fully clipped
  */
+
 export function isClipped(
   device: Device,
   context: Context = Context.empty()
 ): Predicate<Node> {
-  return (node) =>
-    cache
+  return function isClippedInternal(node: Node): boolean {
```
Jym77, line 27:
```suggestion
  return function isClipped(node: Node): boolean {
```
We can keep the same name here. It's OK to do so because the outer function returns the inner one, so they are indeed the same thing just with some wrapper around.

packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -37,19 +41,27 @@ export function isClipped(
...
             ),
             // Or its parent is clipped
-            (node: Node) =>
-              node
-                .parent({
-                  flattened: true,
-                  nested: true,
-                })
-                .some(isClipped(device, context))
+            hasRelevantParent(device, isClipped(device, context))
```
Jym77, line 44:
```suggestion
            hasRelevantParent(device, isClipped)
```
That needs to call the inner function. Or at least, it can call the inner function which avoids piling up repeated arguments on the call stack.

packages/alfa-rules/src/common/predicate/is-clipped.ts
```diff
@@ -37,19 +41,27 @@ export function isClipped(
...
+): Predicate<Node> {
+  return (node) =>
+    (and(isElement, not(isPositioned(device, "static")))(node)
+      ? getOffsetParent(node, device)
+      : node.parent({
+          flattened: true,
+          nested: true,
+        })
+    ).some(predicate);
+}
```
Jym77, line 64: It seems that the other place where we do use relevant parent is in R83 where we always have the construction `getRelevantParent(…).some(predicate)`.
So we should probably do the following refactoring:
* Add in `common/expectation/` a `getPositionningParent` which does the full branching based on `position`.
* Add in `common/predicate` a `hasPositionningParent` which calls it and does the `.some(predicate)` bit.
* Use `hasPositionningParent` here and in R83, deleting the `RelevantParent` stuff in both places.

Jym77 (changes requested):
Small improvements.

Jym77 (changes requested):
Small clean up.

Jym77 (approved):

PR #967, opened 2021-11-02
SIA-R62: Accept differences in `box-shadow`

### LINKED ISSUE #888: SIA-R62 should accept differences in `box-shadow`
Yet another way of removing the default underline and replace it with a custom one 🙄 

```css
a {
  box-shadow: inset 0 -1px 0 rgb(0 38 99 / 30%);
  text-decoration: none;
}
```

elenamongelli:
Resolves #888


packages/alfa-rules/src/sia-r62/rule.ts
```diff
@@ -277,7 +278,8 @@ function isDistinguishable(
     // should hopefully not happen (too often) in practice. When it does, we
     // risk false negatives.
     hasOutline(device, context),
-    hasBorder(device, context)
+    hasBorder(device, context),
+    hasBoxShadow(device, context) //Checks for color != transparent and spread => 0
```
zsofitoth, line 282: What does the comment here mean "check any differences in box-shadow"? https://github.com/Siteimprove/alfa/issues/888#issuecomment-904429476 I interpret it as if it's not transparent or it has spread, then it's a passed test case

Jym77: Yes, almost 👍 . More or less the same that happens for `border`, for example.
We can be sure there is no shadow in three cases: it's actually `none`; or it is fully transparent; or it has a spread of `0` (🤔 actually it seems to be only invisible if all of horizontal offset, vertical offset, blur **and* spread are 0). If we are in any of these 3 cases, we can consider there is no `box-shadow` at all.

The thing is that we can have a `box-shadow` which is not actually visible (e.g. because it has the same colour as the background), or barely visible (e.g. `#111` vs `#000`).

But we won't try and be clever about that (we might decide to later) because this is either fairly complex (determining background colour is a mess) or arbitrarily (when is the limit between "barely visible" and "OK?) This fits the basic goal that Alfa flags stuff it is (fairly) sure are errors but let it go when it cannot be sure. In these case, Alfa would see a shadow and conclude that it might be sufficient and thus shouldn't be flagged.

packages/alfa-rules/src/sia-r62/serialise.ts
```diff
@@ -71,6 +72,33 @@ export namespace Serialise {
...
   }
 
+  export function boxShadow(style: Style): string {
+    const boxShadow = style.computed("box-shadow").value;
+
+    if (Keyword.isKeyword(boxShadow)) {
+      return "";
+    }
+
+    const serializedShadows = [] as Array<string>;
```
Jym77, line 82:
```suggestion
    const serializedShadows: Array<string> = [];
```
`as` is doing an actual cast while `:` is just a hint, and compiler is more likely to complain about a bad `:`.

Jym77 (changes requested):
Great work 🎉

Jym77 (approved):

PR #968, opened 2021-11-02
SIA-R62: Accept differences in `font-family`

### LINKED ISSUE #952: SIA-R62 should accept more differences in `font`
Currently only accepting differences in `font-weight`.
Seen a case like
```css
p { font-family: VAGRoundedStd-Light}
a { font-family: VAGRoundedStd-Bold}
```

elenamongelli:
Resolves #952


Jym77 (changes requested):
We should build a "unified" `hasDistinguishableFont` helper to tackle all the associate longhands.

packages/alfa-rules/src/sia-r62/rule.ts
```diff
@@ -344,25 +343,35 @@ function hasDistinguishableBackground(
...
+  return or(
+    hasComputedStyle(
+      "font-weight",
+      not((weight) => weight.equals(referenceWeight)),
+      device,
+      context
+    ),
+    hasComputedStyle(
+      "font-family",
+      not((family) => family.equals(referenceFamily)),
```
Jym77, line 371: We should only look at the first in the list, not at everything. So that `font-family: some-font, foo` and `font-family: some-font, bar` are considered the same (under the assumption that `some-font` loads, of course).

packages/alfa-rules/src/sia-r62/serialise.ts
```diff
@@ -71,6 +71,33 @@ export namespace Serialise {
...
+    const family = style.computed("font-family");
+
+    if (optional !== " ") {
+      // Optional properties were changed, need to output the mandatory ones.
+      return normalize(`${optional} ${size} ${family}`);
+    }
+
+    if (
+      size.value.equals(Property.get("font-size").initial.value) &&
+      family.value.equals(Property.get("font-family").initial.values[0])
```
Jym77, line 89: We should only get the first in the list of the computed one. So, I  guess,
```suggestion
      family.value.values[0].equals(Property.get("font-family").initial.values[0])
```
```suggestion
      family.value.equals(Property.get("font-family").initial.values[0])
```

Jym77 (changes requested):
Let's add examples with `font-family` list, to make clear what we want to have:
* `a { font-family: some-font, serif }` => passes because `some-font` is different from the default.
* `a { font-family: serif, some-font }` => fails, because only the first font is looked at.
* `a { font-family: serif, some-font }` and `p { font-family: some-other-font }` => passes, but the diagnostic shouldn't contain a `font` because the link's font is the default one. (`some-font` is ignored).

Jym77:
Closing in favour of #994

PR #989, opened 2021-11-25
SIA-R28: Reject default name 'Submit Query' for inputs with type image

### LINKED ISSUE #973: SIA-R28 should reject default name ("Submit Query")
Following act-rules/act-rules.github.io#1717

zsofitoth:
Resolves #973

### DIFF
--- a/packages/alfa-rules/src/sia-r28/rule.ts
+++ b/packages/alfa-rules/src/sia-r28/rule.ts
@@ -6,9 +6,7 @@ import { Criterion, Technique } from "@siteimprove/alfa-wcag";
 import { Page } from "@siteimprove/alfa-web";
 
 import { expectation } from "../common/expectation";
-
-import { hasAccessibleName } from "../common/predicate/has-accessible-name";
-import { isIgnored } from "../common/predicate/is-ignored";
+import { hasAccessibleName, isIgnored } from "../common/predicate";
 
 const { isElement, hasInputType, hasNamespace } = Element;
 const { and, not, test } = Predicate;
@@ -40,10 +38,7 @@ export default Rule.Atomic.of<Page, Element>({
         return {
           1: expectation(
             test(
-              hasAccessibleName(
-                device,
-                (accessibleName) => accessibleName.value !== "Submit Query"
-              ),
+              hasAccessibleName(device, (name) => name.source.length !== 0),
               target
             ),
             () => Outcomes.HasAccessibleName,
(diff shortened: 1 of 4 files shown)


packages/alfa-rules/src/sia-r28/rule.ts
```diff
@@ -39,9 +39,15 @@ export default Rule.Atomic.of<Page, Element>({
...
           1: expectation(
-            hasNonEmptyAccessibleName(device)(target),
-            () => Outcomes.HasName,
-            () => Outcomes.HasNoName
+            test(
+              hasAccessibleName(
+                device,
+                (accessibleName) => accessibleName.value !== "Submit Query"
+              ),
+              target
```
Jym77, line 47: Thinking back on this…
We might be able to determine that the name comes from the default value by checking its source: https://github.com/Siteimprove/alfa/blob/f37f76b14fc856f8c76633be8bd21eff91778f9d/packages/alfa-aria/src/name.ts#L47-L49
If I get it properly, default names should have an empty source, so
```suggestion
              hasAccessibleName(
                device,
                (name) => name.sources.length !==0
              ),
              target
```
That way we're solving the problem of duplicating the default name 😄 ; and we accept `<input type="image" aria-label="Submit Query" />` where we can assume that the devs know what they do and that the name is actually good.

Jym77 (changes requested):
Looks good 👍 
A bit of streamlining in what we do in other places.

Note on testing accessible names: contrarily to ACT rules, we do have extensive tests on accessible name computation (https://github.com/Siteimprove/alfa/blob/main/packages/alfa-aria/test/name.spec.tsx) so we don't usually need to be as heavy on the "has an accessible name"/"has no accessible name" examples (these essentially test the accessible name computation through `alt`, `aria-label`, `aria-labelledby`, … but do not really test the code in the rule). It's usually enough to have one case with a name and one without.
(no need to erase the tests, they're here and they won't harm…)

packages/alfa-rules/src/sia-r28/rule.ts
```diff
@@ -40,10 +38,7 @@ export default Rule.Atomic.of<Page, Element>({
         return {
           1: expectation(
             test(
-              hasAccessibleName(
-                device,
-                (accessibleName) => accessibleName.value !== "Submit Query"
-              ),
+              hasAccessibleName(device, (name) => name.source.length !== 0),
```
Jym77, line 41: Let's help our future selves.
```suggestion
              // Rejecting default name.
              hasAccessibleName(device, (name) => name.source.length !== 0),
```

Jym77 (dismissed):

Jym77 (approved):

PR #991, opened 2021-11-29
SIA-R14: Ignore punctuation in label names

### LINKED ISSUE #972: SIA-R14 should ignore punctuation
[Understanding 2.5.3](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html) has a section about punctuation, citing it as often being a valid exception. While the phrasing is not clear that punctuation should always be ignored, it is nonetheless clear that it should sometimes be.

We should likely stay on the safe side and always trim punctuation from the strings before matching.

zsofitoth:
Resolves #972


packages/alfa-rules/src/sia-r14/rule.ts
```diff
@@ -152,6 +150,11 @@ function childrenPerceivableText(node: Node, device: Device): string {
   return result;
 }
 
+// Includes removing all punctiation (underscore, hypen, brackets, quotation marks, etc)
+function normalize(input: string): string {
+  return input.trim().toLowerCase().replace(/\p{P}/gu, "").replace(/\s+/g, " ");
+}
+
```
Jym77, line 157: I think it might be better to keep the common `normalise` and use here `normalise(foo).replace(/\p{P}/gu, "")`.
That way we'll avoid having several normalisations and keep the shared one with a single meaning. Then callers may do additional treatments.

Also great find with the `\p` magic 👍

Jym77 (changes requested):

packages/alfa-rules/src/sia-r14/rule.ts
zsofitoth: We can do that, I'm suspicious that it might turn into `normalise(foo).replace(/\p{P}/gu, "").replace(/\s+/g, " ")` though to handle `Hello? - ( world )`

Jym77: Ah, yes…
Then `normalise(foo.replace(/\p{P}/gu, "")` ?

zsofitoth: It will look like this within the expectation:

```Javascript
const perceivableInnerTextFromElement =
  getPerceivableInnerTextFromElement(target, device);
const textContent = normalize(
  // Removes all punctiation (underscore, hypen, brackets, quotation marks, etc)
  perceivableInnerTextFromElement.replace(/\p{P}/gu, "")
);
let name = "";

const accessibleNameIncludesTextContent = test(
  hasAccessibleName(device, (accessibleName) => {
    const value = accessibleName.value;
    // Removes all punctiation (underscore, hypen, brackets, quotation marks, etc)
    name = normalize(value.replace(/\p{P}/gu, ""));
    return name.includes(textContent);
  }),
  target
);
```

Should I leave the comments at both places? I'll push it up if this looks good :)

Jym77: Maybe make a local function so that we don't duplicate code (I do foresee we might want to add some other processing at some point…)

packages/alfa-rules/src/sia-r14/rule.ts
```diff
@@ -54,14 +55,21 @@ export default Rule.Atomic.of<Page, Element>({
...
           getPerceivableInnerTextFromElement(target, device)
         );
+        const textContent = normalize(perceivableInnerTextFromElement);
         let name = "";
 
         const accessibleNameIncludesTextContent = test(
           hasAccessibleName(device, (accessibleName) => {
-            name = normalize(accessibleName.value);
+            const value = removePunctuation(accessibleName.value);
+            name = normalize(value);
```
Jym77, line 72:
```javascript
        // Removes all punctuation (underscore, hypen, brackets, quotation marks, etc)
        // and normalise
        function removePunctuationAndNormalise(input: string): string {
          return normalise(input.replace(/\p{P}/gu, ""));
        }

        const textContent = removePunctuationAndNormalize(getPerceivableInnerTextFromElement(target, device));

        let name = "";
        const accessibleNameIncludesTextContent = test(
          hasAccessibleName(device, (accessibleName) => {
            name = removePunctuationAndNormalize(accessibleName.value);
```

Jym77 (changes requested):

packages/alfa-rules/src/sia-r14/rule.ts
```diff
@@ -56,20 +56,20 @@ export default Rule.Atomic.of<Page, Element>({
...
+        // and normalise
+        function removePunctuationAndNormalise(input: string): string {
+          return normalize(input.replace(/\p{P}/gu, ""));
         }
 
-        const perceivableInnerTextFromElement = removePunctuation(
+        const perceivableInnerTextFromElement = removePunctuationAndNormalise(
           getPerceivableInnerTextFromElement(target, device)
         );
         const textContent = normalize(perceivableInnerTextFromElement);
```
Jym77: Let's avoid the intermediate variable here
```suggestion
        const textContent = removePunctuationAndNormalise(
          getPerceivableInnerTextFromElement(target, device);
```

packages/alfa-rules/src/sia-r14/rule.ts
```diff
@@ -56,20 +56,20 @@ export default Rule.Atomic.of<Page, Element>({
...
         let name = "";
 
         const accessibleNameIncludesTextContent = test(
```
Jym77, line 70: And group lines logically.
```suggestion

        let name = "";
        const accessibleNameIncludesTextContent = test(
```

Jym77 (changes requested):

packages/alfa-rules/src/sia-r14/rule.ts
zsofitoth: Must have been really off yesterday 🙈

Jym77 (approved):

PR #996, opened 2021-12-01
Add function to create a WCAG `Criterion` from its URI

Jym77:
Usage:
```javascript
import { Conformance, Criterion } from "@siteimprove/alfa-wcag";

const contrast = Criterion.fromURI(
  "https://www.w3.org/TR/WCAG/#contrast-minimum"
).get();

console.log(Conformance.isA()(contrast)); // false
console.log(Conformance.isAA()(contrast));  // true
console.log(Conformance.isAAA()(contrast));  // true
```

### DIFF
--- a/packages/alfa-wcag/src/criterion.ts
+++ b/packages/alfa-wcag/src/criterion.ts
@@ -1,5 +1,6 @@
 import { Requirement } from "@siteimprove/alfa-act";
 import { Branched } from "@siteimprove/alfa-branched";
+import { None, Option } from "@siteimprove/alfa-option";
 
 import { Criteria } from "./criterion/data";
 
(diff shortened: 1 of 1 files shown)


packages/alfa-wcag/src/criterion.ts
```diff
@@ -224,4 +225,16 @@ export namespace Criterion {
   export function isCriterion(value: unknown): value is Criterion {
     return value instanceof Criterion;
   }
+
+  export function fromURI(uri: string): Option<Criterion> {
+    for (const [chapter, value] of Object.entries(Criteria)) {
+      for (const version of value.versions) {
+        if (version[1].uri === uri) {
```
zsofitoth, line 232: Is it always guaranteed that the `uri` is accessible from `version[1]`?

zsofitoth (approved):
LGTM

PR #997, opened 2021-12-01
Add some framework tests

Jym77:
Add basic test for building pages out of React or Vue components.
Didn't manage to build any Angular component 😖


packages/alfa-angular/tsconfig.json
```diff
@@ -1,4 +1,7 @@
 {
+  "compilerOptions": {
+      "experimentalDecorators": true
+  },
```
zsofitoth, line 4: Is this change necessary? I don't see other updates to the `angular.spec.tsx`

zsofitoth (approved):
Looks good to me 👍

packages/alfa-angular/tsconfig.json
Jym77: Indeed, not needed until we actually get to write a test that needs it.

PR #998, opened 2021-12-02
SIA-ER62: Accept difference in contrast of 3:1 or more as distinguishing feature

### LINKED ISSUE #796: SIA-ER62 should accept a difference in contrast of 3:1 or more as valid indicator
We left that on the side when initially writing the rule, for the sake of simplicity (https://github.com/Siteimprove/alfa/pull/538#issuecomment-725428875). I have seen this week 3 feedbacks about this being incorrect, each pointing to [G183 ](https://www.w3.org/TR/WCAG20-TECHS/G183.html)(so I guess each from accessibility experts…)

So, it is maybe time for us to consider improving the rule…

- [ ] Follow up on DEV-12271.
- [ ] Follow up on DEV-12416.
- [ ] Follow up on DEV-12447.

zsofitoth:
Resolves #796


packages/alfa-rules/test/common/expectation/get-colors.spec.tsx
```diff
@@ -130,3 +131,67 @@ test("getForeground() handles mix of opacity and transparency", (t) => {
...
+          color: "red",
+        }),
+      ]),
+    ]
+  );
+
+  t.deepEqual(
+    Iterable.first(getForeground(target, device, Context.hover(target)).get())
+      .get()
+      .toJSON(),
```
Jym77, line 152: To make the test more complete, we should also check that `getForeground` without context return the non-hovered colour.

packages/alfa-rules/test/common/expectation/get-colors.spec.tsx
```diff
@@ -130,3 +131,67 @@ test("getForeground() handles mix of opacity and transparency", (t) => {
...
+        h.rule.style("a:hover", {
+          color: "red",
+        }),
+      ]),
+    ]
+  );
+
+  t.deepEqual(
+    Iterable.first(getForeground(target, device, Context.hover(target)).get())
+      .get()
```
Jym77, line 151:
```suggestion
getForeground(target, device, Context.hover(target)).get()[0]
```
Should work 🤔

packages/alfa-rules/test/sia-er62/rule.spec.tsx
```diff
@@ -904,3 +904,44 @@ test(`evaluate() is inapplicable to an <a> element with a <p> parent element
...
+  t.deepEqual(await evaluate(ER62, { document }), [
+    passed(ER62, target, {
+      1: Outcomes.IsDistinguishable([style], [style], [style]),
+    }),
+  ]);
+});
+
+// Check gradient to non-gradient
+// Check non-gradient to gradient
+// Check gradient to gradient
```
Jym77, line 946: We already have (or should 🤞 ) tests that `getColor` behaves in presence of gradients, so we don't need to be too heavy here (is the new test going to detect bugs in the rule or in `getColor`?)
We still want one test with several pairings (especially once we have the extended diagnostic) to be sure that we do compute and send all of them correctly.

Jym77 (commented):
Good work 👍

packages/alfa-rules/test/sia-er62/rule.spec.tsx
zsofitoth: I gave this one a shot, now the gradient specific tests should be in `get-colors.spec.tsx`

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -133,22 +143,71 @@ export default Rule.Atomic.of<Page, Element>({
...
+          link: Element,
+          device: Device,
+          context?: Context
+        ): Array<Contrast.Pairing> {
+          const containerForegrounds = getForeground(
+            container,
+            device,
+            context
+          ).get();
+          const linkForegrounds = getForeground(link, device, context).get();
```
zsofitoth, line 157: error case not handled with `.get()`

packages/alfa-rules/src/sia-er62/diagnostics.ts
```diff
@@ -0,0 +1,203 @@
...
+      ["color", Serialise.getLonghand(style, "color")] as const,
+      ["font", Serialise.font(style)] as const,
+      [
+        "vertical-align",
+        Serialise.getLonghand(style, "vertical-align"),
+      ] as const,
+      ["background", Serialise.background(style)] as const,
+      ["outline", Serialise.outline(style)] as const,
+      ["text-decoration", Serialise.textDecoration(style)] as const,
+      // ["cursor", Serialise.getLonghand(style, "cursor")] as const,
```
zsofitoth, line 120: This I found commented out when moved, not sure if it's on purpose or a leftover

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -272,343 +314,216 @@ function hasNonLinkText(device: Device): Predicate<Element> {
...
+    container: Element,
+    device: Device,
+    context: Context = Context.empty()
+  ): Predicate<Element> {
+    return or(
+      // Things like text decoration and backgrounds risk blending with the
+      // container element. We therefore need to check if these can be distinguished
+      // from what the container element might itself set.
+      hasDistinguishableContrast(container, device, context),
+      hasDistinguishableBackground(container, device, context),
```
zsofitoth, line 328: This is a bit fragile, because of the caching and calculating the contrast pairings once, `hasDistinguishableContrast` has to be the first one otherwise some tests fail. My theory is that if anything else is "true" that's calculated before the things after won't be checked.

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -38,6 +38,12 @@ const { isText } = Text;
 const { or, not, test } = Predicate;
 const { and } = Refinement;
 
+const hasNonLinkTextCache = Cache.empty<Element, boolean>();
```
Jym77, line 41: This cache is only used for one function, so it's better to declare it just before the function (keep the info local).

I'm also realising that `hasNonLinkText` is device dependant, so this cache should be 🙈 not breaking anything since we don't run test with different devices yet, we'll fix that later…

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -38,6 +38,12 @@ const { isText } = Text;
...
+const pairwiseContrastCache = Cache.empty<
+  Element,
+  Cache<Element, Contrast.Pairing[]>
+>();
+
```
Jym77, line 46: This cache should also depend on device (not very important currently), but especially on context (very important for the rule, otherwise we might fetch an `hover` answer when asking for a `focus` pairing or similar problem).
```suggestion
const pairwiseContrastCache = Cache.empty<Device, Cache<Context, Cache<
  Element,
  Cache<Element, ReadonlyArray<Contrast.Pairing>>
>>>();

```
(also, we normally use the `Array<type>` notation rather than `foo[]`, and we likely want that array to be read-only to avoid the risk of changing it after computing it)

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -186,8 +161,14 @@ export default Rule.Atomic.of<Page, Element>({
...
               const distinguishableContrast = Set.from(
                 Array.from(nonLinkElements)
-                  .map((container) =>
-                    getPairwiseContrast(container, link, device, context)
+                  .map(
+                    (container) => {
+                      const pairings = pairwiseContrastCache.get(
+                        container,
+                        Cache.empty
+                      ).get(link, () => []);
```
Jym77: We should (re)-do the computation here. If, for some reason, the cache is empty when we reach that point, then we do need to compute the contrast and finding it.

Since that contrast computation is duplicated between the two places that use the cache, extracting it to a function will help (and after all the `Cache.get(…)` when calling it we do have `device`, `context`, `link` and `container` already available, so it should be the simple version of it.

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -186,8 +161,14 @@ export default Rule.Atomic.of<Page, Element>({
               );
               const distinguishableContrast = Set.from(
                 Array.from(nonLinkElements)
```
Jym77, line 163: We probably don't need to pile up `Set.from(Array.from(…))`. `Set.from()` should work on any `Iterable` and `nonLinkElements` is already a `Sequence` (iirc), so we can just use `Set.from`.

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -186,8 +161,14 @@ export default Rule.Atomic.of<Page, Element>({
...
+                  .map(
+                    (container) => {
+                      const pairings = pairwiseContrastCache.get(
+                        container,
+                        Cache.empty
+                      ).get(link, () => []);
+                      return pairings;
+                    }
                   )
                   .flat(1)
```
Jym77, line 169: Consider using `.flatMap` instead of `map.flat`.

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -349,8 +328,8 @@ namespace Distinguishable {
       // Things like text decoration and backgrounds risk blending with the
       // container element. We therefore need to check if these can be distinguished
       // from what the container element might itself set.
-      hasDistinguishableBackground(container, device, context),
       hasDistinguishableContrast(container, device, context),
+      hasDistinguishableBackground(container, device, context),
```
Jym77: Indeed, the `or` is lazy and escape as soon as one condition is `true`. Which can make the cache not being filled, and then creates problem in the expectation when trying to `.get` it.
By copying the cache filling in the expectation, we ensure the cache always get the correct data (and the cache implementation makes sure we do not compute the same thing twice and don't waste time). This should allow to keep the checks here in alphabetical order which is easier to navigate for humans 😄

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -442,17 +421,26 @@ namespace Distinguishable {
...
               Array.map(linkColors, (linkColor) => {
-                return contrast(containerColor, linkColor);
+                return Contrast.Pairing.of(
+                  containerColor,
+                  linkColor,
+                  contrast(containerColor, linkColor)
+                );
               })
             ),
           ];
```
Jym77: That computation should be inside the nested `Cache.get` calls.
Currently, this always compute the pairing, then look in the cache and use the computation if the cache is empty. But the computation always happens…
The reason the second argument of `Cache.get` is a function is precisely to delay computation and don't do it if not needed.

Jym77 (changes requested):
Caches are taking care for us of only computing the inner call when needed. So when using cache, we can heavily put computations inside `Cache.get`, it will only be done once. That let us (humans) not care about call order or other stuff and not worry whether the data has been filled or not.
(Comparatively, `containers` in that rule is only a `Map`, so we do have to make sure we fill it before we query it, and this is ensured by always filling it in Applicability just before `yield`ing a new target, and querying it only in Expectation. In the case of `containers`, the rule structure ensures that we know when is the first time we need it (for each target) and should fill it. In the case of pairings, it may depend on many things. `Cache` let us delegate the "is this the first time I need it?" bit to the code)

packages/alfa-rules/src/sia-er62/diagnostics.ts
```diff
@@ -0,0 +1,202 @@
...
+type Name = Property.Name | Property.Shorthand.Name;
+
+export class ElementDistinguishable
+  implements Equatable, Hashable, Serializable
+{
+  public static of(
+    style: Iterable<readonly [Name, string]> = [],
+    pairings: ReadonlyArray<Contrast.Pairing> = []
+  ): ElementDistinguishable {
+    return new ElementDistinguishable(Map.from(style), pairings);
```
Jym77, line 29: At this point, we can be a bit more generous in what we accept
```suggestion
    pairings: Iterable<Contrast.Pairing> = []
  ): ElementDistinguishable {
    return new ElementDistinguishable(Map.from(style), Array.from(pairings));
```

packages/alfa-rules/src/sia-er62/diagnostics.ts
```diff
@@ -0,0 +1,202 @@
...
+  public get style(): Map<Name, string> {
+    return this._style;
+  }
+
+  public get pairings(): ReadonlyArray<Contrast.Pairing> {
+    return this._pairings;
+  }
+
+  public withStyle(
+    ...styles: ReadonlyArray<readonly [Name, string]>
```
Jym77, line 52:
```suggestion
    ...styles: Iterable<readonly [Name, string]>
```
Let's also improve that one.

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -135,20 +143,69 @@ export default Rule.Atomic.of<Page, Element>({
...
+                        Distinguishable.getPairwiseContrast(
+                          container,
+                          link,
+                          device,
+                          context
+                        )
+                      )
+                  )
+                )
+              ).toArray();
```
Jym77, line 183:
```suggestion
              );
```
And we can remove the `toArray` here (after making `ElementDistinguishable` accept any `Iterable`).
That will make future changes easier, since it is now the `.of` who takes care about what precise format is needed to build the instance, and callers don't care.

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -135,20 +143,69 @@ export default Rule.Atomic.of<Page, Element>({
...
+                ? Ok.of(
+                    ElementDistinguishable.of(
+                      ElementDistinguishable.serialise(
+                        link,
+                        device,
+                        target,
+                        context
+                      ),
+                      distinguishableContrast
+                    )
```
Jym77, line 195: Let's make an `ElementDistinguishable.from(link, device, target, context, pairings)` that takes care of calling `.serialise`.
Again, no need for the caller to know it has to do some serialisation, and future versions of this diagnostic could well handle it differently.


As a rule of thumb, we often have:
* In the class, a private constructor that takes the data in the same shape as the the `private readonly` variables storing it (and as flat as needed, the getters may nest data that makes sense being nested).
* In the class, a `public static of` that takes the data almost in the same shape as the constructor, maybe taking care of turning any `Iterable` into the correct kind, or having optional arguments turned into actual `Option`.
* In the namespace, if needed, a `.from` that takes data in a fairly different format and perform some possibly complex computation to turn it into the correct format.

Here, `Element.serialise` is the old `ComputedStyles.from`, but it is more in line with the rest of the code to keep that one flat also; even if the resulting `.from` might end up calling subfunction to modify each piece of data in a different way.

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -135,20 +143,69 @@ export default Rule.Atomic.of<Page, Element>({
...
+                : Err.of(
+                    ElementDistinguishable.of(
+                      ElementDistinguishable.serialise(
+                        link,
+                        device,
+                        target,
+                        context
+                      ),
+                      distinguishableContrast
+                    )
```
Jym77, line 206: Same.

Jym77 (changes requested):
Looks great 👍 
A bit of polish to keep the code closer to what is done elsewhere.

packages/alfa-rules/src/sia-er62/diagnostics.ts
zsofitoth: I could no longer use the rest operator due to a compile error - _A rest parameter must be of an array type._ so I rewrote it without the spread and adjusted the tests

packages/alfa-rules/src/sia-er62/diagnostics.ts
```diff
@@ -49,7 +49,7 @@ export class ElementDistinguishable
   }
 
   public withStyle(
-    ...styles: ReadonlyArray<readonly [Name, string]>
+    styles: Iterable<readonly [Name, string]>
```
Jym77, line 52: Indeed, my bad… `...` accept any iterable when used as spread operator, not as rest operator (i.e. when using a variable, not when declaring a parameter) 🙈 

We should keep the rest operator here (and thus, the `ReadonlyArray`) in order to allow `.withStyle(["prop1", "value"], ["prop2", "value2"])` without needing to wrap everything in `[ ]` again.

Sorry about that 😞

zsofitoth: It's alright, I'll revert that change, I just thought it was easier to read even though it's annoying to wrap

Jym77 (approved):

PR #1003, opened 2021-12-07
Accept title attribute for name of input type=image

Jym77:
We were failing [this example](https://act-rules.github.io/rules/59796f#passed-example-3)

### DIFF
--- a/packages/alfa-aria/test/name.spec.tsx
+++ b/packages/alfa-aria/test/name.spec.tsx
@@ -1,3 +1,4 @@
+/// <reference lib="dom" />
 import { h } from "@siteimprove/alfa-dom/h";
 import { test } from "@siteimprove/alfa-test";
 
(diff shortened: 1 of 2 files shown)


packages/alfa-aria/test/name.spec.tsx
```diff
@@ -1083,6 +1084,20 @@ test(`.from() determines the name of an <input> element with a placeholder
   });
 });
 
+test(`.from() determines the name of an \`<input type="image">\` with a
+      title attribute.`, (t) => {
+  const input = <input type="image" src="foo" title="Search" />;
+
+  // console.dir(Name.from(input, device).toJSON(), { depth: null });
```
zsofitoth, line 1091: Leftover `console.dir` and `reference`

packages/alfa-aria/test/name.spec.tsx
```diff
@@ -1167,7 +1182,7 @@ test(`.from() determines the name of an <input type="image"> element`, (t) => {
   t.deepEqual(Name.from(input, device).toJSON(), {
     type: "some",
     value: {
-      value: "Submit",
+      value: "Submit Query",
```
zsofitoth, line 1185: How come this didn't fail before? 😮

zsofitoth (approved):
Looks good 👍 Just some leftover console logging :)

packages/alfa-aria/test/name.spec.tsx
Jym77: I don't know and at this point I'm too afraid to dig 🙈

PR #1010, opened 2021-12-14
Add scopes

### LINKED ISSUE #993: Add Tags to rule (notably scopes)
We have a bit of tagging support to rules, scope should be added to it.
https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/tags/scope.ts

Jym77:
Resolves #993 

I got need of scopes for some of the examples I want to write… Took the opportunity to clean up a bit the imports since I was touching every rule…


packages/alfa-rules/src/sia-r109/rule.ts
```diff
@@ -24,8 +24,8 @@ const { and } = Refinement;
 export default Rule.Atomic.of<Page, Document, Question>({
   uri: "https://alfa.siteimprove.com/rules/sia-r109",
   requirements: [Criterion.of("3.1.1"), Technique.of("H57")],
-  tags: [Stability.Experimental],
-  evaluate({ device, document }) {
+  tags: [Scope.Page, Stability.Experimental],
+  evaluate({ document }) {
```
zsofitoth, line 28: How come the device is no longer needed?

packages/alfa-rules/src/sia-r75/rule.ts
```diff
@@ -7,15 +7,16 @@ import { Page } from "@siteimprove/alfa-web";
...
 
-import { hasTextContent } from "../common/predicate/has-text-content";
-import { isVisible } from "../common/predicate/is-visible";
+import { hasTextContent, isVisible } from "../common/predicate";
+import { Scope } from "../tags";
 
 const { isElement, hasNamespace } = Element;
 const { isText } = Text;
-const { and, or, not } = Predicate;
+const { and, or } = Predicate;
```
zsofitoth, line 15: Was the `not` predicate unused?

zsofitoth (approved):
Looks good 👍 Just a few questions

Jym77:
> Looks good 👍 Just a few questions

Apparently, they were both unused, but never cleaned up 🤷

PR #1020, opened 2022-01-07
Map questions' uri to the expected answer type and register questions

### LINKED ISSUE #1018: Consider linking questions' URI to their type
Questions expect a certain type as answer, and this is used in various places.
Questions with the same URI should likely expect the same answer type.

Enforcing this will require some way to register URI-to-type mapping. We currently have half a dozen question types which can easily be stored in one place, but URIs should be expandable at will, so this is likely going to be a bit tricky to work out correctly…

Jym77:
Resolves #1018 

---
Questions have a `type` parameter which is used to link the question to its expected answer type in a programmatic way. Notably, this ensures type safety in Oracles by making sure that a question expecting a `boolean` answer (`question.type === "boolean"`) will indeed get it (by case analysis on `question.type` and narrowing on `answer`).

To ensure this stays true, `Rule` (and most `alfa-act` classes) have a generic "question shape" type variable which is instantiated by  interfaces like:
```javascript
interface Question {
  boolean: boolean;
  "node[]": Iterable<Node>;
  …
}
```

With this PR, the type of questions is also linked to their URI. That is, two questions with the same URI must now have the same `type` and therefore expect the same type for answers. Since the URI is supposed to identify questions, it makes no sense to have the "same" question (same URI) expecting sometimes a boolean and sometimes a node.

While the `type` parameter of questions becomes a bit redundant, it is nonetheless kept as a programmatic way to find the expected type of answer. While it is not possible to branch code on the `ANSWER` type variable of a Question; it is possible to do so on `question.type`. Oracles can use that to ensure type safety when answering questions.

Thus, the "question shape" generic now looks like:
```javascript
interface Question {
  "question1": ["boolean", boolean];
  "question2": ["boolean", boolean];
  "question3": ["node[]", Iterable<Node>];
…
}
```
Note that using string litterals, who are the same as their type, makes this manipulation possible.

---
As a consequence and to be able to build the question shape generic, Questions now need to be registered by their URI before being used in rules.

This also let questions register a default message and ease a bit building of questions in rules. However, since the message is now optional, the order of parameters has been changed.

---
`Interview.conduct` now wraps answers in `question.answer`; therefore Oracles don't need to do it.
This allows to get rid of the free type variable in Oracles.

---
Small changes:
* SIA-R82 is now marked and exported as Experimental.
* Give explicit name to type variables in `alfa-act` + some doc.
* Remove the `"color"` type of questions, which was not used.

### DIFF
--- a/packages/alfa-rules/src/experimental.ts
+++ b/packages/alfa-rules/src/experimental.ts
@@ -1,3 +1,4 @@
 export { default as ER62 } from "./sia-er62/rule";
 export { default as ER87 } from "./sia-er87/rule";
+export { default as R82 } from "./sia-r82/rule";
 export { default as R109 } from "./sia-r109/rule";
--- a/docs/review/api/alfa-act.api.md
+++ b/docs/review/api/alfa-act.api.md
@@ -114,7 +114,7 @@ export type Interview<QUESTION, SUBJECT, CONTEXT, ANSWER, D extends number = 3> 
     [URI in keyof QUESTION]: Question<QUESTION[URI] extends [infer T, any] ? T : never, SUBJECT, CONTEXT, QUESTION[URI] extends [any, infer A] ? A : never, D extends -1 ? ANSWER : Interview<QUESTION, SUBJECT, CONTEXT, ANSWER, Depths[D]>, URI extends string ? URI : never>;
 }[keyof QUESTION];
 
-// @public
+// @public (undocumented)
 export namespace Interview {
     // (undocumented)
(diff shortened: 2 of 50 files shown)


packages/alfa-rules/src/experimental.ts
```diff
@@ -1,3 +1,4 @@
 export { default as ER62 } from "./sia-er62/rule";
 export { default as ER87 } from "./sia-er87/rule";
+export { default as R82 } from "./sia-r82/rule";
```
zsofitoth, line 3: Is this a breaking change or has `R82` not been used yet?

zsofitoth (approved):
Looks very good 👍

packages/alfa-rules/src/experimental.ts
Jym77: R82 is not used yet (because it asks questions about every form field, which is too much…)
(plus it already uses questions' context which Dory/Nemo/DB can't handle yet)

PR #1027, opened 2022-01-13
Replace Questions' message by Diagnostic

Jym77:
In `Question`, replace the `message` field by a full blown `Diagnostic`.

Questions now need to be built wrapping the message in `Diagnostic.of` and the message can be accessed by replacing `question.message` with `question.diagnostic.message`.


packages/alfa-rules/src/sia-r39/rule.ts
```diff
@@ -6,6 +6,7 @@ import { Criterion, Technique } from "@siteimprove/alfa-wcag";
 import { Page } from "@siteimprove/alfa-web";
 
 import { expectation } from "../common/expectation";
+import { normalize } from "../common/normalize";
```
zsofitoth, line 9: Is `normalize` used in this file? 👓

zsofitoth (approved):
LGTM 👍

PR #1031, opened 2022-01-20
SIA-ER62: Only apply to links that have different foreground from their container

### LINKED ISSUE #951: SIA-ER62 should only apply to links with colour different than their container
Otherwise, 1.4.1 doesn't apply.

- [ ] Follow-up on DEV-12438

zsofitoth:
Resolves #951


packages/alfa-rules/test/sia-er62/rule.spec.tsx
```diff
@@ -1070,229 +1225,255 @@ test(`evaluate() passes an <a> element that is distinguishable from the <p> pare
...
-      1: Outcomes.IsNotDistinguishable([noStyle], [noStyle], [noStyle]),
-    }),
-  ]);
+        h.rule.style("a", {
+          color: "rgba(255, 255, 255, 0.5)",
+          textDecoration: "none",
+          outline: "none",
+          cursor: "auto",
+        }),
+      ]),
```
zsofitoth, line 1412: This might be a bad test case because the link will inherit the `<p>` background, but I would expect it is still inapplicable - when inspecting the paragraph has 2 foreground colors computed (let's say `rgba(1,2,3,1) ` and `rgba(4,5,6,1) `) and the link has 4 foreground colors computed ( `rgba(1,2,3,1) `, `rgba(1,2,3,1) `, `rgba(4,5,6,1) ` and `rgba(4,5,6,1) `)

packages/alfa-rules/test/sia-er62/rule.spec.tsx
```diff
@@ -1070,229 +1225,255 @@ test(`evaluate() passes an <a> element that is distinguishable from the <p> pare
...
+test(`evaluate() is inapplicable to an <a> element with a <p> parent element
+    when the parent element has the text content wrapped in a <span> element and it has the same foreground color as the <a> element`, async (t) => {
   const target = <a href="#">Link</a>;
 
   const document = h.document(
-    [<p>Hello {target}</p>],
+    [
+      <p>
+        <span>Hello</span>
+        {target}
```
zsofitoth, line 1366: In this case since the `<p>`'s text content is wrapped into a `<span>`  I would only like to formulate something like:
```Javascript
if ( parentOfLink.hasNoDirectTextDescendant() && hasDifferentForeground(link, parentOfLink)) {
   containers = containers.set(link, parentOfLink);
   return yield link;
}
```
(lines 154-156 from the rules file)
where this bit `parentOfLink.hasNoDirectTextDescendant()` I might be able to use a common predicate to calculate?

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -115,7 +123,58 @@ export default Rule.Atomic.of<Page, Element>({
...
+          }
+        }
+
+        function* getApplicableLinks(): Iterable<Element> {
+          // Check if foreground is the same with the parent <p> element
+          const hasDifferentForeground = (
+            element1: Element<string>,
+            element2: Element<string>
+          ): boolean =>
+            getForeground(element1, device)
```
zsofitoth, line 136: Here I'm missing the `context` and it will fall back to the default defined in `get-colors`, is that expected?

packages/alfa-rules/test/sia-er62/rule.spec.tsx
```diff
@@ -1070,229 +1225,255 @@ test(`evaluate() passes an <a> element that is distinguishable from the <p> pare
...
+  const target = <a href="#">Link</a>;
+
+  const document = h.document(
+    [
+      <div>
+        <p>Hello</p>
+        <p>
+          <span>World</span>
+          Hello
+          <p>nested p</p>
```
Jym77, line 1333: Nested `<p>` are not allowed in the DOM. Not sure whether we enforce that during parsing (I guess not), but we should never have that as input (UA should fix the code and build a correct DOM, likely moving the nested `<p>` around).

(nested `<div>` or `<span>` work fine to test deep trees, of course)

packages/alfa-rules/test/sia-er62/rule.spec.tsx
```diff
@@ -1070,229 +1225,255 @@ test(`evaluate() passes an <a> element that is distinguishable from the <p> pare
...
+    failed(ER62, target, {
+      1: Outcomes.IsNotDistinguishable([noStyle], [noStyle], [noStyle]),
     }),
   ]);
 });
 
-test(`evaluate() passes an <a> element that is distinguishable from the <p> parent element as some foregrounds have a contrast of 3:1 in the parent element`, async (t) => {
+/******************************************************************
+ *
+ * Inapplicable tests
```
Jym77, line 1245: We should also have an applicable test where the link has the same colour as **part of** the container, but not all of it.

Jym77 (changes requested):
Spotted a few potential problems that may explain weirdness.
Plus possibilities for some streamlining and simplification of the Applicability.

packages/alfa-rules/test/sia-er62/rule.spec.tsx
Jym77: Because we need to draw the line somewhere 😅 and we need something to compare to.

---
[WCAG Success Criterion 1.4.1 Use of Color](https://www.w3.org/TR/WCAG21/#use-of-color) states:
> Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element.

So, for the rule to correctly map to it (i.e. to avoid false positives where the rule would fail but WCAG would be satisfied), we need to be (reasonably) "sure" that color is the **only** visual mean to convey the information "this is a link"…

There are cases, such as breadcrumbs or a site map where that info is conveyed just by being text in the breadcrumb, or list item in the site map. So we can't look at all links and flag all of them.

While I do agree that a `<div>`  works more or less like a `<p>` and we could flag these, we need to draw the line on the safe side. By restricting the rule to link inside paragraphs, we can be fairly sure that there is nothing more than style to convey the information "this is a link" (this excludes icons such as 🔗 that are sometimes used to prefix link, there is an assumption about that in the rule, we're not seeing too many complains about them, so the assumption seems to mostly hold).

So, we work in a frame where we can be confident we report true problems.

---
And of course, if there is no surrounding text at all, there is nothing to compare to 😃 

We could totally look at links inside `<div>`, but `<div>` may be used for more or less anything. Role of `paragraph` has the nice semantic linking, that these should only contain "block of text" where the rule make perfect sense.

---
The "no false positives" motto is putting strong constraints and often lead us to write weaker rules 😞

zsofitoth: Thank you, this explains it all :)

packages/alfa-rules/test/sia-er62/rule.spec.tsx
zsofitoth: The above is something I misread, I haven't realized it's no longer sufficient to compare all link colors with an opacity and a gradient background to the container color that has the same color. We could do it pairwise, but unfortunately that also doesn't guarantee success. For example these 2 colors are not the same:

`linear-gradient(to right, #0000D1 20%, #0000FF 100%)`
`linear-gradient(to right, #0000D1 50%, #0000FF 100%)`

and it could get more complicated with with different spreads and variations. I think we should just make the cases applicable with a gradient background (and foreground with opacity) as it seems quite rare and the expectation is quite lenient in terms of flagging it as a problem (there we compare every color to every other color regardless of spread etc)

Jym77: > I think we should just make the cases applicable with a gradient background (and foreground with opacity) as it seems quite rare and the expectation is quite lenient in terms of flagging it as a problem (there we compare every color to every other color regardless of spread etc)

Yes, I think that make sense. We'll revisit the decision if we get faced with such a case in real life 😄

packages/alfa-rules/test/sia-er62/rule.spec.tsx
```diff
@@ -587,6 +587,480 @@ test(`evaluate() passes an applicable <a> element that removes the default text
...
+  const document = h.document(
+    [
+      <div>
+        <p class="p-one">
+          Hello
+          {target}
+        </p>
+        <p class="p-two">
+          Hello
+          {target}
```
Jym77, line 978: We can't use the same element in two different places. Because nodes contain a parent pointer, and `target.parent` is now weird (and `alfa-dom` will actually build something different and not duplicate the element).

These should be two different `target1`, `target2` (and only one being applicable indeed).

Jym77 (changes requested):
Great work. Some small corrections.

packages/alfa-rules/src/common/expectation/get-colors.ts
```diff
@@ -232,62 +233,84 @@ export namespace Foreground {
...
-    background.map((backdrop) => Color.composite(color.get(), backdrop, 1))
-  );
+      // If the color is not transparent, and the element is fully opaque,
+      // then we do not need to dig further
+      if (color.get().alpha.value * opacity.value === 1) {
+        const result: Result<
+          Color[],
+          Layer.Error | Foreground.Error | Background.Error
+        > = Result.of([color.get()]);
+        return result;
```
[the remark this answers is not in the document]
zsofitoth: Yes, it seems just returning `Result.of([color.get()])` makes the compiler complain:

```
Argument of type '() => Result<Color[], unknown>' is not assignable to parameter of type
'Mapper<Cache<Element<string>, Result<Foreground, Layer.Error | Foreground.Error | Background.Error>>, Result<...>, []>'.
  Type 'Result<Color[], unknown>' is not assignable to type 'Result<Foreground, Layer.Error | Foreground.Error | Background.Error>'.
    Type 'unknown' is not assignable to type 'Layer.Error | Foreground.Error | Background.Error'.
      Type 'unknown' is not assignable to type 'Error'
```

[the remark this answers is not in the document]
Jym77: OK. 👍 
You can stil be a bit lighter by typing `of` directly:
`Result.of<Foreground, Layer.Error | Foreground.Error | Background.Error>([color.get()])`
https://github.com/Siteimprove/alfa/blob/main/packages/alfa-result/src/result.ts#L103-L105

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -69,60 +61,128 @@ export default Rule.Atomic.of<Page, Element>({
...
+                return true;
+              }
+              return getForeground(container, device).some(
+                (containerColors) => {
+                  // Similalry to the link, we assume it's applicable if the container has different foreground colors
+                  if (containerColors.length !== 1) {
+                    return true;
+                  }
+                  return !linkColors[0].equals(containerColors[0]);
+                }
```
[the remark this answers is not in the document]
zsofitoth: Is it possible that in this case the second `getForeground` should be negated like this:
```Javascript
getForeground(link, device).none(
  (linkColors) =>
	linkColors.length === 1 &&
	!getForeground(container, device).none(
	  (containerColors) =>
		containerColors.length === 1 &&
		linkColors[0].equals(containerColors[0])
	)
);
```
In this case maybe this version is more readable:
```Javascript
getForeground(link, device).some(
  (linkColors) =>
	linkColors.length !== 1 ||
	getForeground(container, device).none(
	  (containerColors) =>
		containerColors.length === 1 &&
		linkColors[0].equals(containerColors[0])
	)
);
```

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -69,60 +61,128 @@ export default Rule.Atomic.of<Page, Element>({
...
+              .get(containers.get(link).get())
+              .get();
+
+            for (const linkElement of linkTexts) {
+              for (const nonLinkElement of nonLinkTexts) {
+                if (hasDifferentForeground(linkElement, nonLinkElement)) {
+                  applicableContainers = applicableContainers.set(
+                    link,
+                    nonLinkElement
+                  );
```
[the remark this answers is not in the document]
zsofitoth: It seems if we include the extra values, then in some cases we will send extra contrast diagnostics, for example:
```Javascript
test(`evaluate() is applicable to an <a> element with a <p> parent element
    when at least one child textnode of the <p> element has a different foreground color from the <a> element`, async (t) => {
  const target = <a href="#">Link</a>;

  const document = h.document(
    [
      <p>
        <span>
          Hello
          <span>World</span>
        </span>
        {target}
        <strong>Differently colored text</strong>
      </p>,
    ],
    [
      h.sheet([
        h.rule.style("span", {
          color: "#0000EE",
        }),
      ]),
    ]
  );

  t.deepEqual(await evaluate(ER62, { document }), [
    passed(ER62, target, {
      1: Outcomes.IsDistinguishable([defaultStyle], [hoverStyle], [focusStyle]),
    }),
  ]);
});
``` 
Pairwise contrast diagnostics are generated then for the `<a>` and `<span>` comparison [here](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/sia-er62/rule.ts#L173-L190).

packages/alfa-rules/src/sia-er62/rule.ts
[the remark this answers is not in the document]
Jym77: > ```js
> getForeground(link, device).none(
>   (linkColors) =>
> 	linkColors.length === 1 &&
> 	getForeground(container, device).some(
> 	  (containerColors) =>
> 		containerColors.length === 1 &&
> 		linkColors[0].equals(containerColors[0])
> 	)
> );
> ```

This one should be good indeed.
The problem with the `foo.length !==1` tests is that they would pass if the length is 0 (not fully sure this does happen in practice), which would break the `foo[0]` later on.

packages/alfa-rules/src/sia-er62/rule.ts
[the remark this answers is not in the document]
Jym77: 🤔 That case shouldn't be affected by this change. Both `containers` and `applicableContainers` contain the `<a>` => `<p>` mapping in that case.

Now there is an interesting question of what the rule should do for
```html
<p>
Hello <a href="…" style="text-decoration: none">link</a> <strong style="color: "#0000EE">world</strong>
</p>
```
(or similar case) The link has different color from "hello" and is therefore applicable, but it has same styling as "hello"; and different styling from "world". Should it pass or fail?

I think this is again a borderline case that we want to accept just because there are way too many corner cases to be sure this would be bad (and I am not sure this would actually be bad for WCAG…)

So, I am OK with comparing the style of the link text with all the container text, including the ones that had the same color to start with.

---
The change would be for
```html
<p>Hello <a href="…" style="color: black">link</a> </p>
<p>World <a href="…">link</a> </p>
```
where the first link is not applicable (due to same color) and thus `container` should contain both `<a>` => `<p>` mappings, but `applicableContainers` only contain the second.
However, in that case, `nonLinkElements` (in the expectation) being built from `containers.get(target)` should never contain the first `<p>`.

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -69,54 +61,112 @@ export default Rule.Atomic.of<Page, Element>({
...
   evaluate({ device, document }) {
+    // Contains links (key) and their containing paragraph (value)
     let containers: Map<Element, Element> = Map.empty();
 
     return {
       applicability() {
-        return visit(document, None);
+        // Contains links (key) and the parents of the textnodes the links contain (value)
+        let linkText: Map<Element, Set<Element>> = Map.empty();
+        // Contains containers (key) without link descendants and the parents of the textnodes the containers have (value)
```
Jym77, line 71:
```suggestion
        // Contains containers (key) and the parents of the text nodes (non included in links) the containers have (value)
```

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -69,54 +61,112 @@ export default Rule.Atomic.of<Page, Element>({
...
-                return yield node;
-              }
+            const isLink = hasRole(device, (role) => role.is("link"));
+            const isParagraph = and(
+              hasRole(device, "paragraph"),
+              hasNonLinkText(device)
+            );
+
+            if (container.isSome() && isLink(node)) {
+              // For each link, know store its containing paragraph
```
Jym77, line 90:
```suggestion
              // For each link, store its containing paragraph
```

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -69,54 +61,112 @@ export default Rule.Atomic.of<Page, Element>({
...
+            if (link.isSome()) {
+              linkText = linkText.set(
+                link.get(),
+                linkText
+                  .get(link.get())
+                  .getOr(Set.empty<Element>())
+                  .add(parent.filter(isElement).get())
+              );
+            }
+
```
Jym77, line 115:
```suggestion
```

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -69,54 +61,112 @@ export default Rule.Atomic.of<Page, Element>({
...
-                node
-              )
-            ) {
+            if (isParagraph(node)) {
               container = Option.of(node);
             }
           }
 
+          const isTextNode = test(and(isText, isVisible(device)), node);
+          const parent = node.parent();
```
Jym77, line 103:
```suggestion
          const parent = node.parent().filter(isElement);
```
It will be safer to keep that as part of the guard and avoid the risk of getting from `none` later.

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -69,54 +61,112 @@ export default Rule.Atomic.of<Page, Element>({
...
+          if (isTextNode && container.isSome() && parent.isSome()) {
+            // For each link, store the parent of the text nodes it contains
+            if (link.isSome()) {
+              linkText = linkText.set(
+                link.get(),
+                linkText
+                  .get(link.get())
+                  .getOr(Set.empty<Element>())
+                  .add(parent.filter(isElement).get())
```
Jym77, line 112:
```suggestion
                  .add(parent.get())
```

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -69,54 +61,112 @@ export default Rule.Atomic.of<Page, Element>({
...
+            }
+
+            // For each container, store the parent of the text nodes it contains
+            else {
+              nonLinkText = nonLinkText.set(
+                container.get(),
+                nonLinkText
+                  .get(container.get())
+                  .getOr(Set.empty<Element>())
+                  .add(parent.filter(isElement).get())
```
Jym77, line 123:
```suggestion
                  .add(parent.get())
```

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -69,54 +61,112 @@ export default Rule.Atomic.of<Page, Element>({
...
+
+          for (const [link, linkTexts] of linkText) {
+            const nonLinkTexts = nonLinkText
+              .get(containers.get(link).get())
+              .get();
+
+            for (const linkElement of linkTexts) {
+              for (const nonLinkElement of nonLinkTexts) {
+                if (hasDifferentForeground(linkElement, nonLinkElement)) {
+                  return yield link;
```
Jym77, line 166: Shouldn't it be just
```suggestion
                  yield link;
```
🤔

Jym77 (changes requested):
Great 🎉 Some small clean up and polishing.

packages/alfa-rules/src/sia-er62/rule.ts
zsofitoth: I think we need the `return`, mostly because one of the tests fails when I remove it 😄 Originally, I thought it's for efficiency purposes and by returning we break out of the loop - we find the first match that is different in color and we won't look anymore. I don't quite understand yet why the test case fails though when I remove the `return` though

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -69,54 +61,111 @@ export default Rule.Atomic.of<Page, Element>({
...
```
Jym77, line 165:
```suggestion
            loop:
            for (const linkElement of linkTexts) {
              for (const nonLinkElement of nonLinkTexts) {
                if (hasDifferentForeground(linkElement, nonLinkElement)) {
                  yield link;
                  break loop;
```
should work.

or something in the line of
```suggestion
            if (linkTexts.some(linkElement =>
              nonLinkTexts.some(nonLinkElement =>
                hasDifferentForeground(linkElement, nonLinkElement))))
                  {yield link;}
```
The the lazyness of `.some` should take care of everything for us.

Jym77 (approved):
Looks great 💯 🎉

PR #1041, opened 2022-01-24
`isRendered`: Consider fallback content (descendant of media elements) as not rendered

### LINKED ISSUE #1017: `isVisible` should not consider descendants of media elements as visible
```html
<video src="foo">
  Hello
</video>
```
The text is only used for legacy browsers without support for `<video>` and isn't visible otherwise. Since all major browsers currently support the element, we should consider the text as invisible.
Other media elements are likely affected in the same way.

elenamongelli:
Resolves #1017

### DIFF
--- a/packages/alfa-rules/src/common/predicate/is-fallback.ts
+++ b/packages/alfa-rules/src/common/predicate/is-fallback.ts
@@ -6,17 +6,19 @@ const { not, or } = Predicate;
 const { and } = Refinement;
 const { isElement, hasName } = Element;
 
-// https://html.spec.whatwg.org/multipage/#the-video-element
-// https://html.spec.whatwg.org/multipage/#the-audio-element
+/**
+ * https://html.spec.whatwg.org/multipage/#the-video-element
+ * https://html.spec.whatwg.org/multipage/#the-audio-element
+ */
 
 // Children of <iframe>, <audio>, <video> elements act as fallback content in legacy user
-// agents and should therefore never be considered rendered.
-/**
- * 
- */
+// agents.
+
 export const isFallback = or(
   hasParentName("iframe"),
   and(
+    //<Track> and <source> children of audio and video are allowed by the content model of
+    // both. Thus, they will be not act as fallback
     hasParentName("audio", "video"),
     not(and(isElement, hasName("track", "source")))
   )
--- a/packages/alfa-rules/src/common/predicate/is-fallback.ts
+++ b/packages/alfa-rules/src/common/predicate/is-fallback.ts
@@ -1,3 +1,4 @@
+/// <reference lib="dom" />
 import { Element, Node } from "@siteimprove/alfa-dom";
 import { Predicate } from "@siteimprove/alfa-predicate";
 import { Refinement } from "@siteimprove/alfa-refinement";
--- a/packages/alfa-rules/src/common/predicate/is-fallback.ts
+++ b/packages/alfa-rules/src/common/predicate/is-fallback.ts
@@ -1,4 +1,3 @@
-/// <reference lib="dom" />
 import { Element, Node } from "@siteimprove/alfa-dom";
 import { Predicate } from "@siteimprove/alfa-predicate";
 import { Refinement } from "@siteimprove/alfa-refinement";
@@ -18,7 +17,7 @@ const { isElement, hasName } = Element;
 export const isFallback = or(
   hasParentName("iframe"),
   and(
-    //<Track> and <source> children of audio and video are allowed by the content model of
+    // <track> and <source> children of audio and video are allowed by the content model of
     // both. Thus, they will be not act as fallback
     hasParentName("audio", "video"),
     not(and(isElement, hasName("track", "source")))
--- a/packages/alfa-rules/src/common/predicate/is-fallback.ts
+++ b/packages/alfa-rules/src/common/predicate/is-fallback.ts
@@ -7,8 +7,8 @@ const { and } = Refinement;
 const { isElement, hasName } = Element;
 
 /**
- * https://html.spec.whatwg.org/multipage/#the-video-element
- * https://html.spec.whatwg.org/multipage/#the-audio-element
+ * {@link https://html.spec.whatwg.org/multipage/#the-video-element}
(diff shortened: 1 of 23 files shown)


packages/alfa-rules/src/common/predicate/is-fallback.ts
```diff
@@ -0,0 +1,34 @@
+/// <reference lib="dom" />
```
Jym77, line 1:
```suggestion
```

packages/alfa-rules/src/common/predicate/is-fallback.ts
```diff
@@ -0,0 +1,34 @@
...
+ * {@link https://html.spec.whatwg.org/multipage/#the-audio-element}
+ */
+
+// Children of <iframe>, <audio>, <video> elements act as fallback content in legacy user
+// agents.
+
+export const isFallback = or(
+  hasParentName("iframe"),
+  and(
+    //<Track> and <source> children of audio and video are allowed by the content model of
```
Jym77, line 21:
```suggestion
    // <track> and <source> children of audio and video are allowed by the content model of
```

packages/alfa-rules/src/common/predicate/is-rendered.ts
```diff
@@ -4,8 +4,7 @@ import { Element, Comment, Node } from "@siteimprove/alfa-dom";
 import { Predicate } from "@siteimprove/alfa-predicate";
 import { Context } from "@siteimprove/alfa-selector";
 import { Style } from "@siteimprove/alfa-style";
-
-const { isElement, hasName } = Element;
+import { isFallback } from "../predicate";
```
Jym77, line 7:
```suggestion

import { isFallback } from "./is-fallback.ts";
```
Here, we should get it from the individual file, to avoid the risk of circular import (`predicate.ts` does import `is-rendered.ts`…

packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -1,3 +1,4 @@
+/// <reference lib="dom" />
```
Jym77, line 1:
```suggestion
```

packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -42,7 +43,9 @@ function isInvisible(device: Device, context?: Context): Predicate<Node> {
       .get(node, () =>
         test(
           or(
-            not(isRendered(device, context)),
+            not(
+              isRendered(device, context)
+              ),
```
Jym77, line 48:
```suggestion
            not(isRendered(device, context)),
```
No reason to make it multiline 🤔

packages/alfa-rules/src/common/predicate/is-visible.ts
```diff
@@ -77,7 +80,7 @@ function isInvisible(device: Device, context?: Context): Predicate<Node> {
                 // If the element is replaced, visible when empty, or has set dimensions,
                 // it is assumed to be visible
                 nor(isReplaced, isVisibleWhenEmpty, hasDimensions(device)),
-                // otherwise, the element is invisible iff all its children are.
+                // otherwise, the element is invisible if all its children are.
```
Jym77, line 83:
```suggestion
                // otherwise, the element is invisible iff all its children are.
```
"iff" is a somewhat standard abbreviation for "if and only if"…

packages/alfa-rules/test/common/predicate/is-visible.spec.tsx
```diff
@@ -4,6 +4,7 @@ import { test } from "@siteimprove/alfa-test";
 import { Device } from "@siteimprove/alfa-device";
 
 import * as predicate from "../../../src/common/predicate/is-visible";
+import { isRendered } from "../../../src/common/predicate";
```
Jym77, line 7:
```suggestion
```
Doesn't seem to be used 🤔

packages/alfa-rules/test/common/predicate/is-visible.spec.tsx
```diff
@@ -24,6 +25,41 @@ test(`isVisible() returns false when an element is hidden using the \`hidden\`
...
+});
+
+test(`isVisible() returns false when a track element is a child of video`, (t) => {
+  const track = <track kind="description" />;
+  const element = <video src="foo.mp4">{track}</video>;
+
+  h.document([element]);
+
+  //<track> element is rendered in the browser, but isn't rendered by our user agent.
+  // Thus it is not visible and this test returns false
```
Jym77, line 44:
```suggestion
  // While <track> elements are not fallback, they are nonetheless invisible
```

packages/alfa-rules/test/common/predicate/is-visible.spec.tsx
```diff
@@ -24,6 +25,41 @@ test(`isVisible() returns false when an element is hidden using the \`hidden\`
...
+  t.equal(isVisible(track), false);
+});
+
+test(`isVisible() returns false when a div element is child of an iframe element`, (t) => {
+  const div = <div>hidden</div>;
+
+  const element = (
+    <iframe srcdoc="Hello">
+      {" "}
+      {div}{" "}
```
Jym77, line 55:
```suggestion
      {div}
```
Should be good.

Jym77 (dismissed):
Assorted cleanups 🧹

Jym77 (approved):

packages/alfa-rules/test/common/predicate/is-visible.spec.tsx
```diff
@@ -24,6 +24,39 @@ test(`isVisible() returns false when an element is hidden using the \`hidden\`
...
+
+test(`isVisible() returns false when a div element is child of an iframe element`, (t) => {
+  const div = <div>hidden</div>;
+
+  const element = (
+    <iframe srcdoc="Hello">
+      {div}
+    </iframe>
+  );
+  h.document([div]);
```
zsofitoth, line 55: Should this be like this? ⬇️ 
```suggestion
  h.document([element]);
```

Jym77 (approved):

PR #1045, opened 2022-01-26
Consider stretched elements as visible

### LINKED ISSUE #985: `isVisible` does not account for some "background-only" elements
```html
<style>
  .bg {
    position: absolute;
    top: 1px;
    bottom: 1px;
    left: 1px;
    right: 1px;
    background-color: blue;
  }
</style>
<div class="bg"> </div>
```
This is considered has not visible by `isVisible` (because it has no content) while it pretty much is (depending on the offset parent). In turn, this causes it to be not considered by `hasInterposedDescendants` while it is pretty much the kind of cases we want to catch 🙈 

- [ ] Follow up on DEV-12555

Jym77:
Resolves #985 
I needed it to streamline test cases for #1033 😄


zsofitoth (approved):
LGTM 👍

PR #1047, opened 2022-01-27
Added version tag and marked ER62 and ER87 as version 2

### LINKED ISSUE #1043: Add "versions" tags to experimental rules
Create a version tag (https://github.com/Siteimprove/alfa/tree/main/packages/alfa-rules/src/tags) that holds a number, and add it to the experimental rules that need to live next to the existing version of the rule (ER62, ER87).
These rules can also share the URI since they are now differentiated by version.
Rules with no version are implicitly version 1.

zsofitoth:
Resolves #1043

### DIFF
--- a/docs/review/api/alfa-rules.api.md
+++ b/docs/review/api/alfa-rules.api.md
@@ -596,6 +596,33 @@ export namespace Stability {
     const Stable: Stability<"stable">;
 }
 
+// @public (undocumented)
+export class Version<N extends number = number> extends Tag<"version"> {
+    // (undocumented)
+    equals(value: Version): boolean;
+    // (undocumented)
+    equals(value: unknown): value is this;
+    // (undocumented)
+    static of<N extends number>(version: N): Version<N>;
+    // (undocumented)
+    toJSON(): Version.JSON<N>;
+    // (undocumented)
+    get type(): "version";
+    // (undocumented)
+    get version(): N;
+}
+
+// @public (undocumented)
+export namespace Version {
+    // (undocumented)
+    export function isVersion<T extends string>(value: unknown, type?: T): value is Version;
+    // (undocumented)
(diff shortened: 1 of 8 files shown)


packages/alfa-rules/src/tags/version.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+/**
+ * @public
+ */
+export namespace Version {
+  export interface JSON<N extends number = number> extends Tag.JSON<"version"> {
+    version: N;
+  }
+}
```
Jym77, line 49: We should have a `isVersion` type predicate here, it can help for filtering later.

(we can also add one to `Scope` and `Stability` tags… not sure why they are missing there… I guess because scope and stability are essentially enumerated, but version is not and we might want to filter outcomes by `rule.hasTag(tag => Version.isVersion(tag) && tag.version > 3)`)

Jym77:
Also, you need to run `yarn document` and commit the changes to update the documentation. The CI doesn't like it otherwise.

docs/review/api/alfa-rules.api.md
```diff
@@ -576,7 +576,7 @@ export class Stability<S extends string = string> extends Tag<"stability"> {
     // (undocumented)
     equals(value: unknown): value is this;
     // (undocumented)
-    static of<S extends string>(stabiltiy: S): Stability<S>;
+    static of<S extends string>(stability: S): Stability<S>;
```
zsofitoth, line 579: It seems this is the only change to the documentation, nothing to the version. I wonder if I should export version similarly as it was done with `scope` and `stability` here: https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/index.ts#L6-L7

I'm thinking Dory would probably need it

Jym77 (approved):

PR #1050, opened 2022-01-28
Sub, sup elements are ignored

### LINKED ISSUE #971: SIA-R75 should not assess font size of superscript and subscript
Hello! I received another question about superscript and subscript HTML elements causing errors. 

This time the question is if it is reasonable to fail "Font size are not too small" when the font element in question is within a `<sup>` or `<sub>` HTML element. They are by nature smaller than the paragraph font size, but should we still require that textual content to be above 9px computed font size?

elenamongelli:
Resolves #971


zsofitoth (approved):
LGTM 👍

Jym77 (approved):

PR #1053, opened 2022-02-01
SIA-ER62: Tell in diagnostics all distinguishable properties

### LINKED ISSUE #1044: SIA-ER62: Tell in diagnostic which property is distinguishing
The current diagnostic sends all the (relevant) CSS properties that are different from their initial value so that the link can be displayed in a "look alike" version of the real one.

We want to expand the decorator to also say which of these properties are **different** from the surrounding container. So that
```html
<p style="font-weight: 700">
  Lorem <a href="…" style="text-decoration: underline; font-weight: 700">Link</a>
</p>
```
We send both boldness, color and underline to display the link currently, but we point that only `text-decoration` is good as distinguishing factor.

It might be easier to just add a new field to the diagnostic (rather than shoehorn it into the existing `styles` serialisation). This will likely require [`isDistinguishable`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/sia-er62/rule.ts#L381-L404) to stop being lazy 😖 so we can gather all the distinguishing factors rather than stopping at the first (should be OK because computed values of style are cached, so the extra computation should be low).

See also CE-5787

zsofitoth:
Resolves #1044 
Splits tests up for ER62 
Removes 2 duplicated tests (`evaluate() passes an <a> element in superscript` and `evaluates() passes on link with a different background-image than text`)


packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -383,26 +397,68 @@ function hasNonLinkText(device: Device): Predicate<Element> {
 namespace Distinguishable {
   export function isDistinguishable(
     container: Element,
+    target: Element,
     device: Device,
     context: Context = Context.empty()
-  ): Predicate<Element> {
-    return or(
+  ): Predicate<Element>[] {
+    let predicates: Array<
```
Jym77, line 404: We should keep the same order as they used to be (more or less).
There are two group of distinguishing factors: those that need to be different from the container, and those whose mere presence is enough. The new order is mixing these two groups.
We can still `.sort` the array immediately upon building, since the code doesn't care about the order.

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -383,26 +397,68 @@ function hasNonLinkText(device: Device): Predicate<Element> {
 namespace Distinguishable {
   export function isDistinguishable(
     container: Element,
+    target: Element,
     device: Device,
     context: Context = Context.empty()
-  ): Predicate<Element> {
-    return or(
+  ): Predicate<Element>[] {
```
Jym77, line 403:
```suggestion
  ): Array<Predicate<Element>> {
```

Jym77 (changes requested):
A bit of polishing.

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -237,14 +244,21 @@ export default Rule.Atomic.of<Page, Element>({
                 )
               );
 
+              const properties = distinguishingProperties
+                .get(context)
+                .getOr(Map.empty<Element, List<DistinguishingProperty>>())
+                .get(link)
+                .getOr(List.empty<DistinguishingProperty>());
```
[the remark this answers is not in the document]
zsofitoth: It looks like when it comes to evaluation of `List.empty` I get the error:
```
...alfa/packages/alfa-list/src/list.ts:39
    return this._empty;
                ^

TypeError: Cannot read properties of undefined (reading '_empty')
```

`.getOrElse(() => List.empty())` seems to work fine instead.

Jym77 (approved):

PR #1055, opened 2022-02-01
Grouping rules helpers

Jym77:
Moving stuff around to group helpers thematically, especially now that most are used both in Applicability and Expectation.


packages/alfa-rules/src/common/dom/get-offset-parent.ts
```diff
@@ -6,7 +6,9 @@ import { Predicate } from "@siteimprove/alfa-predicate";
 const { and, or, not, test } = Predicate;
 const { hasName, isElement } = Element;
 
-import { isPositioned, isRendered, isRoot } from "../predicate";
+import { isPositioned } from "../predicate/is-positioned";
+import { isRendered } from "../predicate/is-rendered";
+import { isRoot } from "../predicate/is-root";
```
zsofitoth, line 11: How come these imports are no longer grouped together?

Jym77: I want to avoid the risk of circular imports. Some predicates depend on DOM helpers (e.g. currnetly `hasInterposedDescendants` depends on `getOffsetParent`, and some DOM helpers depend on predicate (as here). By importing from `predicate.ts` we almost create an import loop; it seems to be OK by some JS magic only importing the bare minimum… But it felt incomfortable and I prefered breaking the loop from the start 😅

zsofitoth (approved):

PR #1063, opened 2022-02-08
SIA-R7: Applies to accessible names

### LINKED ISSUE #893: SIA-R7 should apply to `<body>` and accessible names
Following https://github.com/act-rules/act-rules.github.io/pull/1547 and https://github.com/Siteimprove/sanshikan/issues/203

Note that it already tries to ignores effectless attribute but still has problem with whitespace only text nodes: #892

elenamongelli:
`body` elements were already being evaluated by the rule, thus only accessible names have been added in this PR
Resolves #893

### DIFF
--- a/packages/alfa-rules/test/sia-r7/rule.spec.tsx
+++ b/packages/alfa-rules/test/sia-r7/rule.spec.tsx
@@ -102,7 +102,7 @@ test("evaluate() fails a text element with a lang attribute within <body> with a
   ]);
 });
 
-test("evaluate() fails an element with a lang attribute within <body> with an invalid value", async (t) => {
+test("evaluate() fails an element with an invalid lang attribute controlling an accessible name", async (t) => {
   const element = (
     <span lang="invalid">
       <img src="/test-assets/shared/fireworks.jpg" alt="Fireworks over Paris" />


packages/alfa-rules/src/sia-r7/rule.ts
```diff
@@ -46,22 +47,36 @@ export default Rule.Atomic.of<Page, Attribute>({
...
+                  hasAccessibleName(
+                    device,
+                    (accessibleName) => !isWhitespace(accessibleName.value)
+                  )
+                ),
+                node
+              )
+            ) {
+              yield * lang;
+            }
```
zsofitoth, line 78: Could this be rewritten to something like this?
```JavaScript
const isVisibleText = test(
  and(
    isText,
    and(
      or(isVisible(device), not(isIgnored(device))),
      (text: Text) => !isWhitespace(text.data)
    )
  ),
  node
);

const isElementWithAccessibleName = const isElementWithAccessibleName = test(
  and(
    isElement,
    hasAccessibleName(
      device,
      (accessibleName) => !isWhitespace(accessibleName.value)
    )
  ),
  node
); 

if(isVisibleText || isElementWithAccessibleName) {
  yield * lang;
}
```

Jym77: Ah yes, that would be even more readable 👍 
We can go more "predicate" by doing
```javascript
const isVisibleText = and(
    isText,
    and(
      or(isVisible(device), not(isIgnored(device))),
      (text: Text) => !isWhitespace(text.data)
    )
  );

const isElementWithAccessibleName = and(
    isElement,
    hasAccessibleName(
      device,
      (accessibleName) => !isWhitespace(accessibleName.value)
    )
  ) 

if(test(or(isVisibleText, isElementWithAccessibleName), node)) {
  yield* lang;
}
```

elenamongelli: good spot!  I have changed that

zsofitoth (approved):
LGTM 🚀

packages/alfa-rules/test/sia-r7/rule.spec.tsx
```diff
@@ -102,6 +102,38 @@ test("evaluate() fails an element with a lang attribute within <body> with an in
   ]);
 });
 
+test("evaluate() fails an element with a lang attribute within <body> with an invalid value", async (t) => {
```
Jym77, line 105:
```suggestion
test("evaluate() fails an element with an invalid lang attribute cortrolling an accessible name", async (t) => {
```

Jym77 (changes requested):
Nitpicking a test name.

Jym77 (approved):

PR #1065, opened 2022-02-08
SIA-R84: Doesn't apply to browsing context container

### LINKED ISSUE #979: SIA-R84 shouldn't apply to browsing context containers
Following the definition from ACT rules: https://act-rules.github.io/rules/0ssw9k#scrollable-element

Also investigate cases where the browsing context container is the only content of an element.

- [ ] Follow up on DEV-12500

elenamongelli:
Resolves #979

### DIFF
--- a/packages/alfa-rules/test/sia-r84/rule.spec.tsx
+++ b/packages/alfa-rules/test/sia-r84/rule.spec.tsx
@@ -121,7 +121,7 @@ test(`evaluate() is inapplicable to an element that restricts its width while
   t.deepEqual(await evaluate(R84, { document }), [inapplicable(R84)]);
 });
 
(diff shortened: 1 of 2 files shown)


packages/alfa-rules/test/sia-r84/rule.spec.tsx
```diff
@@ -120,3 +120,11 @@ test(`evaluate() is inapplicable to an element that restricts its width while
 
   t.deepEqual(await evaluate(R84, { document }), [inapplicable(R84)]);
 });
+
+test("evaluate() is inapplicable to a document with an <iframe> element", async (t) => {
```
Jym77, line 124:
```suggestion
test("evaluate() is inapplicable to a browsing context container", async (t) => {
```

Jym77 (changes requested):

PR #1067, opened 2022-02-09
Parsing TabIndex as HTML specification

### LINKED ISSUE #813: Parse integer according to HTML specification (i.e. ignores trailing non-digits)
As per specs, `tabindex` [must be a valid integer](https://html.spec.whatwg.org/multipage/interaction.html#attr-tabindex).

However, it seems that Chrome is a bit more lax in parsing it and parses `tabindex="-1-"` (extra dash) as "-1" while Alfa sees the extra dash as an error (according to specs) and default to 0.
tabindex

We should consider whether we want to stick with the specs or follow suit with browser (also investigate what other browsers are doing).

Reference: https://github.com/Siteimprove/alfa/blob/20b58b5e55271026eef7984b05c205512ccb18b6/packages/alfa-dom/src/node/element.ts#L228

- [ ] Follow up on DEV-11797

elenamongelli:
Resolves #813


packages/alfa-dom/src/node/element.ts
```diff
@@ -230,8 +230,10 @@ export class Element<N extends string = string>
    */
   public tabIndex(): Option<number> {
     for (const tabIndex of this.attribute("tabindex")) {
-      const number = Number(tabIndex.value);
 
+      const number = parseInt(tabIndex.value, 10);
+      
+      //Checking if tabindex isn't NaN and undefined
```
Jym77, line 236:
```suggestion
      //Checking if tabindex isn't NaN, undefined, null, Infinity
```

Jym77 (changes requested):
Also add test cases to https://github.com/Siteimprove/alfa/blob/main/packages/alfa-dom/test/node/element.spec.tsx

packages/alfa-dom/test/node/element.spec.tsx
```diff
@@ -4,6 +4,10 @@ test("#tabIndex() returns the tab index explicitly assigned to an element", (t)
   t.equal((<div tabindex="42" />).tabIndex().get(), 42);
 });
 
+test("#tabIndex() returns the non-valid integer tab index as parsed integer", (t) => {
```
Jym77, line 7:
```suggestion
test("#tabIndex() ignores characters after the first non-digit", (t) => {
```

Jym77 (changes requested):

Jym77 (approved):

PR #1068, opened 2022-02-09
SIA-R23, SIA-R24, SIA-R33: ask only for transcript link

### LINKED ISSUE #1059: SIA-R23, SIA-R24, SIA-R33: remove question about perceivability of the transcript on another page
After `transcript-link`, R24 and R33 are happy, but R23 asks `transcript-perceivable`.

R24 (and [the corresponding ACT rule](https://act-rules.github.io/rules/1a02b0)) does lack the part about the transcript being visible and accessible in its expectation, but I believe it is an overlook in ACT rules  (https://github.com/act-rules/act-rules.github.io/issues/1788)

R33 does contain that bit.

In any case, the `transcript-perceivable` question is confusing and should be integrated into the `transcript-link` question.

elenamongelli:
Resolves #1059


packages/alfa-rules/src/sia-r24/rule.ts
```diff
@@ -15,48 +13,15 @@ export default Rule.Atomic.of<Page, Element, Question.Metadata>({
   uri: "https://alfa.siteimprove.com/rules/sia-r24",
   requirements: [Criterion.of("1.2.8"), Technique.of("G69")],
   tags: [Scope.Component],
-  evaluate({ device, document }) {
+  evaluate: ({ device, document }) => {
```
Jym77, line 16:
```suggestion
  evaluate({ device, document }) {
```

Jym77 (changes requested):

Jym77 (approved):

PR #1073, opened 2022-02-14
Typescript version

Jym77:
* Update TS to 4.6.2.
* Flatten R82 to bound the interview's depth.
* Add depth type parameter to `expectation` helper to allow type hints from callers.


packages/alfa-rules/src/common/act/question.ts
```diff
@@ -242,14 +243,23 @@ export namespace Question {
...
+        "Which of these error indicators identifies, in text, the form field it relates to?",
     },
     "error-indicator-describes-resolution": {
-      type: "boolean",
+      type: "deprecated",
       message: `Does the error indicator describe, in text, the cause of the error or how to resolve it?`,
     },
     "error-indicator-identifies-form-field": {
-      type: "boolean",
+      type: "deprecated",
```
zsofitoth, line 262: I'm not sure I follow the deprecation related changes, are they happening as part of the general refactoring of `r82`? When writing the changelog, would there be 2 entry - one for the TS update and one for this deprecation part pointing to this PR?

zsofitoth (commented):
Looks good to me, just 2 questions

packages/alfa-rules/src/common/act/question.ts
zsofitoth: Now I actually see the label is `internal` so then none of these should affect the consumers?

Jym77: Yes, there needs to be an entry in the changelog with breaking changes for R82. The question deprecation is part of it.

Previously, R82 was asking for a list of all error indicators (related to a form field), then for each of these, asking whether it identifies the field (or describes the resolution). This creates an unbounded amount of questions since the number of indicators is also unbounded. The Interview type limits the number of questions via its Depth to avoid infinite recursion problems. 

So now R82 asks which of the error indicator identifies the field. This bounds the depth.
However, this is a different question, notably expecting a different type of answer. I didn't want to re-use the same URI since the expected type is different (better to break untyped oracles…) The old questions shouldn't be used, but I wasn't sure about fully deleting them either 🤔 

Maybe deleting them is better currently given that we don't have many consumers and the rule is clearly marked as experimental…

zsofitoth (approved):

PR #1077, opened 2022-02-16
SIA-R83: ignores text in `<option>` in single-line `<select>`

### LINKED ISSUE #984: SIA-R83 should ignore text in `<option>` in single-line `<select>`
```html
<style>
  select {width: 20px; overflow: hidden}
  option {width: 30px; overflow: hidden}
</style>
<select>
  <option value="Foo" selected>Toto</option>
  <option value="Bar">Titi</option>
  <option value="Baz">Super long text, I want that to be really huge</option>
</select>
```
On Chrome, the menu that opens ignores the `width` and is wide enough to show everything.

This gets different when a `size` attribute is added to the `select`. This is likely due to the difference between rendering it as a "list box" or "drop-down box". Need to investigate precise details. Text in `<option>` in drop-down box should most likely be ignored by the rule. The rest is less clear.

- [ ] Follow-up on DEV-12561

elenamongelli:
Resolves #984

### DIFF
--- a/packages/alfa-rules/test/sia-r83/rule.spec.tsx
+++ b/packages/alfa-rules/test/sia-r83/rule.spec.tsx
@@ -605,7 +605,7 @@ test(`evaluate() is inapplicable to a text node that is excluded from the
   t.deepEqual(await evaluate(R83, { document }), [inapplicable(R83)]);
 });
 
-test(`evaluate() ignores text in <option> into a single line <select> element`, async (t) => {
+test(`evaluate() is inapplicable to text in <option> within a single line <select> element`, async (t) => {
   const document = h.document(
     [
       <body>


packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -1,7 +1,13 @@
 import { Rule, Diagnostic } from "@siteimprove/alfa-act";
 import { Cache } from "@siteimprove/alfa-cache";
 import { Device } from "@siteimprove/alfa-device";
-import { Element, Text, Namespace, Node } from "@siteimprove/alfa-dom";
+import {
+  Element,
+  Text,
+  Namespace,
+  Node,
+  Attribute,
```
Jym77, line 9:
```suggestion
```

packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -104,6 +111,36 @@ export default Rule.Atomic.of<Page, Text>({
...
+      ) {
+        displaySize = valueMenuSize;
+      }
+    }
+
+    if (element.attribute("multiple").isSome()) {
+      displaySize = 4;
+    }
+
+    displaySize = 1;
```
Jym77, line 138: Now, that one is always executed (there is no `return` in the test) and erase the value 😓 
Easiest way would just be to initialise `displaySize` to 1 instead of 0 and remove this line.

packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -104,6 +111,36 @@ export default Rule.Atomic.of<Page, Text>({
...
+      const valueMenuSize = parseInt(size.value, 10);
+      if (
+        valueMenuSize === valueMenuSize &&
+        valueMenuSize === (valueMenuSize | 0)
+      ) {
+        displaySize = valueMenuSize;
+      }
+    }
+
+    if (element.attribute("multiple").isSome()) {
```
Jym77, line 134: `<select size="1" multiple">` get its `size` overwritten. So we need to only do that if no size was computed yet.
Maybe we can chain the option mapper 🤔 
```typescript
const displaySize = element.attribute("size")
  .map(/* parse and check the value*/).
  .getOrElse(() =>element
    .attribute("multiple")
    .map(() => 4)
    .getOr(1)
  )
```

packages/alfa-rules/test/sia-r83/rule.spec.tsx
```diff
@@ -503,81 +502,109 @@ test(`evaluate() fails a relatively positioned node clipped by a static parent`,
...
 
-  t.deepEqual(await evaluate(R83, { document }), [
-    passed(R83, target, { 1: Outcomes.WrapsText }),
-  ]);
+  t.deepEqual(await evaluate(R83, { document }), [inapplicable(R83)]);
 });
 
-test(`evaluates() passes a clipping element with font-relative width`, async (t) => {
-  const target = h.text("Hello World");
+test(`evaluate() ignores text in <option> into a single line <select> element`, async (t) => {
```
Jym77, line 546: Actually, we should have test with multi-line select (one with `size`, one with `multiple`), to make sure that `hasDisplaySize` is correct.

Jym77 (changes requested):

packages/alfa-rules/test/sia-r83/rule.spec.tsx
```diff
@@ -531,53 +285,387 @@ test(`evaluate() passes a relatively positioned node with a handling static pare
...
+    ]
+  );
+
+  t.deepEqual(await evaluate(R83, { document }), [
+    failed(R83, target1, { 1: Outcomes.ClipsText(Option.of(clipping), None) }),
+    passed(R83, target2, { 1: Outcomes.WrapsText }),
+  ]);
+});
+
+test(`evaluate() passes text in <option> into a non-single <select> element`, async (t) => {
```
Jym77, line 322:
```suggestion
test(`evaluate() passes text in <option> within a multi-line <select> element`, async (t) => {
```

packages/alfa-rules/test/sia-r83/rule.spec.tsx
```diff
@@ -531,53 +285,387 @@ test(`evaluate() passes a relatively positioned node with a handling static pare
...
+    passed(R83, target1, { 1: Outcomes.WrapsText }),
+    passed(R83, target2, { 1: Outcomes.WrapsText }),
+    passed(R83, target3, { 1: Outcomes.WrapsText }),
   ]);
 });
 
-test(`evaluates() passes a clipping element with font-relative width`, async (t) => {
-  const target = h.text("Hello World");
-
+test(`evaluate() passes text in <option> with multiple <select> element`, async (t) => {
```
Jym77, line 353:
```suggestion
test(`evaluate() passes text in <option> within a <select> with a \`multiple\` attribute`, async (t) => {
```

packages/alfa-rules/test/sia-r83/rule.spec.tsx
```diff
@@ -531,53 +285,387 @@ test(`evaluate() passes a relatively positioned node with a handling static pare
...
+          whiteSpace: "nowrap",
+        }),
+      ]),
+    ]
+  );
+
+  t.deepEqual(await evaluate(R83, { document }), [inapplicable(R83)]);
+});
+
+test(`evaluate() ignores text in <option> into a single line <select> element`, async (t) => {
```
Jym77, line 608:
```suggestion
test(`evaluate() is inapplicable to text in <option> within a single line <select> element`, async (t) => {
```

Jym77 (changes requested):
Nitpicking the test names.

PR #1078, opened 2022-02-16
SIA-ER62: Distinguish between link and container color name in diagnostics

### LINKED ISSUE #1066: SIA-ER62: Distinguish between link and container color name in diagnostics
Currently the `Contrast.Pairing` in the Diagnostics refers to the link and container color as 'background' and 'foreground', instead we should extend it to support different names

zsofitoth:
Resolves #1066


packages/alfa-rules/src/common/diagnostic/contrast.ts
```diff
@@ -125,9 +135,58 @@ export namespace Contrast {
...
       [key: string]: json.JSON;
-      foreground: RGB.JSON;
-      background: RGB.JSON;
+      color1: Color.JSON;
+      color2: Color.JSON;
       contrast: number;
     }
   }
+
+  type Name = "container" | "link" | "foreground" | "background";
```
Jym77, line 144:
```suggestion
  type Name = "container" | "link" | "foreground" | "background";
  
```

packages/alfa-rules/test/sia-er62/common.ts
```diff
@@ -22,6 +22,13 @@ export function addOutline(
   );
 }
 
+export function getContainerColor(color: RGB): Contrast.Color {
+  return Contrast.Color.of("container", color);
+}
+
+export function getLinkColor(color: RGB): Contrast.Color {
+  return Contrast.Color.of("link", color);
+}
```
Jym77, line 31: Since these are always used together, it might be worth doing a 
```typescript
function makePairing(container, link, contrast) {
return Pairing.of(Color.of("container", container), Color.of("link", link), contrast)
}
```
Will make call sites more compact.

(or it calls the new `Pairing.from(["container", container], ["link", link], contrast`)

packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -244,6 +244,11 @@ export default Rule.Atomic.of<Page, Element>({
                 )
               );
 
+              const sortedDistinguishableContrast = Array.sortWith(
+                [...distinguishableContrast],
+                (a, b) => a.compare(b)
+              );
+
```
Jym77, line 251: I would put the sorting directly in `ElementDistinguishrable.from` (and `ElementDistinguishable#withPairing`). This ensure the sorting is done all the time and notably we don't need to remember to sort the pairings in the tests.

packages/alfa-rules/src/common/diagnostic/contrast.ts
```diff
@@ -125,9 +135,58 @@ export namespace Contrast {
...
+  export class Color implements Equatable, Serializable, Hashable {
```
Jym77, line 145: This could have a `<N extends Name = Name>` type parameter, so that the type system knows which kind of colour we use. We could even go as far as adding a `<N extends ["container", "link"] | ["foreground", "background"]>` parameter to `Pairing` (and use types `N[0]`, `N[1]`) to ensure we only build meaningful pairings (and we unpack them correctly).

There could be 
```
Pairing.from<N extends ["container", "link"] | ["foreground", "background"]>(color1: [N[0], number], color2: [N[1], number], contrast: number) {
  return Pairing.of(Color.of(...color1), Color.of(...color2), contrast)
}
```

`ElementDistinguishable` could then make sure that it's pairing list is correctly typed (`Pairing<"container", "link">`).

Jym77 (changes requested):
Good stuff. Slightly nitpicking to have cleaner interface.

packages/alfa-rules/src/common/diagnostic/contrast.ts
```diff
@@ -7,22 +7,24 @@ import * as json from "@siteimprove/alfa-json";
 import { Hash, Hashable } from "@siteimprove/alfa-hash";
 import { Comparable, Comparison } from "@siteimprove/alfa-comparable";
 
+type Name = ["container", "link"] | ["foreground", "background"];
+
 export class Contrast extends Diagnostic {
```
Jym77, line 12: Let's add the `<N extends Name>` parameter on `Contrast` (so we can forward it to the `Pairing` inside).

To make `Color` cleaner, let's also add a `type FirstColor<N extends Name> = N[0]` (and `SecondColor`), the `Color` type parameter would then extend `FirstColor<Name> | SecondColor<Name>` rather than `Name` and the `N[0] | N[1]` ambiguity can be removed there. The `_color1`, `_color2` private variables (and associated getters, …) in `Pairing` can then be typed directly typed `FirstColor<N>`, … so that the type system is aware of the colors' names all along.

Doing it that way (rather than directly using `N[0]`) also makes it easy to change the representation of `Name` if we decide it's not the best, we only need to adapt `FirstColor` and the rest flows.

Jym77 (changes requested):
Looks good. We can polish a bit more the typing information.

Jym77 (approved):

packages/alfa-rules/src/common/diagnostic/contrast.ts
```diff
@@ -102,11 +102,11 @@ export namespace Contrast {
...
     }
 
-    public get color1(): Color<FirstColor<N>> {
+    public get color1(): Color<Name> {
       return this._color1;
     }
 
-    public get color2(): Color<SecondColor<N>> {
+    public get color2(): Color<Name> {
       return this._color2;
```
zsofitoth, line 110: I needed to revert this bit, because it looks like when the `Color` is tied to `FirstColor<N>` or `<SecondColor<N>>` it allows pairings like `container, background` and the build failed 😨

Jym77 (approved):

PR #1083, opened 2022-02-18
SIA-R54: Parsed `aria-atomic` case insensitively

elenamongelli:


Jym77:
(removing the resolevs link to keep the issue open since we still need to devise a global way to do it)

PR #1087, opened 2022-02-22
SIA-R18: Accepts attributes allowed by ARIA in HTML

### LINKED ISSUE #856: SIA-R18 should accept attributes allowed by ARIA in HTML
See https://github.com/Siteimprove/sanshikan/pull/185

https://www.w3.org/TR/html-aria/#docconformance

- [ ] Follow up on DEV-12322

---
- [ ] `<body>` accepts attributes allowed on `document` role;
- [ ] `<input>` with type `date`, `datetime-local`, `email`, `month`, `password`, `time`, `week`  accept attributes allowed on `textbox` role;
- [ ] `<select>` with no `multiple` and no `size` greater than 1 (i.e. display size of `1`) accepts attributes allowed on `combobox` and `menu` roles.
- [ ] `<video>` accepts attributes allowed on the `application` role.

elenamongelli:
Resolves #856


packages/alfa-rules/src/sia-r18/rule.ts
```diff
@@ -57,6 +63,51 @@ export default Rule.Atomic.of<Page, Attribute>({
   },
 });
 
+function ariaHtmlAllowed(target: Attribute): boolean {
+  const attributeName = target.name as aria.Attribute.Name;
+  switch (target.owner.get().name) {
```
Jym77, line 68:
```suggestion
  switch (target.owner.getOr("").name) {
```
Let's make this error-safe, if we somehow end up with an attribute with no parent this will trigger the default case and return false.

Actually, `target.owner.get()` is used in a couple other places (input and select), so we can refactor the whole as:
```typescript
for (const element of target.owner) {
  switch (element.name) {
    …
  }
}

return false;
```

packages/alfa-rules/src/sia-r18/rule.ts
```diff
@@ -57,6 +63,51 @@ export default Rule.Atomic.of<Page, Attribute>({
...
+          "week"
+        )(target.owner.get()) &&
+        Role.of("textbox").isAttributeSupported(
+          attributeName
+        )
+      );
+
+    case "select":
+      return (
+        (and(hasDisplaySize((size: Number) => size !== 1))(
```
Jym77, line 92:
```suggestion
        (hasDisplaySize((size: Number) => size !== 1)(
```

Jym77 (dismissed):
A bit of clean up.

packages/alfa-rules/src/sia-r18/rule.ts
```diff
@@ -65,47 +65,36 @@ export default Rule.Atomic.of<Page, Attribute>({
...
+        return (
+          hasInputType(
+            "date",
+            "datetime-local",
+            "email",
+            "month",
+            "password",
+            "time",
+            "week"
+          ) && Role.of("textbox").isAttributeSupported(attributeName)
```
Jym77, line 83:
```suggestion
          )(element) && Role.of("textbox").isAttributeSupported(attributeName)
```

packages/alfa-rules/src/sia-r18/rule.ts
```diff
@@ -65,47 +65,36 @@ export default Rule.Atomic.of<Page, Attribute>({
...
-          Role.of("combobox").isAttributeSupported(
-            attributeName
-          )) ||
-        Role.of("menu").isAttributeSupported(attributeName)
-      );
+      case "select":
+        return (
+          (hasDisplaySize((size: Number) => size !== 1) &&
+            Role.of("combobox").isAttributeSupported(attributeName)) ||
+          Role.of("menu").isAttributeSupported(attributeName)
```
Jym77, line 90:
```suggestion
          (hasDisplaySize((size: Number) => size !== 1)(element) &&
            (Role.of("combobox").isAttributeSupported(attributeName)) ||
          Role.of("menu").isAttributeSupported(attributeName))
```
(I'm never comfortable with logical operation precedence and prefer extra parenthesis)

Jym77 (changes requested):

PR #1098, opened 2022-03-02
Add diagnostics in `cantTell` outcomes

### LINKED ISSUE #1034: Add diagnostic in `cantTell` outcomes
In #1027 we've added diagnostic to questions, but the questions are not directly part of the result of the audit.
We should add diagnostic to the `cantTell` outcomes instead.
This probably requires to change `Interview` output from an `Option` to an `Either` (feels like a `Result` is not super well suited as no answer is not really an error, just another kind of result).

Then `evaluate` can evolve from turning `Some` into `Passed/Failed` and `None` into `CantTell` to branching on `Left/Right`; and in case of `CantTell`, include the data in the diagnostic.

`Interview.conduct` should probably just carry over the question's diagnostic to an outcome diagnostic. Alternatively, we could have the oracle produce an outcome diagnostic, but this seems ill-placed since the oracle comes from external source (users) and therefore doesn't know why Alfa cannot tell…

- [x] Add a diagnostic field in [`cantTell`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-act/src/outcome.ts#L388) like the expectations in other outcomes (no need to wrap in `Result` or `Record` here). This will break build until next point.
- [x] Update [`resolve`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-act/src/rule.ts#L411) to feed a dummy diagnostic when it create can't tell; update [`canttell`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/test/common/outcome.ts#L36) the same way; maybe there are a few other places where we build Can't Tell outcome… This should fix build form previous step.
- [x] Update [`Interview.conduct`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-act/src/interview.ts#L64) to produce a `Either<ANSWER, Diagnostic>`; use the question's own diagnostic to fill in the `Right` side of it (breaks stuff again)
- [x] Update [`resolve`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-act/src/rule.ts#L411) to handle the new output from `conduct` and feed it to the `CantTell` outcome.
- [x] Update [`canttell`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/test/common/outcome.ts#L36) to accept a diagnostic parameter; update all needed test cases to feed that 😖 Should use the [default question messages](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/common/question.ts#L101) as much as possible…

zsofitoth:
Resolves #1034

### DIFF
--- a/packages/alfa-act/src/diagnostic.ts
+++ b/packages/alfa-act/src/diagnostic.ts
@@ -48,6 +48,8 @@ export namespace Diagnostic {
   export function isDiagnostic(value: unknown): value is Diagnostic {
     return value instanceof Diagnostic;
   }
+
+  export const dummyDiagnostic = Diagnostic.of("dummy");
 }
 
 function normalize(input: string): string {
--- a/packages/alfa-act/src/diagnostic.ts
+++ b/packages/alfa-act/src/diagnostic.ts
@@ -49,7 +49,7 @@ export namespace Diagnostic {
     return value instanceof Diagnostic;
   }
 
-  export const dummyDiagnostic = Diagnostic.of("dummy");
+  export const defaultDiagnostic = Diagnostic.of("No extra information");
 }
 
 function normalize(input: string): string {
--- a/packages/alfa-act/src/diagnostic.ts
+++ b/packages/alfa-act/src/diagnostic.ts
@@ -49,7 +49,7 @@ export namespace Diagnostic {
     return value instanceof Diagnostic;
   }
 
-  export const defaultDiagnostic = Diagnostic.of("No extra information");
+  export const empty = Diagnostic.of("No extra information");
 }
 
 function normalize(input: string): string {
--- a/packages/alfa-act/src/rule.ts
+++ b/packages/alfa-act/src/rule.ts
@@ -4,7 +4,7 @@ import { Future } from "@siteimprove/alfa-future";
 import { Iterable } from "@siteimprove/alfa-iterable";
 import { Serializable } from "@siteimprove/alfa-json";
 import { List } from "@siteimprove/alfa-list";
-import { None, Option } from "@siteimprove/alfa-option";
+import { None, Option, Some } from "@siteimprove/alfa-option";
 import { Performance } from "@siteimprove/alfa-performance";
 import { Predicate } from "@siteimprove/alfa-predicate";
 import { Record } from "@siteimprove/alfa-record";
@@ -22,6 +22,7 @@ import { Oracle } from "./oracle";
 import { Outcome } from "./outcome";
 import { Requirement } from "./requirement";
 import { Tag } from "./tag";
+import { Either, Left, Right } from "@siteimprove/alfa-either";
 
 const { flatMap, flatten, reduce } = Iterable;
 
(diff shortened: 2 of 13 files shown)


Jym77:
Need to run `yarn install` and commit the changed `yarn.lock`.

packages/alfa-act/src/diagnostic.ts
```diff
@@ -48,6 +48,8 @@ export namespace Diagnostic {
   export function isDiagnostic(value: unknown): value is Diagnostic {
     return value instanceof Diagnostic;
   }
+
+  export const defaultDiagnostic = Diagnostic.of("No extra information");
```
Jym77, line 52:
```suggestion
  export const default = Diagnostic.of("No extra information");
```
It's already namespaced as `Diagnostic.default`, so we can keep a DRY name.

packages/alfa-act/src/rule.ts
```diff
@@ -257,9 +258,10 @@ export namespace Rule {
 
           return Future.traverse(applicability(), (interview) =>
             Interview.conduct(interview, this, oracle).map((target) =>
-              target.flatMap((target) =>
-                Option.isOption(target) ? target : Option.of(target)
-              )
+              target
+                .left()
+                // In case of a question we will drop that
```
Jym77, line 263:
```suggestion
                // If questions are left unanswered in Applicability,
                // we return an Inapplicable outcome; hence we can
                // just drop any Right that stays after conducting the
                // interview.
                .left()
```

packages/alfa-act/src/rule.ts
```diff
@@ -644,31 +646,50 @@ function resolve<I, T, Q, S>(
   rule: Rule<I, T, Q, S>,
   oracle: Oracle<I, T, Q, S>
 ): Future<Outcome.Applicable<I, T, Q, S>> {
+  function processExpectation(
```
Jym77, line 649: That one is complex enough that we can move it out of `resolve` and add a docstring to explain what it does, notably when the result is going to be a `Left` or a `Right`, and if it's a `Right`, which diagnostic it has (the one of the first `Right` in `acc`?)

packages/alfa-act/src/rule.ts
```diff
@@ -644,31 +646,50 @@ function resolve<I, T, Q, S>(
...
+    acc: Either<List<[string, Option<Result<Diagnostic>>]>, Diagnostic>,
+    [id, expectation]: readonly [
+      string,
+      Either<Option.Maybe<Result<Diagnostic, Diagnostic>>, Diagnostic>
```
Jym77, line 653:
```suggestion
      Either<Option.Maybe<Result<Diagnostic>>, Diagnostic>
```

Jym77 (changes requested):
Looks great 👍 
Some possible polishing and cleanup around `processExpectation`.

packages/alfa-act/src/diagnostic.ts
zsofitoth: Unfortunately, `default` seems to be reserved, would `defaultValue` be better?

Jym77: Ah yes 🙈 
`fallback`? `empty`?
`empty` sounds good and convey the idea: there is nothing to see here.

Jym77 (changes requested):

Jym77 (approved):

PR #1099, opened 2022-03-03
Simplify expectation helper

Jym77:
In #286, we allowed expectation to be `Option.Maybe` instead of just `Option` ([this change](https://github.com/Siteimprove/alfa/pull/286/files#diff-5e38584e1360a99b4db83cc874a7c2eb41f343d6b80484b462280cb95df8bc21L254)) This made `toExpectation` completely useless since it was doing this transformation, but we never cleaned it 🙈 

Note that now, the cases where the thunks are inexpensive (e.g. static) could be rewritten from `expectation(test, () => Outcomes.OK, () => Outcomes.KO)` to `test ? Outcomes.OK : Outcomes.KO`, but it probably still make sense to keep wrapping in the `expectation helper for the sake of keeping things similar, and of course there are cases where the thunks are expensive (notably if they contain questions) and thus should be frozen as continuations…


zsofitoth (approved):
LGTM 👍

PR #1102, opened 2022-03-04
Removes `aria-*` global deprecated attributes in 1.2

### LINKED ISSUE #947: Handle `aria-*` attributes that are no longer global in 1.2
Some global `aria-*` attributes have lost their globality in ARIA 1.2: https://www.w3.org/TR/wai-aria-1.2/#global_states

However, they are still listed in the global list, with a deprecation mention 🙈 This causes our script to incorrectly consider them as global and, among other, incorrectly trigger presentational role conflict resolution upon seeing them.

Grabbing aria attributes is done in https://github.com/Siteimprove/alfa/blob/main/packages/alfa-aria/scripts/attributes.js
List of allowed attributes built here: https://github.com/Siteimprove/alfa/blob/main/packages/alfa-aria/src/role.ts#L34-L42

- [ ] Follow-upo on DEV-12420.

elenamongelli:
Resolves #947


Jym77 (approved):
Well done 💯

PR #1108, opened 2022-03-08
SIA-R55, SIA-R56: Ignore `form` and `section` with no name 

### LINKED ISSUE #1079: SIA-R55, SIA-R56: Ignore `<from>` and `<section>` with no name.
Due to #298 / #307, we incorrectly compute the implicit role of `<form>` and `<section>` with no name.
R40 is handling `<section>` specifically, but R55 and R56 are not and thus create false positives.

We should quick-patch R55 and R56 the same way we did for R40 until we find a way to fix the root cause.

elenamongelli:
Resolves #1079


packages/alfa-rules/src/common/predicate/has-role-depending-on-name.ts
```diff
@@ -0,0 +1,16 @@
+import { Device } from "@siteimprove/alfa-device";
+import { Element } from "@siteimprove/alfa-dom";
+import { Predicate } from "@siteimprove/alfa-predicate";
+
+import { hasExplicitRole, hasAccessibleName } from "../predicate";
+
+const { hasName } = Element;
+const { and, not } = Predicate;
+
+export function hasRoleDependingOnName(device: Device): Predicate<Element> {
```
Jym77, line 10:
```suggestion
// <form> and <section> element have an implicit role only if they have an accessible name
// Alfa currently can't handle that and always give them a role.
// see https://github.com/Siteimprove/alfa/issues/298
//
// This detects the cases where such a role was incorrectly given.
export function hasRoleDependingOnName(device: Device): Predicate<Element> {
```

Jym77: Maybe let's make the name even more explicit like `hasIncorrectRoleWithoutName` 🤔

packages/alfa-rules/src/sia-r55/rule.ts
```diff
@@ -35,6 +39,7 @@ export default Rule.Atomic.of<Page, Group<Element>, Question.Metadata>({
                 hasRole(device, (role) => role.is("landmark"))
               )
             )
+            .reject(hasRoleDependingOnName(device))
```
Jym77, line 42:
```suggestion
            // circumventing https://github.com/Siteimprove/alfa/issues/298
            .reject(hasRoleDependingOnName(device))
```

packages/alfa-rules/src/sia-r56/rule.ts
```diff
@@ -35,6 +39,7 @@ export default Rule.Atomic.of<Page, Group<Element>>({
               hasRole(device, (role) => role.is("landmark"))
             )
           )
+          .reject(hasRoleDependingOnName(device))
```
Jym77, line 42:
```suggestion
            // circumventing https://github.com/Siteimprove/alfa/issues/298
          .reject(hasRoleDependingOnName(device))
```

Jym77 (changes requested):
R40 should be simplified the same way: replace `hasExplicitRole` with `hasRole`, and add a `.reject` afterwards. Plus clean up the long comment and only keep the short reference.
Both `.filter` in R40 can also be merged by using `Refinement.and` to combine the predicates.

Jym77 (approved):

PR #1109, opened 2022-03-08
SIA-R11: applies to role inheriting from `link`

### LINKED ISSUE #1103: SIA-R11 should apply to role inheriting from `link`
to follow ACT rules.
Essentially replace `hasRole(device, "link")` by ` hasRole(device, (role) => role.is("link")),` + add [test case](https://act-rules.github.io/rules/c487ae#failed-example-11)

elenamongelli:
Resolves #1103


packages/alfa-rules/test/sia-r11/rule.spec.tsx
```diff
@@ -16,32 +16,46 @@ test(`evaluate() passes a link with a name given by content`, async (t) => {
...
 });
 
-test(`evaluate() passes an image link with name given by the alt text`, async (t) => {
+test(`evaluate() fails an 'a' element, which inherits semantically a link role, with no accessible name`, async (t) => {
   const target = (
-    <a href="#">
-      <img src="foo.jpg" alt="Foo" />
+    <a href="#" role="I am giving a link role">
+      <img src="Foo.jpg" alt="" />
     </a>
```
Jym77, line 47: This should be an existing role 😅 
Also would be better with a non`<a>` element. Currently, the role is invalid, so it defaults to the implicit role of `link`, and this does not test the change.
```suggestion
    <span role="doc-biblioref">
      <img src="Foo.jpg" alt="" />
    </span>
```

Jym77 (changes requested):

packages/alfa-rules/test/sia-r11/rule.spec.tsx
```diff
@@ -16,32 +16,46 @@ test(`evaluate() passes a link with a name given by content`, async (t) => {
...
-  t.deepEqual(await evaluate(R11, { document }), [inapplicable(R11)]);
+  const document = h.document([target]);
+
+  t.deepEqual(await evaluate(R11, { document }), [
+    failed(R11, target, { 1: Outcomes.HasNoName }),
+  ]);
 });
 
-test(`evaluate() passes an image link with name given by the alt text`, async (t) => {
+test(`evaluate() fails a 'span' element with no accessible name`, async (t) => {
```
Jym77, line 43:
```suggestion
test(`evaluate() fails a `doc-biblioref` with no accessible name`, async (t) => {
```

Jym77 (changes requested):

Jym77 (approved):

PR #1110, opened 2022-03-09
Pre-parse media and import rules when building style tree

### LINKED ISSUE #1071: Pre-parse media queries when building style tree
#1051 was only doing half of the job 🙈 
`@import` rules should also have their queries pre-parsed, and when matching the queries to resolve the cascade, we should use the pre-parsed version instead of re-parse it.

---
- [x] Re-do what #1051 was doing in `alfa-dom` (+R44).
- [x] Do the similar thing for `@import` rules in `alfa-dom`.
- [x] In https://github.com/Siteimprove/alfa/blob/main/packages/alfa-cascade/src/selector-map.ts avoid calling `Media.parse` and instead use the pre-parsed value. (this should also allow to remove the dependence from `alfa-cascade` to `alfa-media`)

elenamongelli:
Resolves #1071

### DIFF
--- a/docs/review/api/alfa-dom.api.md
+++ b/docs/review/api/alfa-dom.api.md
@@ -8,6 +8,7 @@ import * as earl from '@siteimprove/alfa-earl';
 import { Equatable } from '@siteimprove/alfa-equatable';
 import { Iterable as Iterable_2 } from '@siteimprove/alfa-iterable';
 import * as json from '@siteimprove/alfa-json';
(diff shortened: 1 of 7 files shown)


Jym77 (approved):

Jym77:
Need to add `alfa-css` in `alfa-dom`'s package.json, plus run `yarn install` and commit `yarn.lock` (plus run `yarn document`).

Jym77 (approved):

PR #1115, opened 2022-03-21
Allow rules to ask optional questions by providing default answer

### LINKED ISSUE #1035: Consider adding `options` object to questions, e.g. for optional questions
`Question` could contain an object for some slightly different behaviour.
Most obvious case would be to have a `default` field that provides a default answer for the question. So, if the oracle doesn't provide any, `Interview.conduct` would use that one instead.
(e.g. R14 could optionally ask for non-text content to ignore)

This is different from rhetorical questions since rhetorical questions do not give the oracle a possibility to answer. This could possibly override rhetorical question (replacing `.answerIf(test, value)` by `if (test) {question.options.default = value}` but probably shouldn't since rhetorical questions are cases where Alfa knows the answer for sure and we do not want to give users possibility to override it.

---
- [x] Add an `Options<A>` interface in `alfa-act/src/question.ts` (in the `Question` namespace) with an optional `default` field of type `A`.
- [x] Allow an `Question.Options` object to be passed when building questions, defaults to empty object, use it to fill an field.
- [x] In `alfa-act/src/interview.ts`, update `conduct` to use the `default` field if the oracle provides no answer.
- [x] In `alfa-rules/src/common/act/question.ts`, allow to pass a `Question.Options` object.

Testing:
- [x] In `alfa-rules/src/common/act/question.ts`, add a question type for `Array<string>`. Add a question URI `visible-focus-classes` with answer of this type.
- [x] In R65, before comparison, asks for `visible-focus-classes` with default value `[]`, and accept elements matching one of these classes.
- [x] Test case.

Should be backward-compatible (thanks to default value), so no need to bump rule version number.

Jym77:
Resolves #1035


packages/alfa-rules/src/common/act/question.ts
```diff
@@ -18,6 +18,7 @@ export namespace Question {
     "node[]": Iterable<Node>;
```
zsofitoth, line 17: There is an unused import in this file (`Diagnostic`)

zsofitoth (commented):
It looks good, there is just one unused import

packages/alfa-rules/src/common/act/question.ts
Jym77: Was cleaned in https://github.com/Siteimprove/alfa/pull/1115/commits/392f5461dee6e6b9045097464cf5b49d306ef734

zsofitoth (approved):

Issue #955, opened 2021-10-19
SIA-R83 should accept `text-overflow` on distant ancestors of the text node.


Jym77:
In #924 we stop considering `white-space` and `text-overflow` on ancestors of the parent of the text node. It seems `text-overflow` still work higher up the tree and should be considered.
```html
<style>
  .clip { overflow: hidden}
  .ellipsis { text-overflow: ellipsis}
  .nowrap {white-space: nowrap}
</style>
<div class="clip ellipsis">
  <span class="nowrap">Hello World</span>
</div>
```

- [ ] Follow-up on DEV-12455

PR #1116, opened 2022-03-24
SIA-R83: Accept font relative media queries as handling overflow

### LINKED ISSUE #941: SIA-R83: Accept font-relative media queries as handling overflow
```css
@media (min-width: 100em) {...}
```

Maybe consider that elements whose style is set via a font-size relative media query definitely handle overflow, by assuming that if text becomes too big the element will be presented differently.

~Same goes for elements whose `width` is set via a font relative unit. These can be considered as handling horizontal overflow by growing with text. (the same we do for `height` in `hasFixedHeight`).~

- [ ] Follow-up on DEV-12394
- [ ] Follow-up on DEV-12556

Jym77:
Resolves #941


packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -214,6 +243,13 @@ function horizontallyClippingAncestor(
           return None;
         }
 
+        if (test(usesFontRelativeMediaRule(device, isWidth), element)) {
+          // The element uses a (font relative) media rule, and we can't guess what
+          // the page would like upon resizing and triggering a different media
+          // query, so we just accept it as good enough
+          return None;
+        }
```
zsofitoth, line 251: It looks like we are checking for a 2nd time - the check on lines 195-201 appears the same

Jym77: Yes, that's due to a specificity of horizontal wrapping which does not only depend on `overflow-x` but also on `text-overflow` and `white-space`. The later two sort of only work on the parent of the text (but not fully, #955 ).
So, we have a first `horizontallyClipper` on the parent of text, if text overflows its parent, we look for a `horizontallyClippingAncestor` considering that it is now content and not text (so `white-space` doesn't act anymore).
Both of these are quite similar (might be room for improvement) and both need to accept a font relative media query. The former calls the later which then recurse up the DOM tree.

PR #1117, opened 2022-03-25
Remove `alfa-protractor`

### LINKED ISSUE #1058: Consider deprecating `alfa-protractor`
Protractor is development [will stop at the end of 2022](https://github.com/angular/protractor/issues/5502), so we probably don't need to maintain an adaptation to it.

Jym77:
Resolves #1058 

With [`protractor` support ending soon](https://github.com/angular/protractor/issues/5502), we shouldn't keep maintaining the integration. It seems that Github packages do not support deprecation 😞 so we'll just remove the package and stop publishing new version. v0.35.0 will be the last published version of `alfa-protractor`.


zsofitoth (approved):
👍

Issue #995, opened 2021-11-30
Enumerated `aria-*` attributes should be parsed case-insensitively


Jym77:
According to [Mapping WAI-ARIA Value types to languages)(https://www.w3.org/TR/wai-aria-1.1/#typemapping), `aria-*` attribute of type `true/false`, `true/false/undefined` and `tristate` should be parsed as [enumerated attributes](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#keywords-and-enumerated-attributes), hence case-insensitively, which seems to not be the case currently 😕

Jym77:
Find list of affected rules.

PR #1121, opened 2022-03-31
Export helpers

Jym77:
Moves helpers and predicates from `alfa-rules` to `alfa-dom`, `alfa-style`, and `alfa-aria`.

Unfortunately it is not really possible in our setup to merge, e.g., `Element` namespace from the 3. This result in a somewhat unsatisfactory naming scheme 😕 

Some functions still feel like they are fairly specific to our rules and not really a common concept, so they are kept in `alfa-rules`.

Leftovers:
- `get/hasInterposedDescendants` => could go in `alfa-style`, but "interposed descendant" is not an official concept and pretty much our thing to circumvent problems around getting colours.
- `getColors` => very tailored to our specific need, even if the concepts are more generic.
- `isAtTheStart` => implements an ACT rules concept.
- `isLargeText` => implements a WCAG concept.
- `isWhitespace` => should go in an `alfa-text` package (together with all the copies of `normalize`).
- `isWideEnough` => pretty much a WCAG/ACT concept.
- `referenceSameResoure` => very tailored for the specific needs of our rules, could almost go in `common/expectation`…

And also `isProgrammaticallyHidden` and `isSemanticallyDisabled` technically only depend on `alfa-dom` concepts but are morally related to ARIA stuff, plus we should have proper parsing of `aria-*` attributes (#995) given that they are [fairly strongly typed](https://www.w3.org/TR/wai-aria-1.2/#propcharacteristic_value), in which case accessing them would depend on `alfa-aria`.


packages/alfa-rules/src/sia-r92/rule.ts
```diff
@@ -85,7 +80,7 @@ export default Rule.Atomic.of<Page, Element>({
                               wordSpacing.value >= 0.16 * fontSize.value
                           ),
                           not(
-                            hasCascadedValueDeclaredInInlineStyleOf(
+                            Style.hasCascadedValueDeclaredInInlineStyleOf(
```
zsofitoth, line 83: +1

packages/alfa-rules/src/sia-r93/rule.ts
```diff
@@ -97,7 +92,7 @@ export default Rule.Atomic.of<Page, Element>({
                             }
                           ),
                           not(
-                            hasCascadedValueDeclaredInInlineStyleOf(
+                            Style.hasCascadedValueDeclaredInInlineStyleOf(
```
zsofitoth, line 95: +1

packages/alfa-dom/src/node/traversal/get-nodes-between.ts
```diff
@@ -62,7 +63,11 @@ export function getNodesBetween(
 /**
  * Get all nodes between node1 and node2, included.
  */
-function getNodesInclusivelyBetween(node1: Node, node2: Node): Sequence<Node> {
+function getNodesInclusivelyBetween(
+  node1: Node,
+  node2: Node,
+  treeOptions: Node.Traversal
```
zsofitoth, line 69: Should this also have the default value `{ flattened: true, nested: true }`?

zsofitoth (commented):
It looks good, I found minor things where I thought there is a possibility for grouping - tried to mark all occurrences but there might be more I missed. If these are not possible to resolve, then just ignore comments otherwise I'll give another look to see if I can spot more.

`yarn document` seems necessary to be run again

packages/alfa-rules/src/sia-r39/rule.ts
```diff
@@ -36,7 +36,7 @@ export default Rule.Atomic.of<Page, Element, Question.Metadata>({
               not(isIgnored(device)),
               (element) =>
                 test(
-                  hasAccessibleName(device, (accessibleName) =>
+                  DOM.hasAccessibleName(device, (accessibleName) =>
```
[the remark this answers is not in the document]
Jym77: Yes, switched mind on import style mid-way and didn't fixed the old stuff. Thanks.

packages/alfa-dom/src/node/traversal/get-nodes-between.ts
Jym77: This is not needed because that function is not exported and only called from the "parent" `getNodesBetween`. If we need to promote it, it should have a default, but that shouldn't be necessarily given that the `includeOptions` of the "parent" would cover the functionality…

zsofitoth (dismissed):
Looks like `yarn document` is needed once more because of the last commit

zsofitoth (approved):

zsofitoth (approved):

PR #1130, opened 2022-04-21
`getColors`: ignore interposed elements which have a fully transparent background

### LINKED ISSUE #1118: R66/R69: consider skipping over interposed descendants with transparent background
In `getColors`, we bail out when we encounter an element with interposed descendants since we cannot know where exactly the interposed descendants will be located and whether there is text on top of them or not. This is mostly for handling the case of absolutely positioned images in carousels and similar situation.

However, we've also seen a lot of cases where absolutely positioned elements have been use for text or empty elements to, e.g. place a "follow us" link on the bottom right corner of a twitter feed, … In these case, not only the absolute position is **not** use to overlap elements, but the absolutely positioned element is mostly transparent with some text and doesn't affect colour contrast (especially since the rules only look at the highest contrast).

We could instead, when encountering interposed descendants, check if their background is fully transparent and simply ignore them in that case. Trying to compute their background colour and add it to the layer list is much more tricky due to not being clear at which layer they should be added, plus the possibility of semi-opaque stuff at unknown positions 🙈 We may keep that for a later time…

We need to take care that it can be a descendant of the absolutely positioned element which is not transparent (e.g. `<div style="position: absolute"></div>`, so we need to go down the subtree and search for transparency.

elenamongelli:
Resolves #1118


packages/alfa-style/src/element/predicate/has-transparent-background.ts
```diff
@@ -0,0 +1,64 @@
...
+): Predicate<Element> {
+  return (element) =>
+    cache
+      .get(device, Cache.empty)
+      .get(context, Cache.empty)
+      .get(element, () => {
+        if (
+          test(
+            or(
+              isReplaced,
```
Jym77, line 32: Let's add some comments.
```suggestion
              // Replaced elements are assumed to be replaced by non-transparent content.
              isReplaced,
```

packages/alfa-style/src/element/predicate/has-transparent-background.ts
```diff
@@ -0,0 +1,64 @@
...
+              Style.hasComputedStyle(
```
Jym77, line 33:
```suggestion
              // Elements with non-transparent background are not transparent.
              Style.hasComputedStyle(
```

packages/alfa-style/src/element/predicate/has-transparent-background.ts
```diff
@@ -0,0 +1,64 @@
...
+                "background-color",
+                (color) => !Color.isTransparent(color),
+                device,
+                context
+              ),
+              Style.hasComputedStyle(
```
Jym77, line 39:
```suggestion
              // Elements with a background image are not transparent.
              Style.hasComputedStyle(
```

packages/alfa-style/src/element/predicate/has-transparent-background.ts
```diff
@@ -0,0 +1,64 @@
...
+                context
+              )
+            ),
+            element
+          )
+        ) {
+          return false;
+        }
+
+        return element
```
Jym77, line 56:
```suggestion
        // The element itself has transparent background, but it may have non-transparent content.
        return element
```

packages/alfa-style/src/element/predicate/has-transparent-background.ts
```diff
@@ -0,0 +1,64 @@
...
+
+
+const { or, test } = Predicate;
+
+const cache = Cache.empty<Device, Cache<Context, Cache<Element, boolean>>>();
+
+/**
+ * @public
+ */
+
```
Jym77, line 19:
```suggestion
/**
 * Checks if an element has transparent background.
 * The element may be not fully transparent (e.g., have text) while still having transparent background.
 * @public
 */
```

packages/alfa-style/test/node/predicate/has-transparent-background.spec.tsx
```diff
@@ -0,0 +1,77 @@
+import { h } from "@siteimprove/alfa-dom";
+import { test } from "@siteimprove/alfa-test";
+
+import { Device } from "@siteimprove/alfa-device";
+
+import { hasTransparentBackground } from "../../../src/element/predicate/has-transparent-background";
+
+const device = Device.standard();
+
+test("hasTransparentBackground() fails on image which is not transparent", (t) => {
```
Jym77, line 10:
```suggestion
test("hasTransparentBackground() returns false on replaced elements", (t) => {
```

packages/alfa-style/test/node/predicate/has-transparent-background.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const target = <img src="foo.jpg" />;
+
+  h.document([target]);
+
+  t.deepEqual(hasTransparentBackground(device)(target), false);
+});
+
+test("<div> has a color background", (t) => {
```
Jym77, line 18:
```suggestion
test("hasTransparentBackground() returns false on element with a background color", (t) => {
```
Test names should be meaningful without context. If the test fails, this name is the first thing that will be printed on the terminal, so the first thing we read to figure out where (and why) the test fails.
* frontloading the tested function helps locating the problem ("<div> has a color background" doesn't help understanding in which file this is, or which function we tested).
* Trimming information to relevant level (It doesn't matter that this is a `<div>` and not a `<span>`, it only matters that it is an element).
* The predicate itself doesn't pass or fail. It returns `true` or `false`, but that is neither a success or failure in itself.

packages/alfa-style/test/node/predicate/has-transparent-background.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+  const target = <div style={{ backgroundColor: "red" }}> Hello</div>;
+
+  h.document([target]);
+
+  t.deepEqual(hasTransparentBackground(device)(target), false);
+});
+
+test("<div> has an image background", (t) => {
```
Jym77, line 26:
```suggestion
test("hasTransparentBackground() returns false on an element with a background image", (t) => {
```

packages/alfa-style/test/node/predicate/has-transparent-background.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      Hello
+    </div>
+  );
+
+  h.document([target]);
+
+  t.deepEqual(hasTransparentBackground(device)(target), false);
+});
+
+test("<div> background is not transparent when one child background isn't", (t) => {
```
Jym77, line 43:
```suggestion
test("hasTransparentBackground() returns false on an element with non-transparent child", (t) => {
```

packages/alfa-style/test/node/predicate/has-transparent-background.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <span style={{ backgroundColor: "#005BBB" }}> Third</span>
+    </div>
+  );
+
+  h.document([target]);
+
+  t.deepEqual(hasTransparentBackground(device)(target), false);
+});
+
+test("<div> background is transparent when all children backgrounds are transparent", (t) => {
```
Jym77, line 57:
```suggestion
test("hasTransparentBackground() returns true on an element whose children all have transparent background", (t) => {
```

packages/alfa-style/test/node/predicate/has-transparent-background.spec.tsx
```diff
@@ -0,0 +1,77 @@
...
+      <span> Third</span>
+    </div>
+  );
+
+  h.document([target]);
+
+  t.deepEqual(hasTransparentBackground(device)(target), true);
+});
+
+test("<div> background is transparent", (t) => {
```
Jym77, line 71:
```suggestion
test("hasTransparentBackground() returns true on elements without background colour nor image", (t) => {
```

Jym77 (dismissed):

Jym77 (dismissed):

Jym77 (approved):

PR #1134, opened 2022-05-02
SIA-ER62: `background-image` should ignore transparent link element

### LINKED ISSUE #1131: R62: Comparing `background-image` should ignore transparent link element
We currently accept any difference in `background-image` between link and container, which is incorrect since a fully transparent `background` on the link results in no difference, even if the `background` image is technically different (between some image and `none`).

We should add the same logic of "if the link's background is transparent, `background-image` is not a distinguishing factor" as what happens when comparing `background-color`.

- [ ] Follow up on CE-5894

zsofitoth:
Resolves #1131

### DIFF
--- a/packages/alfa-rules/src/sia-er62/rule.ts
+++ b/packages/alfa-rules/src/sia-er62/rule.ts
@@ -16,6 +16,7 @@ import { Context } from "@siteimprove/alfa-selector";
 import { Sequence } from "@siteimprove/alfa-sequence";
 import { Set } from "@siteimprove/alfa-set";
 import { Style } from "@siteimprove/alfa-style";
+import { Computed as BackgroundImage } from "@siteimprove/alfa-style/src/property/background-image";
 import { Criterion } from "@siteimprove/alfa-wcag";
 import { Page } from "@siteimprove/alfa-web";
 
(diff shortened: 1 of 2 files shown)


packages/alfa-rules/src/sia-er62/rule.ts
```diff
@@ -522,9 +522,18 @@ namespace Distinguishable {
...
       hasComputedStyle(
         "background-image",
-        not((image) => image.equals(imageReference)),
+        not((image) => {
+          let hasNoBackgroundImage: boolean = false;
+          for (const value of image.values) {
+            if (Keyword.isKeyword(value) && value.equals(Keyword.of("none"))) {
+              hasNoBackgroundImage = true;
+            }
+          }
```
zsofitoth, line 534: This is not super elegant, but it turns out when it comes to the `background-image` property, transparent means that the value of the css property is `none`.

Jym77: 🤔 I feel we should merge both transparent tests. Something like `hasNoBackground = [color: transparent] and [image: none]` . That will make the logic more apparent and a bit safer.

Jym77 (changes requested):

Jym77 (approved):

PR #1135, opened 2022-05-03
Change :focus-visible to match :focus elements

Jym77:
We incorrectly understood that `:focus-visible` was an addition to `:focus`, and that it would only apply in a combination `:focus:focus-visible`.
It turns out that `:focus-visible` already matches a subset of `:focus` (based on UA choices).

This PR changes the `matches` of `FocusVisible` from "always" to "same as `Focus`" to reflect that.


zsofitoth (approved):
LGTM, is this a `patch` though? Seems like there is a change to the `alfa-selector` api and could qualify as `minor`.

PR #1141, opened 2022-05-10
`getColors`: ignores transparent text-shadow

### LINKED ISSUE #1120: R66/R69: run some basic test on text shadows
We are currently [bailing out when encountering a `text-shadow`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/common/dom/get-colors.ts#L372-L375).
We should probably be able to run some basic check on its existence (transparent colour, 0 spread, …)

We can likely use [isVisibleShadow](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/common/predicate/has-box-shadow.ts#L33-L50)

- [ ] Follow-up on DEV-12979

elenamongelli:
Resolves #1120


packages/alfa-rules/src/common/dom/get-colors.ts
```diff
@@ -371,8 +371,8 @@ export function getBackground(
     .get(device, Cache.empty)
     .get(context, Cache.empty)
     .get(element, () => {
-      // If the element has a text-shadow, we don't try to guess how it looks.
-      if (textShadow.type !== "keyword") {
+      // We ignore invisible text-shadow
```
Jym77, line 374:
```suggestion
      // If the element has a visible text-shadow, we don't try to guess how it looks.
```

Jym77 (changes requested):

Jym77 (approved):

PR #1151, opened 2022-05-23
Do not expose elements with a non-rendered ancestor, even if referenced by `aria-owns`

### LINKED ISSUE #1111: Hidden elements should not be included in the accessibility tree through `aria-owns`
```html
<div>
  <div style="display: none">
    <div id="target" role="listitem">Lorem ipsum…</div>
  </div>
  <div aria-owns="target" role="list"></div>
</div>
```

The `target` element is not rendered and therefore should be [excluded from the accessibility tree](https://www.w3.org/TR/wai-aria-1.2/#tree_exclusion). However, Alfa still includes it because it does not check tree exclusion rules on ancestors when [following `aria-owns`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-aria/src/node.ts#L408-L410) 🙈 

Note that `aria-hidden` on the container (rather than `display: none`) would not exclude the target since `aria-hidden` only affects descendants in the accessibility tree…

Jym77:
Resolves #1111

### DIFF
--- a/packages/alfa-aria/src/node.ts
+++ b/packages/alfa-aria/src/node.ts
@@ -413,6 +413,7 @@ export namespace Node {
         // different from the DOM tree, but being rendered is a property of the
         // DOM, we may "jump" onto a node which is not rendered due to some DOM
         // ancestor (so, unknowingly of the current accessibility tree traversal).
+        // Therefore, we cannot just look at some property of the current node.
         //
         // Since `isRendered` is cached, and evaluating it is needed for almost
         // all nodes in the DOM, this is inexpensive.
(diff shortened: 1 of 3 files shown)


packages/alfa-aria/src/node.ts
```diff
@@ -401,59 +406,51 @@ export namespace Node {
...
         }
 
         let children: (state: State) => Iterable<Node>;
 
-        // Children of <iframe> elements act as fallback content in legacy user
-        // agents and should therefore never be included in the accessibility
-        // tree.
-        if (node.name === "iframe") {
-          children = () => [];
-        }
```
Jym77, line 426: Since `isRendered` already includes `isFallback`, this test is not needed anymore.
This does change result a bit by returning a list of `Inert` children (rather than an empty one), which is likely OK.

elenamongelli (commented):
Rule tests are passing and looks a good clean

zsofitoth (approved):

zsofitoth:
> Rule tests are passing and looks a good clean

+1️⃣

PR #1153, opened 2022-05-31
Refactor DOM and ARIA tree in abstract tree structure, plus cache some traversal operations

### LINKED ISSUE #1144: Investigate caching DOM tree traversal operations
Since #283 (see discussion in #204), DOM nodes are "free floating" until they are frozen, e.g. by attaching them to parent. This notably ensure that **sub**trees never change. However, **super**trees may change by attaching the (current) root into another node.
As a consequence, tree traversal functions cannot be changed because `node.ancestor` can change. However, several of these are fairly costly and called often…

However, in our most frequent use case, the DOM tree does not change (it is just read from the JSON). So we could introduce a "global frozen" state and cache traversal functions when the node is part of a globally frozen tree. Globally frozen could either be stored on each node, or just be equivalent to `node.root().frozen`. (in turn, `node.root()` would need to be always cached to be fast, but can change 😖 Double rooting could work (`node.root() = {node._root = node._root.root(); return node._root}`) and be fast enough 🤔 ).

Care must be taken for the traversal options in the cache. See #1132 (and `.path()` should also hook onto the globally frozen state).
When building a `Document` from JSON, it would be globally frozen (or maybe only as part of a `Page.from()` build…) Leave an option to build an non-frozen one…

We've seen a page with ~9000 CSS rules about `a` name (!), including ~4000 with sibling selectors (`div + a`).
Matching these requires to find the preceding element, which in turn requires the list of sibling which can be a bit expensive… and is done 4000 times for each link on the page 🙈

Jym77:
Resolves #1144 
Replaces #1152 

* Introduces an `alfa-flags` package to model sets of boolean flags (stored as bits in a number).
* Introduce an `alfa-tree` package to model n-ary trees, with optional traversal flags.
* Use `alfa-tree` as a base for DOM and ARIA trees.


zsofitoth (approved):
It looks good to me 👍

zsofitoth (approved):

Jym77:
Had to keep API extractor to a previous version due to https://github.com/cypress-io/cypress/issues/22417 🙃

zsofitoth (approved):

Issue #1036, opened 2022-01-21
Override factory methods `.of` with extra mandatory parameters in subclasses.


Jym77:
This is blocked until https://github.com/microsoft/TypeScript/issues/4628 / https://github.com/microsoft/TypeScript/pull/39699 has landed.

We make a heavy use of factory method `public static of()` in our classes. Unfortunately, TS doesn't let us change the signature in derived classes, even through our case is fairly safe. This has been a very discussed topic in the TS community, we should keep an eye on it and streamline code where possible.

Currently, instead of using:
```javascript
class Foo {
  public static of(x: number): Foo { … }
}

class Bar extends Foo {
  // Rejected by TS due to incompatible signature of static function.
  public static of(x: number, y: number): Bar { … }
}
```
We use patterns like:
```javascript
class Foo {
  public static of(x: number): Foo { … }
}

class Bar extends Foo {
  public static of(x: number, y: Option<number> = None): Bar { … }
}
```
In turn, this forces instances of `Bar` to store `y` as an option (or to have `Bar.of` possibly erroring at runtime without clean way of detecting cases where we don't provide `y`). Further down, users of `Bar` need some extra `isSome` checks when getting `y`, which clutters the full construction.

A notable case for this is extended `Diagnostic` where all the extra data needs to be "optional" due to this. Sometimes we can provide a decent default (e.g. empty array), but in several cases, we need to wrap it in `Option` even though we know there will always be something.

PR #1164, opened 2022-06-20
SIA-R65: Provide classes and number of matches in diagnostic

### LINKED ISSUE #1161: SIA-R65: provide classes and number of matches in diagnostic
Build extended diagnostic with all classes found on target + number of matches on the page for each of these classes.
For each found class: number of matching targets, and number of matching non-target.

See CE-5935

### LINKED ISSUE #1162: SIA-ER65: Split result to show why we pass
SIA-R65 can pass for several reasons (auto found, told so, good class). We should split the outcome to tell which one.

Create an extended diagnostic with a reason why we pass the rule (include which class(es) have been found as good, if any).

zsofitoth:
Resolves #1161


packages/alfa-rules/src/sia-er65/diagnostics.ts
```diff
@@ -0,0 +1,60 @@
+import { Diagnostic } from "@siteimprove/alfa-act";
+
+export type Reason = "auto-detected" | "user-answered" | "good-class";
```
Jym77, line 3: I think it's better to use an enum for cases like that (which we don't always do 😞) This provides safer typing.
```suggestion
export enum Reason = 
  | AutoDetected = "auto-detected" 
  | UserAnswered = "user-answered" 
  | GoodClass = "good-class";
```

packages/alfa-rules/src/sia-er65/diagnostics.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * @internal
+ */
+export class ExtendedDiagnostic extends Diagnostic {
+  public static of(
```
Jym77, line 9: To make the reason not (really) an optional argument, we can use an overload and default to `Diagnostic` when none is provided. See #1036.

packages/alfa-rules/src/sia-er65/diagnostics.ts
```diff
@@ -0,0 +1,60 @@
...
+      reason: this._reason,
+    };
+  }
+}
+
+/**
+ * @internal
+ */
+export namespace ExtendedDiagnostic {
+  export interface JSON extends Diagnostic.JSON {}
```
Jym77, line 45:
```suggestion
  export interface JSON extends Diagnostic.JSON {
    reason: string;
  }
```

packages/alfa-rules/src/sia-er65/rule.ts
```diff
@@ -59,6 +61,9 @@ export default Rule.Atomic.of<
         );
 
         const askFocusIndicator = Question.of("has-focus-indicator", target);
+        const reason: Reason = hasFocusIndicator(device)(target)
```
Jym77, line 64: We can store the result of `hasFocusIndicator` to avoid recomputing it.

Jym77 (changes requested):

packages/alfa-rules/src/sia-r65/rule.ts
```diff
@@ -25,10 +28,53 @@ export default Rule.Atomic.of<Page, Element, Question.Metadata>({
...
+        .get(classname)
+        .getOr({ matchingTargets: 0, matchingNonTargets: 0 });
+
+      const matchingTargets = isTarget
+        ? matches.matchingTargets + 1
+        : matches.matchingTargets;
+
+      const matchingNonTargets = isTarget
+        ? matches.matchingNonTargets
+        : matches.matchingNonTargets + 1;
```
Jym77, line 60: I feel this counting logic could be located with the data definition.

* `Matches` could be a full class with methods to increase number of targets or non targets, then we wouldn't need the logic to redefine new constants here.
* There could even be a wrapper class for `Map` that would  have a `addTarget(class: string)` method or something like that. Maybe it's easier to maintain two maps rather than a map whose values are pairs (OTOH, if it's consume as pair, it might be better to provide it that way).

For gathering class names, I'm also unsure about the logic. When computing `targetClassnames` (and `nonTargetClassnames`), we could directly reduce and crunch the numbers to avoid a bunch of `+1` later. (our `Sequence.groupBy` should do the trick). Then we can directly set the correct value. Here also, having a wrapping class can ease the logic with a `.from` function that does the needed (un)packing. So something like `targetClassnames.groupBy(class => class).map(list => list.size)` should directly provide a `Map` from class names to count.

(another way would be to get rid of the `targetClassnames` array and do a single pass through the DOM, adding classes to the diagnostic on the go. Something like `document.descendants().filter(isElement).forEach(element => { const isTarget = tabbales.includes(element); element.classes.forEach(class => matches.addCount(class, isTarget))})` with the `addCount` logic colocated with the data definition. While the call site looks elegant, I'm less sure about the side effects this uses 🤔 )

packages/alfa-rules/src/sia-r65/rule.ts
```diff
@@ -25,10 +28,53 @@ export default Rule.Atomic.of<Page, Element, Question.Metadata>({
...
+
+    const tabbables = document.tabOrder().filter(isTabbable(device));
+    const nonTargets = document
+      .descendants()
+      .filter((node) => isElement(node) && !tabbables.includes(node));
+    const targetClassnames = tabbables.flatMap((tabbable) => tabbable.classes);
+
+    let nonTargetClassnames = Sequence.empty<string>();
+    for (const nonTarget of nonTargets) {
+      if (isElement(nonTarget)) {
```
Jym77, line 41: This `isElement` test shouldn't be needed because `nonTargets` has already be filtered. The `&&` should be replace by `Refinement.and` to keep the type narrowing. Then `nonTargetClassnames` can probably be defined exactly the same way as `targetClassnames`.

Jym77 (changes requested):

packages/alfa-rules/src/sia-r65/rule.ts
zsofitoth: To me it sounds like the easiest solution is to keep 2 maps that would then be serialized to:
```JSON
"targetClassname" : [
  [ "useless", 1 ],
  [ "good-focus", 2 ],
  [ "btn", 1 ],
  [ "irrelevant", 1 ]
],
"nonTargetClassname": [ [ "good-focus", 1 ], [ "irrelevant", 1 ] ]
```
TBH, I don't know how it will be consumed - if we need to look up how many target and non-target occurrences share what classes this might just do.

zsofitoth: Moving mapping of separate dictionaries to Dory

Jym77 (changes requested):
Small polish.

Jym77 (dismissed):

Jym77 (approved):

PR #1167, opened 2022-06-22
Streamline earl

### LINKED ISSUE #1150: Remove duplicate contexts in EARL output
Many of our EARL output add the same context:
```json
"@context": {
        earl: "http://www.w3.org/ns/earl#",
        dct: "http://purl.org/dc/terms/",
      }
```
(some use a sligthly different one).

Repeating the same context many times in the outputted EARL document is a waste of space, we should only have it once at the top of the document.

We should get a global context, including all we need, to be included at the top of audit output, using a flag to include specific context in serialisation functions so we can keep it for non-nested serialisation.
We can also use the [global ACT rules context](https://act-rules.github.io/earl-context.json) we should contain all that we need.

Jym77:
Resolves #1150


packages/alfa-formatter-earl/package.json
```diff
@@ -19,7 +19,8 @@
   ],
   "dependencies": {
     "@siteimprove/alfa-earl": "workspace:^0.39.1",
-    "@siteimprove/alfa-formatter": "workspace:^0.39.1"
+    "@siteimprove/alfa-formatter": "workspace:^0.39.1",
+    "jsonld": "^6.0.0"
```
Jym77, line 23: While this is an external dependency, this is only used to format output of the result, not to actually compute the result.

packages/alfa-formatter/src/formatter.ts
```diff
@@ -8,7 +8,7 @@ export type Formatter<I = unknown, T = unknown, Q = never, S = T> = (
   input: I,
   rules: Iterable<Rule<I, T, Q, S>>,
   outcomes: Iterable<Outcome<I, T, Q, S>>
-) => string;
+) => Promise<string>;
```
Jym77, line 11: Due to `jsonld.compact` being async and used in the EARL formatter, we now need formatters to be async 🙈
Since formatters are only used in the CLI tool which already loads them asynchronously, this is not a real problem…

elenamongelli (approved):

elenamongelli (approved):

PR #1171, opened 2022-06-27
SIA-R42 and R68: include role in diagnostic

Jym77:
Frontend currently tries to display role by grabbing the `role` attribute which often leads to incorrect `role=""` decorator for elements with an implicit role.


elenamongelli (commented):
Logic-wise looks good to me. Seems that there is a few more cleaning to be done 🧹

PR #1172, opened 2022-06-27
SIA-R75: add style declaration to diagnostic

Jym77:


packages/alfa-rules/test/sia-r75/rule.spec.tsx
```diff
@@ -1,70 +1,81 @@
...
+      1: Outcomes.IsSufficient(Option.of(declaration2)),
     }),
     failed(R75, target1, {
-      1: Outcomes.IsInsufficient,
+      1: Outcomes.IsInsufficient(Option.of(declaration1)),
     }),
   ]);
 });
 
 test(`evaluate() passes an element with too small a font size when the font size
```
elenamongelli, line 63: Being nitpick: `evaluate() passes an element with a too small font size when the font size
      does not affect descendant text`

Issue #1160, opened 2022-06-14
Use node id in trees


Jym77:
Tree nodes could have an internal id which could be either provided by user at build time, or auto-generated by Alfa. Using namespaces to avoid collision.

Keeping a node id with the nodes enables:
* Easier translation to-and-back Dory or other system who are already computing such an id.
* Severely shrinking the JSON outputs by storing the `target` of each `Outcome` as an id rather than the full subtree. Our JSON outputs are huge for pages with a lot of issues (sometimes close to JS string size…)
* Loading an ARIA tree from a JSON representation; the pointers to DOM nodes are currently lost when serialising and deserialising, but if they are replaced by node id, then we can rebuild them when deserialising. In turn, this could let us store the ARIA tree and avoid the cost of rebuilding it when we check the same page again.

Jym77:
Add an id abstract class to `alfa-tree`, with prefix + number. (Equatable)
In `alfa-dom` and `alfa-aria`:
* create concrete sub-class for id, hardcoding the prefix.
* auto-build an id if none is provided. (using a local prefix, auto-incrementing number).
* accept an id in `Node.of()`, if only number is provided, use `user` as prefix.

Jym77:
This is probably not the way to go to solve these problems.

PR #1177, opened 2022-07-05
Add a `nodeId` to trees

Jym77:
Part of #1160 

- Add a `nodeId` to `tree.Node` and expand it to DOM and ARIA trees.
- `nodeId` consists of a `kind` (`User` for user-provided ids; `system` for Alfa-generated ones); a `type` (`alfa-dom`/`alfa-aria` for system ids); a `namespace` (user inputed for system keys) and an actual id (a number). System ids are guaranteed to be unique; it is user responsibility to provide unique user ids.
- When serialised, system ids are always serialised as user ids, keeping the same `type` and `namespace`. This greatly limit the risk of de-serialised system ids colliding with freshly created ones. (there is still a risk to create a new system id whose serialisation would collide with a user id 🤔 )
- `namespace`, also for system ids, can be used for grouping ids. Typically, `namespace` is intended to be a `pageId` so that even system ids can be tagged with it.
- When generating system ids for ARIA nodes, the `namespace` is copied from the owning DOM node.

### DIFF
--- a/docs/review/api/alfa-aria.api.md
+++ b/docs/review/api/alfa-aria.api.md
@@ -19,6 +19,7 @@ import * as json from '@siteimprove/alfa-json';
 import { Mapper } from '@siteimprove/alfa-mapper';
 import { Namespace } from '@siteimprove/alfa-dom';
 import { Node as Node_2 } from '@siteimprove/alfa-dom';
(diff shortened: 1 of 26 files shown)


Jym77:
This is becoming too complex for its own good. Killing it.

Issue #1138, opened 2022-05-09
R78 does not detect some valid content


Jym77:
```html
<div class="col col-4 col-md-12 footer-main__social">
        <p class="l" role="heading" aria-level="2">Follow</p>
    <ul>
    <li>
        <a href="https://twitter.com/siteimprove" class="btn btn-white btn-icon">
                <svg viewBox="0 0 22 18" fill="none" xmlns="http://www.w3.org/2000/svg">
                    <path …></path>
                </svg>
            <span class="vh">Twitter</span>
        </a>
    </li>
    </ul>
</div>
```
R78 dose not see the SVG as content, and does not see the `<span>` as perceivable because it incorrectly test "visible AND accessible" instead of "OR". This is due to a mismatch between definitions of perceivability in Sanshikan and Alfa 😖 

May be duplicate of #1039.

Issue #1039, opened 2022-01-24
SIA-R78 does not accept accessible but invisible content


Jym77:
The text of the rule uses the Sanshikan/ACT rules definition of [Perceivable](https://alfa.siteimprove.com/terms/perceivable-content) which requires content to be either visible or included in the accessibility tree.

The Alfa code uses its own `isPerceivable` which requires content to be both visible and included in the accessibility tree 🙈 

We need to update the rule to use the correct definition. We probably want to change the name of the current `isPerceivable` to something else, so that we have a `isPerceivable` that matches the Sanshikan/ACT definition.

PR #1178, opened 2022-07-06
Rename `IsPerceivable`, and introduce `isIncludedInTheAccessibilityTree`

Jym77:
Side work on #1039 and #1138.
This is only renaming two predicates and cascading the changes in rules.

* `isIgnored` was mostly used negatively to check that things are "included in the accessibility tree" (in ACT / Sanshikan terms), clearing that by creating and using `isIncludedInTheAccessibilityTree` directly.
* `isPerceivable` conflicts with ACT / Sanshikan definition of "Perceivable content". `isPerceivable` was checking "visible **and** included in the accessibility tree" (and is renamed `isPerceivableForAll`). "Perceivable content" is currently not used (will be needed for rules that are not there yet); is checking "visible **or** included in the accessibility tree" and is intended to be name `isPerceivableForSome`.

### DIFF
--- a/docs/review/api/alfa-aria.api.md
+++ b/docs/review/api/alfa-aria.api.md
@@ -103,6 +103,7 @@ export namespace DOM {
     hasNonEmptyAccessibleName: typeof dom.hasNonEmptyAccessibleName, // (undocumented)
     hasRole: typeof dom.hasRole, // (undocumented)
     isIgnored: typeof dom.isIgnored, // (undocumented)
+    isIncludedInTheAccessibilityTree: typeof dom.isIncludedInTheAccessibilityTree, // (undocumented)
     isMarkedDecorative: Predicate<Element_2<string>, []>, // (undocumented)
     isPerceivable: typeof dom.isPerceivable, // (undocumented)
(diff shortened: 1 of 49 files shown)


elenamongelli (approved):

docs/review/api/alfa-aria.api.md
```diff
@@ -103,8 +103,9 @@ export namespace DOM {
     hasNonEmptyAccessibleName: typeof dom.hasNonEmptyAccessibleName, // (undocumented)
     hasRole: typeof dom.hasRole, // (undocumented)
     isIgnored: typeof dom.isIgnored, // (undocumented)
```
zsofitoth, line 105: Is `isIgnored` still needed now that `isIncludedInTheAccessibilityTree` exists?

zsofitoth (approved):

docs/review/api/alfa-aria.api.md
Jym77: More or less. There are still a few cases where we use it. We could of course use `not(isIncludedInTheAccessibilityTree)` instead, but I'm not sure this is clearer… So leaving it for now 😃

PR #1182, opened 2022-07-08
SIA-R66, SIA-R69: `getForeground` should resolve `currentcolor` automatically

### LINKED ISSUE #1174: SIA-R66, SIA-R69: `getForeground` should resolve `currentcolor` automatically.
[`getForeground`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/common/dom/get-colors.ts#L267) calls [`resolve`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/common/dom/get-colors.ts#L279) on the computed `color`.

However, when encountering `currentcolor`, `resolve` just [uses the computed `color`](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/common/dom/get-colors.ts#L30-L31). Which is correct for any property **except** `color.

So when `color: currentcolor` is used, `resolve` (and `getForeground`) fail. But should instead [use the inherited value](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword).

Before calling `resolve`, if the color is `currentcolor`, recursively call `getForeground` on the parent.

- [ ] Follow up on DEV-13272

zsofitoth:
Resolves #1174


packages/alfa-rules/test/common/dom/get-colors.spec.tsx
```diff
@@ -289,6 +289,77 @@ test("getForeground() handles a mix of opacity and transparency and a linear gra
...
+                      <div>
+                        <div>Hello {target}!</div>
+                      </div>
+                    </div>
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
```
zsofitoth, line 314: @Jym77 Is there a need to optimize for very deeply nested tree structures and if yes, has it been done before elsewhere?

packages/alfa-rules/test/common/dom/get-colors.spec.tsx
```diff
@@ -289,6 +289,77 @@ test("getForeground() handles a mix of opacity and transparency and a linear gra
...
+      </html>,
+    ],
+    [
+      h.sheet([
+        h.rule.style("p", {
```
Jym77, line 319:
```suggestion
        h.rule.style("div", {
```
(actually, this is OK as it is, but the innermost `<div>` has `color: inherited` not `currentcolor`, so I'm not sure what the deep nesting test 🤔 )

packages/alfa-rules/test/common/dom/get-colors.spec.tsx
Jym77: I think we're OK until we encounter some crazy cases 🙈 (plus, this is the same as `inherited` which is the default).

packages/alfa-rules/src/common/dom/get-colors.ts
```diff
@@ -276,6 +276,17 @@ export function getForeground(
...
+        foregroundColor.type === "keyword" &&
+        foregroundColor.value === "currentcolor" &&
+        element.parent().isSome()
+      ) {
+        const parent = element.parent().get();
+        if (isElement(parent)) {
+          return getForeground(parent, device, context);
+        }
+      }
+
```
Jym77, line 289: Thinking about it, I'm getting scared this may mess up opacity resolution (since the recursive call to `getForeground` will resolve the `opacity` of the parent but likely shouldn't, i.e.
```html
<div style="color: blue; opacity: 0.5">
  <span style="color: currentcolor">Hello</span>
</div>
```
The `<span>` should have a computed `color` of `blue` (which is them resolved with opacity), but `getForeground` on the parent returns 50% blue only.

---
So this should be more something like
```typescript
while(element.isSome() && !isCurrentColor(color)) {
  color = Style.from(element.get()).computed("color")
  element = element.parent().filter(isElement)
}
```

Jym77 (changes requested):

Jym77 (approved):

PR #1185, opened 2022-07-11
SIA-R16: only apply to explicit role

### LINKED ISSUE #949: Investigate managed attributes and whether R16 should apply to implicit roles
See https://github.com/w3c/html-aam/issues/349 and https://github.com/w3c/html-aria/issues/384

This would also make https://github.com/Siteimprove/alfa/blob/main/packages/alfa-aria/src/feature.ts#L580-L585 useless since `<select>` are also in that category.

Jym77:
Resolves #949 

See https://github.com/Siteimprove/sanshikan/pull/249

### DIFF
--- a/packages/alfa-rules/test/sia-r16/rule.spec.tsx
+++ b/packages/alfa-rules/test/sia-r16/rule.spec.tsx
@@ -24,34 +24,6 @@ test(`evaluate() passes a <div> element with a role of checkbox and an
   ]);
 });
 
-test(`evaluate() passes an <input> element with a type of checkbox`, async (t) => {
-  const target = <input type="checkbox" />;
-
-  const document = h.document([target]);
-
-  t.deepEqual(await evaluate(R16, { document }), [
-    passed(R16, target, {
-      1: Outcomes.HasAllStates(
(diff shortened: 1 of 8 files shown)


zsofitoth (approved):

packages/alfa-rules/test/sia-r16/rule.spec.tsx
```diff
@@ -129,22 +101,18 @@ test("evaluate() is inapplicable to elements that are not exposed", async (t) =>
...
 
-test(`evaluate() passes a native \`<input type="text" list="foo">\` combobox`, async (t) => {
-  const target = <input type="text" list="foo" />;
-  const datalist = (
-    <datalist id="foo">
-      <option value="foo">foo</option>
-      <option value="bar">bar</option>
-    </datalist>
-  );
+test("evaluate() is inapplicable to elemnts with no explicit role", async (t) => {
```
zsofitoth, line 104:
```suggestion
test("evaluate() is inapplicable to elements with no explicit role", async (t) => {
```

zsofitoth (approved):

PR #1189, opened 2022-07-13
Improve typing of `Array.sortWith`

Jym77:
```typescript
type Foo: {foo: number}
type Bar: {bar: string}

const comparer: Comparer<Foo> = (a, b) => Comparable.compareNumber(a.foo, b.foo);

const source: Array<Foo & Bar> = []

const dest = source.sortWith(comparer);
```
This used to lost the fact that the array contains `Foo & Bar` and only return an `Array<Foo>` because the same type was used for the array and the comparer. Now we can use a comparer on a super-type and keep the narrow type on the array.


elenamongelli (dismissed):
LGTM 👍

elenamongelli (approved):

Issue #1139, opened 2022-05-09
Consider refactoring R66/R69


Jym77:
R66 and R69 have a lot in common. The only difference beings the contrast threshold in expectation.
We should probably refactor that. This may also be useful with more colour contrast rules (R88, R103, …) so we may want to wait until we tackle them to be sure we refactor the bits we need.

Jym77:
Closed by #1235

PR #1191, opened 2022-07-15
R66/R69 should ignore elements that are used in name of disabled widgets

### LINKED ISSUE #1149: R66/R69 should ignore elements that are used in name of disabled widgets
ACT rule and Sanshikan currently have this exception, but Alfa hasn't.

Gather all `aria-disabled` widgets on the document, compute their name, filter out the text nodes that take part in these names.

zsofitoth:
Resolves #1149

### DIFF
--- a/packages/alfa-rules/src/sia-r66/rule.ts
+++ b/packages/alfa-rules/src/sia-r66/rule.ts
@@ -1,5 +1,6 @@
+/// <reference lib="dom" />
 import { Rule } from "@siteimprove/alfa-act";
-import { DOM } from "@siteimprove/alfa-aria";
+import { DOM, Name } from "@siteimprove/alfa-aria";
 import { Element, Text, Namespace, Node } from "@siteimprove/alfa-dom";
 import { Iterable } from "@siteimprove/alfa-iterable";
 import { Predicate } from "@siteimprove/alfa-predicate";
@@ -19,6 +20,7 @@ import { contrast } from "../common/expectation/contrast";
 import { Contrast as Outcomes } from "../common/outcome/contrast";
 
 import { Scope } from "../tags";
+import { Sequence } from "@siteimprove/alfa-sequence";
 
 const { hasRole, isPerceivableForAll, isSemanticallyDisabled } = DOM;
 const { hasNamespace, isElement } = Element;
@@ -33,33 +35,64 @@ export default Rule.Atomic.of<Page, Text, Question.Metadata>({
   requirements: [Criterion.of("1.4.6")],
   tags: [Scope.Component],
   evaluate({ device, document }) {
+    // 1. Gather all aria-disabled widgets on the document
+    // 2. Compute their name
+    // 3. Filter out text nodes that are contained in these names
+    let disabledWidgets: Sequence<Element> = Sequence.empty();
+    let textNodes: Sequence<Text> = Sequence.empty();
     return {
       applicability() {
-        return visit(document);
+        gather(document);
+        console.log(textNodes.toJSON());
+        return getApplicableLinks();
 
-        function* visit(node: Node): Iterable<Text> {
+        function gather(node: Node): void {
+          // Gather all applicable text nodes
+          if (
+            node.parent().isSome() &&
+            test(
+              and(
+                isElement,
+                and(
+                  hasNamespace(Namespace.HTML),
+                  not(hasRole(device, (role) => role.isWidget())),
+                  and(
+                    not(hasRole(device, "group")),
+                    not(isSemanticallyDisabled)
+                  )
+                )
+              ),
+              node.parent().get()
+            ) &&
+            test(and(isText, isPerceivableForAll(device)), node)
+          ) {
+            textNodes = textNodes.append(node);
+          }
+
+          // Gather all aria-disabled widgets on the document
           if (
             test(
               and(
                 isElement,
-                or(
-                  not(hasNamespace(Namespace.HTML)),
+                and(
                   hasRole(device, (role) => role.isWidget()),
-                  and(hasRole(device, "group"), isSemanticallyDisabled)
+                  isSemanticallyDisabled
                 )
               ),
               node
             )
           ) {
-            return;
-          }
-
-          if (test(and(isText, isPerceivableForAll(device)), node)) {
-            yield node;
+            disabledWidgets = disabledWidgets.append(node);
           }
 
           for (const child of node.children(Node.fullTree)) {
-            yield* visit(child);
+            gather(child);
+          }
+        }
+
+        function* getApplicableLinks(): Iterable<Text> {
+          for (const textNode of textNodes) {
+            yield textNode;
           }
         }
       },
--- a/packages/alfa-rules/src/sia-r66/rule.ts
+++ b/packages/alfa-rules/src/sia-r66/rule.ts
@@ -1,5 +1,5 @@
 import { Rule } from "@siteimprove/alfa-act";
-import { DOM } from "@siteimprove/alfa-aria";
+import { DOM, Node as ariaNode } from "@siteimprove/alfa-aria";
 import { Element, Text, Namespace, Node } from "@siteimprove/alfa-dom";
 import { Iterable } from "@siteimprove/alfa-iterable";
 import { Predicate } from "@siteimprove/alfa-predicate";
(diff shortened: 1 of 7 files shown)


zsofitoth:
For now it's only for R66 - after the review I will apply it to R69 as well. I'm not sure if they should somehow share `disabledWidgetNames` and `textNodes` though to not compute those twice.

packages/alfa-rules/src/sia-r66/rule.ts
```diff
@@ -35,31 +36,66 @@ export default Rule.Atomic.of<Page, Text, Question.Metadata>({
...
               ),
               node
             )
           ) {
-            return;
+            // Compute the names of these elements
+            const names = node.attributes
+              .filter((attribute) => attribute.name === "name")
+              .map((attribute) => attribute.value);
+            disabledWidgetNames = disabledWidgetNames.concat(names);
```
Jym77, line 83: This should be about their accessible name, not the `name` attribute.
So, something like `ariaNode.from(node, device).name`.

The name does contain a `source` which recursively contains all nodes taking part in the computation: https://github.com/Siteimprove/alfa/blob/main/packages/alfa-aria/src/name.ts#L86-L91
(it may be worth to have the source implement `Iterable<dom.Node>` so they can each iterate the nodes taking part in their computation and we can just use this iterator here, this is done like that: https://github.com/Siteimprove/alfa/blob/main/packages/alfa-selector/src/selector.ts#L131-L133)

packages/alfa-rules/src/sia-r66/rule.ts
```diff
@@ -35,31 +36,66 @@ export default Rule.Atomic.of<Page, Text, Question.Metadata>({
...
-          if (test(and(isText, isPerceivableForAll(device)), node)) {
-            yield node;
+          for (const child of node.children(Node.fullTree)) {
+            gather(child);
           }
+        }
 
-          for (const child of node.children(Node.fullTree)) {
-            yield* visit(child);
+        function* getApplicableLinks(): Iterable<Text> {
```
Jym77, line 91:
```suggestion
        function* getApplicableTexts(): Iterable<Text> {
```

packages/alfa-rules/test/sia-r66/rule.spec.tsx
```diff
@@ -262,6 +262,19 @@ test("evaluate() is inapplicable to text nodes in disabled groups", async (t) =>
   t.deepEqual(await evaluate(R66, { document }), [inapplicable(R66)]);
 });
 
+test("evaluate() is inapplicable to elements that are used in name of disabled widgets", async (t) => {
+  const document = h.document([
+    <html>
+      <fieldset name="Disabled widget name" disabled>
+        <button>Hello world</button>
+      </fieldset>
+      <div>widget name</div>
```
Jym77, line 271:
```suggestion
      <fieldset aria-labelledby="label" disabled>
        <button>Hello world</button>
      </fieldset>
      <div id="label">widget name</div>
```
(but a different `<div>widget name</div>` would be applicable)

Jym77 (changes requested):

Jym77:
> For now it's only for R66 - after the review I will apply it to R69 as well. I'm not sure if they should somehow share `disabledWidgetNames` and `textNodes` though to not compute those twice.

Ideally, we'd want to share it of course. But that is likely a topic for #1139

packages/alfa-rules/src/sia-r66/rule.ts
```diff
@@ -35,31 +37,68 @@ export default Rule.Atomic.of<Page, Text, Question.Metadata>({
...
+        let textNodes: Sequence<Text> = Sequence.empty();
 
-        function* visit(node: Node): Iterable<Text> {
+        gather(document);
+        return getApplicableTexts();
+
+        function gather(node: Node): void {
+          // Gather all applicable text nodes
+          if (
+            node.parent().isSome() &&
```
Jym77, line 49:
```suggestion
            node.parent().some(…
```
Will take a predicate as argument and automatically be `false` if the option is `None`.

Jym77 (changes requested):
Looks good. Some refactoring.

packages/alfa-rules/src/sia-r66/rule.ts
```diff
@@ -35,31 +37,68 @@ export default Rule.Atomic.of<Page, Text, Question.Metadata>({
...
+        }
 
-          for (const child of node.children(Node.fullTree)) {
-            yield* visit(child);
+        function* getApplicableTexts(): Iterable<Text> {
+          // Filter out text nodes that are contained in disabledWidgetNames
+          for (const textNode of textNodes) {
+            if (disabledWidgetTexts.none((text) => text.equals(textNode))) {
+              yield textNode;
+            }
```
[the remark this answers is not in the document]
zsofitoth: Tried to do the filtering as part of `gather`, but didn't work with:
https://act-rules.github.io/rules/afw4f7#inapplicable-example-8
As the `input` and `text` there are siblings

packages/alfa-rules/src/sia-r66/rule.ts
```diff
@@ -35,31 +37,67 @@ export default Rule.Atomic.of<Page, Text, Question.Metadata>({
...
           if (
             test(
               and(
                 isElement,
                 or(
-                  not(hasNamespace(Namespace.HTML)),
                   hasRole(device, (role) => role.isWidget()),
-                  and(hasRole(device, "group"), isSemanticallyDisabled)
+                  isSemanticallyDisabled
                 )
```
Jym77, line 54: This should be a `and` (to get disabled widgets: both widget and disabled).
So
```html
<label>
  some text
  <input />
</label>
```
the text is in the name of an active widget and should be tested.

Jym77 (changes requested):

packages/alfa-rules/src/sia-r66/rule.ts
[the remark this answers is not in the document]
Jym77: 🤔 I'm not sure I see why tris cause a problem. By pre-computing the `disableWidgetTexts` we should see that the label is one and correctly be able to filter it out (which is what `getApplicabelTexts` is already doing).
So

packages/alfa-rules/src/sia-r66/rule.ts
```diff
@@ -35,6 +37,34 @@ export default Rule.Atomic.of<Page, Text, Question.Metadata>({
...
+                  and(
+                    or(
+                      hasRole(device, (role) => role.isWidget()),
+                      hasRole(device, "group")
+                    ),
+                    isSemanticallyDisabled
+                  )
+                ),
+                node
+              )
```
Jym77, line 57: Looks like this can be eta-contracted and the call to `test` removed.
```suggestion
            .filter(and(
                  isElement,
                  and(
                    or(
                      hasRole(device, (role) => role.isWidget()),
                      hasRole(device, "group")
                    ),
                    isSemanticallyDisabled
                  )
                )
```

packages/alfa-rules/src/sia-r66/rule.ts
```diff
@@ -35,6 +37,34 @@ export default Rule.Atomic.of<Page, Text, Question.Metadata>({
...
+          document
+            .descendants(Node.fullTree)
+            .filter((node) =>
+              test(
+                and(
+                  isElement,
+                  and(
+                    or(
+                      hasRole(device, (role) => role.isWidget()),
+                      hasRole(device, "group")
```
Jym77, line 51: We can probably one-line these
```suggestion
                      hasRole(device, (role) => role.isWidget() || role.is("group"))
```

Jym77 (changes requested):
Great. Some more polishing 😅

packages/alfa-rules/src/sia-r66/rule.ts
```diff
@@ -54,7 +84,10 @@ export default Rule.Atomic.of<Page, Text, Question.Metadata>({
             return;
           }
 
-          if (test(and(isText, isPerceivableForAll(device)), node)) {
+          if (
+            test(and(isText, isPerceivableForAll(device)), node) &&
+            !disabledWidgetTexts.has(node)
```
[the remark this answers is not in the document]
zsofitoth: It doesn't quite work with the `isText` as the `node` is still type of `Node` there and the compiler is not happy because in `disabledWidgetTexts` there are `Text`s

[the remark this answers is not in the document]
Jym77: Ah yes, the first `and(isText, …)` should be a `Refinement.and` so that the second `and(isPerceivable, not(set.has))` only receives `Text`. 🤔 (also maybe wrong parenthesis in the suggestion 🙃 should be `and(isText, and(isPerceivable, not(set.has)))`)

Jym77 (approved):

PR #1196, opened 2022-07-21
SIA-R19: Accept negative integers and numbers 

### LINKED ISSUE #1183: SIA-R19 should accept negative integers (and numbers)
In SIA-R19, the [regexp for matching `integer` and `number` types](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-rules/src/sia-r19/rule.ts#L84-L88) only accepts unsigned values, but negative values are explicitly allowed by ARIA, at least for [`aria-setsize`](https://w3c.github.io/aria/#aria-setsize):
> When specifying aria-setsize, authors MUST set the value to an integer equal to the number of items in the set. If the total number of items is unknown, authors SHOULD set the value of aria-setsize to -1.

For now, just update these regex to accept signed numbers.

zsofitoth:
Resolves #1183


packages/alfa-rules/src/sia-r19/rule.ts
```diff
@@ -82,10 +82,10 @@ function isValid(attribute: aria.Attribute): boolean {
       return true;
 
     case "integer":
-      return /^\d+$/.test(value);
+      return /^-?\d+$/.test(value);
```
Jym77, line 85:
```suggestion
      return /^[+-]?\d+$/.test(value);
```
Let's accept `+1` also, just in case.

packages/alfa-rules/src/sia-r19/rule.ts
```diff
@@ -82,10 +82,10 @@ function isValid(attribute: aria.Attribute): boolean {
...
 
     case "number":
-      return /^\d+(\.\d+)?$/.test(value);
+      return /^-?\d+(\.\d+)?$/.test(value);
```
Jym77, line 88:
```suggestion
      return /^[+-]?\d+(\.\d+)?$/.test(value);
```

Jym77 (changes requested):

Jym77 (approved):

PR #1200, opened 2022-07-22
Compute accessible name correctly from the content

### LINKED ISSUE #1054: Accessible name from content is not computed correctly in case of `display: block` descendants
The accessible name specs say to join the content from descendants, but is not clear whether that should be done with or without space between them (https://github.com/w3c/accname/issues/15). Back in #889 we decided to join them without space. This seems to be incorrect when there are spaces between the elements, or when the elements have `display: block` which adds visual spacing (new line) upon rendering.

```html
<button>
  <span>inline</span><span>no space</span>
</button>

<button>
  <span>inline</span> <span>with space</span>
</button>

<button>
  <div>block</div><div>no space</div>
</button>

<button>
  <div>block</div> <div>with space</div>
</button>

<button>
  <div>block</div><span>inline</span>
</button>

<button>
  <span>inline</span><div>block</div>
</button>
```
Chrome (Version 97.0.4692.99) gives
"inlineno space", "inlinewith space", "block no space", "block with space", "block inline", "inline block"
Firefox (V 96.0.3) gives
"inlineno space", "inline with space", "block no space", "block with space", "block inline", "inline block"

Alfa gets no space between the elements in all cases.

zsofitoth:
Resolves #1054


zsofitoth:
Converting back to draft, seems like the name testable statements caught some things

Jym77:
> Converting back to draft, seems like the name testable statements caught some things

Yes, this should precisely fix one of the test exceptions 😀 (with ref to the same issue).
Should be just deleting the corresponding bit of the `switch` + re-generating the test file 🤞

Jym77 (approved):

PR #1201, opened 2022-07-22
Calc expressions

### LINKED ISSUE #980: Support `calc` expressions in more CSS properties.
Back when `calc` support was introduced in #376 we only added support for in in `font-size`.
Other properties also need to accept `calc` values (anything accepting numbers should accept valid `calc`…) Doing it on a case-by-case basis (like for `font-size`) doesn't sound like a good idea (but can be a quick fix for some bugs).

We should investigate how to get a more generic method to parse and compute `calc` expression in properties that accept numbers.

Jym77:
Resolves #980 

Add infrastructure for easy support of `calc()` expressions as visible at https://github.com/Siteimprove/alfa/commit/254ddefd4fcccf293209d9660fb3d97b11a4e60e

Effectively adding support in all other properties should be taken separately.


Jym77:
API extractor is unhappy about the variance indication that is needed here (it's a TS 4.7 syntax): https://github.com/microsoft/rushstack/issues/3486#issuecomment-1197988205

Building a similar hack to circumvent it and wait until fix 😖 as the previous time.

elenamongelli (approved):
LGTM

PR #1206, opened 2022-07-26
Elements are consider hovered if their descendants are

### LINKED ISSUE #1158: Elements should be consider hovered when their descendants are
```html
<style>
  div:hover {background-color: red}
  span:hover {border: 2px green solid}
</style>
<div> <span>Hello</span> </div>
```
When the `<span>` is hovered (and the border show), browsers also consider the `<div>` as hovered (and put a red background), but Alfa doesn't.

We need to update `Hover#matches` in `alfa-selector` to be true if the context is hovering a descendant.
It seems that when the hovered descendant is not painted on top of the element (due to absolute positioning), the element is still considered hovered when is there is some other content in the element (otherwise, it has a 0 size) 🤔 So we are probably good in not caring about descendants being painted out of their ancestors until this proves to be a problem.

elenamongelli:
Resolves #1158

### DIFF
--- a/packages/alfa-selector/src/selector.ts
+++ b/packages/alfa-selector/src/selector.ts
@@ -34,7 +34,7 @@ const {
   zeroOrMore,
 } = Parser;
 
-const { and, not, property, equals } = Predicate;
+const { and, not, property, equals, or } = Predicate;
 const { isElement, hasName } = Element;
 
 /**
@@ -1234,7 +1234,10 @@ export namespace Selector {
       element: Element,
       context: Context = Context.empty()
     ): boolean {
-      return context.isHovered(element);
+      return element
+        .inclusiveDescendants(Node.fullTree)
+        .filter(isElement)
+        .some((element) => context.isHovered(element));
     }
   }
 
--- a/packages/alfa-selector/test/selector.spec.tsx
+++ b/packages/alfa-selector/test/selector.spec.tsx
@@ -1047,6 +1047,15 @@ test("#matches() checks if an element matches a :hover selector", (t) => {
   t.equal(selector.matches(p, Context.hover(p)), true);
 });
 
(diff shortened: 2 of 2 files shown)


packages/alfa-selector/src/selector.ts
```diff
@@ -34,7 +34,7 @@ const {
   zeroOrMore,
 } = Parser;
 
-const { and, not, property, equals } = Predicate;
+const { and, not, property, equals }= Predicate;
```
Jym77, line 37:
```suggestion
const { and, not, property, equals } = Predicate;
```

packages/alfa-selector/src/selector.ts
```diff
@@ -1234,7 +1234,10 @@ export namespace Selector {
       element: Element,
       context: Context = Context.empty()
     ): boolean {
-      return context.isHovered(element);
+      return element
+        .inclusiveDescendants(Node.fullTree)
+        .filter(isElement)
+        .some((element) => context.isHovered(element));
```
Jym77, line 1240:
```suggestion
        .some(context.isHovered);
```
This can likely be eta-contracted.

packages/alfa-selector/test/selector.spec.tsx
```diff
@@ -1047,6 +1047,15 @@ test("#matches() checks if an element matches a :hover selector", (t) => {
   t.equal(selector.matches(p, Context.hover(p)), true);
 });
 
+test("#matches() checks if an element matches a :hover selector inherited from its descendant", (t) => {
```
Jym77, line 1050:
```suggestion
test("#matches() checks if an element matches a :hover selector when its descendant is hovered", (t) => {
```

Jym77 (changes requested):
Typos and polish.

Jym77 (approved):

PR #1215, opened 2022-08-29
Update Dependencies

Jym77:
Breaking a nasty line of out-dated packages that ultimately forced an ugly workaround in the documenter script to use TS 4.7+ features…

### DIFF
--- a/docs/review/api/alfa-xpath.api.md
+++ b/docs/review/api/alfa-xpath.api.md
@@ -136,7 +136,7 @@ export namespace Builder {
         // (undocumented)
         where(predicate: Builder): Path;
     }
-    {};
+        {};
 }
 
 // Warning: (ae-internal-missing-underscore) The name "coerceItems" should be prefixed with an underscore because the declaration is marked as @internal
@@ -311,7 +311,7 @@ export namespace Expression {
         get type(): "decimal";
         // (undocumented)
         get value(): number;
-        }
+    }
     // (undocumented)
     export namespace Decimal {
         // (undocumented)
@@ -338,7 +338,7 @@ export namespace Expression {
         get type(): "double";
         // (undocumented)
         get value(): number;
-        }
+    }
     // (undocumented)
     export namespace Double {
         // (undocumented)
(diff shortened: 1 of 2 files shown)


packages/alfa-dom/src/jsx.ts
```diff
@@ -24,16 +24,16 @@ export function jsx<N extends string = string>(
 
     switch (name) {
       case "style":
-        const properties = value as Record<string, string>;
-
-        for (const [name, value] of entries(properties)) {
-          style[name] = value;
```
Jym77, line 30: Due to [the way `unknown` is now handled](https://devblogs.microsoft.com/typescript/announcing-typescript-4-8/#improved-intersection-reduction-union-compatibility-and-narrowing), the non-null checks at line 21 already narrowed `value` to an object. So there is no need for the type assertion anymore.

packages/alfa-map/test/map.spec.ts
```diff
@@ -49,7 +49,7 @@ test("#apply() applies a map of functions to each corresponding value of a map",
     [
       ...map.apply(
         Map.of(
-          ["foo", (x) => x + 1],
+          ["foo", (x: number) => x + 1],
```
Jym77, line 52: Not sure why these now need some typing. I assume it is also due to the changes around `unknown` in TS 4.8.

packages/alfa-record/src/record.ts
```diff
@@ -17,9 +17,10 @@ export class Record<T>
     Foldable<Record.Value<T>>,
     Iterable<Record.Entry<T>>,
     Equatable,
-    Serializable<Record.JSON<T>> {
+    Serializable<Record.JSON<T>>
+{
   public static of<T>(properties: T): Record<T> {
-    const keys = Object.keys(properties).sort() as Array<Record.Key<T>>;
+    const keys = Object.keys(properties ?? {}).sort() as Array<Record.Key<T>>;
```
Jym77, line 23: By default, `T extends unknown` but with the new handling, we need to explicitly non-null check to narrow to an object.

elenamongelli (approved):

Issue #1202, opened 2022-07-22
Support `calc()` expressions in all properties


Jym77:
New plan of attack… (post #1416 )

- [x] Replace the hybrid `Length` value in `alfa-style` with the one from `alfa-css/src/value/numeric`.
- [x] Switch all `Length` usage from `alfa-css/src/calculation/numeric` to `alfa-css/src/value/numeric`. This will probably require to provide a `parseBase` function in the later that can only parse non-calculated lengths (as a temporary solution). This will also need to expose all methods from the former on the later.
- [x] Wherever `Length` is used as a "final" value type (i.e. in `alfa-style`), replace `parseBase` with `parse` (i.e. accept calculation and provide a `resolve` helper). There may also be places in `alfa-media` that will need update. 
    Note: some style properties do not accept negative values, and this is treated differently for calculated and non-calculated values: [Range checking](https://drafts.csswg.org/css-values-4/#calc-range). 
    Note: this will likely require to buff up the new `Length` with methods from the old one (`isZero`, …)
    Note: it seems that calculation in media queries need to be parenthesised (`min-width: (10px + 10px)`, not `min-width: 10px + 10px`), need to find normative reference for that.
- [x] At this point, it should be possible to strip down `calculation/numeric/length` to only what is needed in math expressions.
- [x] Repeat the switch for other numeric types (number, integer, percentage, angle). 
    Note: it might be worth re-introducing a common `Dimension` abstraction on the new side also.
- [x] Add the `dimension-percentage` types we need, i.e. `length-percentage` and `angle-percentage`. 
    Note: it might be worth having an abstraction here also. 
    Note: it might be worth recording in the type whether the non-calculated value is holding a dimension or a percentage.
- [x] At this point, it should be possible to completely get rid of `alfa-style/src/property/value/compound.ts`.
- [ ] #1444 
- [x] Add type `Value.Parser<V> = Parser<Slice<Token>, V, string>` since we use it everywhere.
- [x] Clean up existing values to explicitly add `implements Resolvable` where relevant.
- [x] Update the other value types in `alfa-css` in order to use `src/value/numeric` rather than `src/calculation/numeric`. It is likely best to first introduce a `parseBase` to parse non-calculated values for easy migration.
  - [x] box
  - [x] collection
  - [x] color
  - [x] image
  - [x] keyword
  - [x] position
  - [x] shadow
  - [x] shape
  - [x] string
  - [x] transform
[...]

PR #1219, opened 2022-09-06
Support calculation in `outline-*`

Jym77:
Part of #1202 which has caused problem in DEV-13509.

### DIFF
--- a/packages/alfa-style/test/property/outline.spec.tsx
+++ b/packages/alfa-style/test/property/outline.spec.tsx
@@ -1,0 +1,39 @@
+import { test } from "@siteimprove/alfa-test";
+import { h } from "@siteimprove/alfa-dom/h";
+
+import { Device } from "@siteimprove/alfa-device";
+
+import { Style } from "../../src/style";
+
+const device = Device.standard();
+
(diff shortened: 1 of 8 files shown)


packages/alfa-style/test/property/outline.spec.tsx
```diff
@@ -0,0 +1,103 @@
...
+  });
+});
+
+test("#computed() rejects `outline-offset: calc(1 + 2)`", (t) => {
+  const element = <div style={{ outlineOffset: "calc(1 + 2)" }} />;
+
+  const style = Style.from(element, device);
+
+  t.deepEqual(style.computed("outline-offset").toJSON(), {
+    // initial value
```
elenamongelli, line 31:
```suggestion
    // initial value
```
Clean up?

elenamongelli (approved):
LGTM

packages/alfa-style/test/property/outline.spec.tsx
Jym77: That's on purpose. The calculation in that example is invalid for the property, so it defaults to its initial value.

Issue #1222, opened 2022-09-12
Provide extended diagnostic for "new" rules


Jym77:
- [x] R17: add list of focusable element inside target
- [x] R55: role and name of the group CE-6152
- [x] R60: add role of target CE-6132
- [x] R61: first heading and its level CE-6133
- [x] R76: add target's role CE-6136
- [x] R78: add heading level + next heading of the same or lower level + level of next heading CE-6138
- [x] R90: add list of focusable element inside target CE-6140
- [x] R91: add `style` attribute + targeted property + used value of property + computed value of `font-size` + (ratio +) threshold CE-6141
- [x] R92: add `style` attribute + targeted property + used value of property + computed value of `font-size` + (ratio +) threshold CE-6142
- [x] R93: add `style` attribute + targeted property + used value of property + computed value of `font-size` + (ratio +) threshold CE-6143
- [x] R95: add list of focusable element inside target CE-6145

PR #1224, opened 2022-09-16
Add extended diagnostics for new-ish rules

Jym77:
Part of #1222.

R91, R92, R93 may undergo heavy changes upstream, so waiting for the outcome before changing them here.


elenamongelli (approved):
LGTM. CodeQL points out that some variable clean up can be done

elenamongelli (approved):

PR #1255, opened 2022-11-03
`isVisible`: treat text with same color as background invisible

### LINKED ISSUE #1251: SIA-R69 Color Contrast is incorrectly checking "offscreen" text
Text that is moved offscreen and/or has a height of 0 is still being checked for color contrast even though there is no visible text on screen. 

Offscreen text is often used to provide accessible names or additional text content for screen reader user experience and is a very common practice but color contrast is irrelevant. 

In this example there is no visible text in the UI, but fails this color contrast check.

```
<div style="background:#fff;">
  <p style="display:block; color:#fff; height:0; text-indent:-999em;" >Offscreen text</p>
</div>
```

Jym77:
Resolves #1251


packages/alfa-rules/test/sia-r62/rule.applicability.spec.tsx
```diff
@@ -352,7 +352,7 @@ test(`evaluate() is applicable to an <a> element with a <p> parent element that
       .withStyle(
         [
           "background",
-          "linear-gradient(to right, rgb(0% 0% 0%) 50%, rgb(0% 0% 93.33333%) 50%) 0% 0%",
+          "linear-gradient(to right, rgb(0% 0% 0%) 50%, rgb(0% 0% 93.33333%) 50%)",
```
Jym77, line 355: Fixing the `initial` value in `background-position-{x,y}` result in these being removed because we filter out initial values during serialisation.

elenamongelli (approved):

PR #1259, opened 2022-11-04
Fix detection of CSS variable circular reference on ancestor styles

Jym77:
In `resolve`, when the variable (custom property) was not found in the current style but searched for on ancestors instead, the recursive call to `substitute` did not record the property as `visited`. This resulted in Alfa not seeing a circular reference on it and wasting time following it until it reached the substitution limit.

This is fixed, in addition, due to inheritance of CSS variable, the search in ancestor style is pre-built by merging the maps before doing the substitution. This removes some recursive search in the style tree.

Native `ReadonlyMap` in style management have also been replaced by Alfa's `Map` (who are readonly by design). Plus a bit of clean up and refactoring of the code.


packages/alfa-style/test/style.spec.tsx
```diff
@@ -647,9 +648,17 @@ test(`#cascaded() correctly resolves var() function references within context
         // for `div` will therefore not apply.
         h.rule.style("div", {
           overflowX: "var(--hidden)",
-
           "--hidden": "var(--really-hidden)",
         }),
+        // This declaration references `--foo`, but inherits its value from
+        // `main` above. The substitutipn of `--foo` therefore happens within
```
elenamongelli, line 654:
```suggestion
        // `main` above. The substitution of `--foo` therefore happens within
```

elenamongelli (dismissed):

PR #1267, opened 2022-11-10
Fix `aria-labelledby` reference into hidden content

### LINKED ISSUE #1236: Name from `aria-labelledby` should accept `aria-hidden` descendant of an `aria-hidden`reference
```html
<button aria-labelledby="label"></button>
<span id="label" aria-hidden="true"><span>Hello</span></span>
```
The element referred to by `aria-labelledby` being itself hidden, the exception of [step 2A](https://w3c.github.io/accname/#step2A) does not trigger and we should still use the inside text.

Jym77:
Resolves #1236

### DIFF
--- a/packages/alfa-aria/test/name.spec.tsx
+++ b/packages/alfa-aria/test/name.spec.tsx
@@ -1,8 +1,8 @@
-import { h } from "@siteimprove/alfa-dom/h";
 import { test } from "@siteimprove/alfa-test";
 
 import { Device } from "@siteimprove/alfa-device";
-import { Element, Namespace, Text } from "@siteimprove/alfa-dom";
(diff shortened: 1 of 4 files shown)


packages/alfa-aria/test/name.spec.tsx
```diff
@@ -1138,6 +1153,41 @@ test(`.from() correctly handles aria-labelledby references to hidden elements
   });
 });
 
+test(`.from() correctly handles aria-labelledby references to elements
+      with hidden child elements with child text content`, (t) => {
```
elenamongelli, line 1157:
```suggestion
      with hidden children elements with child text content`, (t) => {
```

elenamongelli (approved):

elenamongelli (approved):

PR #1283, opened 2022-11-28
Lower typescript target to es2019 from es2020

dlemvigh:
alfa-cypress was causing errors due to nullish coalescing operator (??).
By lowering the typescript build target to es2019, the issue goes away.

Was able to test it locally, by cloning and building alfa with the lowered build target, and then copy the files to my cypress project.

The first instance of the error is thrown from alfa-css `src/value.js`. 
It is therefor not enough to just lower the build target for just alfa-cypress.

Is there any danger for other projects in lowering the build target to es2019?

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,7 +9,7 @@
     "module": "commonjs",
     "sourceMap": true,
     "strict": true,
-    "target": "es2020",
+    "target": "ES2019",
     "types": []
   },
   "files": [],


dlemvigh:
The issue is simililar to this https://github.com/scottohara/loot/issues/185

It isn't directly an issue with alfa-cypress, but a compatibility issue with cypress

Jym77:
This should have been fixed in Cypress 13: https://github.com/cypress-io/cypress/pull/27438
Leaving this open until we managed to get a proof of concept of it working.

Jym77:
Updating Cypress to version 13 resolves the problem as they now accept more modern ES syntax.
See https://github.com/Siteimprove/alfa-examples/pull/39 for a detailed example.

Closing this PR now.

PR #1292, opened 2022-12-13
SIA-R45: exclude tables with non-`table` role

Jym77:
As per https://github.com/act-rules/act-rules.github.io/pull/1968 / https://github.com/Siteimprove/sanshikan/pull/268


elenamongelli (approved):
looks like that there are unused variables

PR #1297, opened 2022-12-23
SIA-R63: guess MIME type, and always expose `<object>`

### LINKED ISSUE #1285: SIA-R63 should be less strict in the inclusion in the accessibility tree
SIA-R63 is currently strict in requesting the target `<object>` to be included in the accessibility tree.
However, things like `<object data="foo.png">` are not having any specific role per specs but are nonetheless exposed by UAs and ATs.

Alfa should probably apply to any `<object>` that is not explicitly hidden.

Jym77:
Resolves #1285 

Non-empty `<object>` elements are always exposed since their content normally needs to be exposed. They do not have a corresponding ARIA role, however.
It is likely that `<audio>` and `<video>` elements might need the same behaviour, but we'll add it later if needed.

Restrict R63 to `<object>` embedding audio, image, or video MIME types. We cannot actually figure out the correct type, so we rely on either the `type` attribute or the file extension of the URL in the `data` attribute. This is suboptimal, but should be enough to avoid the worse false positives.


packages/alfa-rules/src/sia-r63/rule.ts
```diff
@@ -55,3 +56,58 @@ export namespace Outcomes {
...
   );
 }
+
+function embedsMedia(element: Element): boolean {
+  return (
+    element
+      // If there is a type attribute, use it, even if the value is incorrect
+      .attribute("type")
+      .map((type) => MediaMIMEType.includes(type.value.split("/")[0]))
+      // Otherwise, look at the file extension in the URl in the data attribute.
```
elenamongelli, line 66:
```suggestion
      // Otherwise, look at the file extension in the URL in the data attribute.
```

elenamongelli (dismissed):

Issue #1205, opened 2022-07-26
ARIA 1.2 migration


Jym77:
- [x] Update the scripts to gather 1.2 data instead of 1.3

---
With ARIA 1.2, some attributes (`aria-label(ledby)`) are becoming prohibited on certain roles (that do not accept name from author). In turn, they (likely) do not trigger the presentational role conflict resolution due to being prohibited on a role of `presentation`.

- [x] Add prohibited attributes to the role model.
- [x] Add parsing of prohibited attributes in the ARIA parsing scripts.
- [x] Do not accept prohibited attributes in R18.

---
We can do better with the `generic` role:

- [x] Add role ~and attributes~ to `Container`.
- [x] map elements with a `generic` role to `Container`.
- [x] Update features to map to `generic` role where needed.

PR #1306, opened 2023-01-06
Add support for `generic` role

Jym77:
Part of #1205.

* Big clean up on `feature.ts` role detection to avoid the boilerplate `Option.of(Role.of(…))` all over the place.

---
* Include role in `Container` nodes, since some will be `generic` and some have no role.
* Expose elements whose role is `generic` as `Container` nodes (if no other interesting property).
* Update `Feature` to return `generic` role where needed (according to HTML AAM).

---
* Adapt tests.
* Add exception for testable statement with `generic` role, see https://github.com/w3c/accname/issues/180 we may ultimately decide to still compute name of nodes with prohibited naming, hiding it some way, but for now we skip these.

---
* Fix a bug in `hasExplicitRole` which was always false when no argument is provided.
* Add `<aside>` to the list of elements whose role depends on the presence of a name.
* Slightly unrelated refactor in `hasInputType` to allow for nicer role detection of `<input>` elements.


packages/alfa-aria/src/dom/predicate/has-incorrect-role-without-name.ts
```diff
@@ -9,18 +9,23 @@ const { hasName } = Element;
...
+ * Alfa currently can't handle that and always give them a role.
+ * {@link https://github.com/Siteimprove/alfa/issues/298}
  * @public
  */
 export function hasIncorrectRoleWithoutName(
   device: Device
 ): Predicate<Element> {
-  // <form> and <section> element have an implicit role only if they have an accessible name
+  // <aside>, <form> and <section> element have an implicit role only if they
+  // have an accessible name.
```
mvy-siteimprove, line 22: duplicated comment

mvy-siteimprove (approved):

PR #1310, opened 2023-01-16
Support ARIA prohibited attributes, and handle them in SIA-R18v2

### LINKED ISSUE #1205: ARIA 1.2 migration

Jym77:
Resolves #1205 

* Switch ARIA attribute collection to ARIA 1.2 (was forgotten in #1306).
* Add prohibited attributes to roles data and model.
* Streamline some method in `Role`, using already stored data 🙈 
* Add R18 version 2 as experimental, rejecting prohibited attributes (in a second expectation), following https://github.com/Siteimprove/sanshikan/pull/273
* Streamline looking for required attribute in SIA-R16, because now they are exposed by the role model.
* Refactor and streamline Diagnostic building in SIA-R16, it was a very outdated style…

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,3 +17,5 @@ updates:
     labels:
       - "actions"
       - "dependencies"
+    reviewers:
+      - "Siteimprove/alfa-owners"
--- a/.github/workflows/scan.yml
+++ b/.github/workflows/scan.yml
@@ -23,8 +23,8 @@ jobs:
           cache: yarn
       - run: yarn install --immutable
       - run: yarn build
-      - uses: github/codeql-action/init@v2.1.37
+      - uses: github/codeql-action/init@v2.1.38
         with:
           languages: javascript
           config-file: .github/codeql/config.yml
-      - uses: github/codeql-action/analyze@v2.1.37
+      - uses: github/codeql-action/analyze@v2.1.38


packages/alfa-rules/src/index.ts
```diff
@@ -82,9 +82,7 @@ export namespace Flattened {
  *
  * @public
  */
-export const Flattened: Flattened.Flattened = Sequence.from<Flattened.Rule>(
-  Rules.values()
-);
+export const Flattened = Sequence.from(Rules.values()) as Flattened.Flattened;
```
Jym77, line 85: This reverts a bit what was done in #1197 because that version causes the build to break on non-clean state, requiring to `clean` and re-build the full `alfa-rules` package, which is a annoyance 🙃

packages/alfa-rules/src/sia-er18/rule.ts
```diff
@@ -0,0 +1,132 @@
...
+          2: expectation(
+            test(
+              hasRole(device, (role) =>
+                role.isAttributeProhibited(target.name as aria.Attribute.Name)
+              ),
+              owner
+            ),
+            () => Outcomes.IsProhibited,
+            () => Outcomes.IsNotProhibited
+          ),
```
Jym77, line 67: This is the real new bit.

packages/alfa-rules/test/sia-er18/rule.spec.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+  const document = h.document([target]);
+
+  t.deepEqual(await evaluate(R18, { document }), [
+    failed(R18, target.attribute("aria-label").getUnsafe(), {
+      1: Outcomes.IsAllowed,
+      2: Outcomes.IsProhibited,
+    }),
+  ]);
+});
```
Jym77, line 209: This is the new test case.

rcj-siteimprove (approved):
LGTM

PR #1311, opened 2023-01-17
Colour contrast: do not consider ancestors as interposed descendants

Jym77:
This is branching from the old 0.52.1 so we can deploy it.
The actual changes are https://github.com/Siteimprove/alfa/compare/v0.52.1...f0e343ab279f32afe2c6e9ed6473bb8a6badfbca

**DO NOT UPDATE BRANCH**


packages/alfa-rules/src/common/expectation/contrast.ts
```diff
@@ -70,12 +70,14 @@ export function hasSufficientContrastExperimental(
   largeTextThreshold: number,
   normalTextThreshold: number
 ) {
-  // all non-transparent interposed descendants of an ancestor
+  // all non-transparent interposed descendants of an ancestor, except ancestors
```
mvy-siteimprove, line 73:
```suggestion
  // all non-transparent interposed descendants of target's ancestors, except ancestors
```

PR #1313, opened 2023-01-18
SIA-R19: force required ID ref (list) to point to discoverable elements

Jym77:
Following ACT rules and https://github.com/Siteimprove/sanshikan/pull/274

### DIFF
--- a/packages/alfa-rules/src/sia-r19/rule.ts
+++ b/packages/alfa-rules/src/sia-r19/rule.ts
@@ -1,7 +1,9 @@
 import { Rule, Diagnostic } from "@siteimprove/alfa-act";
+import { Cache } from "@siteimprove/alfa-cache";
 import { Attribute, Element, Namespace, Node } from "@siteimprove/alfa-dom";
 import { Iterable } from "@siteimprove/alfa-iterable";
 import { Predicate } from "@siteimprove/alfa-predicate";
+import { Refinement } from "@siteimprove/alfa-refinement";
 import { Ok, Err } from "@siteimprove/alfa-result";
 import { Sequence } from "@siteimprove/alfa-sequence";
 import { Page } from "@siteimprove/alfa-web";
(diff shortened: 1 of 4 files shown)


packages/alfa-rules/src/sia-r19/rule.ts
```diff
@@ -28,21 +32,22 @@ export default Rule.Atomic.of<Page, Attribute>({
           .filter(hasNamespace(Namespace.HTML, Namespace.SVG))
           .flatMap((element) =>
             Sequence.from(element.attributes).filter(
-              and(
-                property("name", aria.Attribute.isName),
-                property("value", not(isEmpty))
-              )
+              and(isAriaAttribute, property("value", not(isEmpty)))
```
Jym77, line 35: Using a `Refinement` allows to correctly narrow the targets to `Attribute<aria.Attribute.Name>`, which in turn let us get rid of that test in the Expectation.

packages/alfa-rules/src/sia-r19/rule.ts
```diff
@@ -97,3 +105,59 @@ function isValid(attribute: aria.Attribute): boolean {
...
+          .root()
+          .descendants()
+          .filter(isElement)
+          .collect((element) => element.id)
+      )
+    )
+    .has(id);
+}
+
+function isRequiredRefValid(
```
mvy-siteimprove, line 131: nitpick: perhaps would make sense to be less specific since we're handling also other-than required reference attributes?
```suggestion
function isAttributeValid(
```

mvy-siteimprove (approved):

packages/alfa-rules/src/sia-r19/rule.ts
Jym77: Struggled a bit to find a correct name 😅 At list the doc should make the intention explicit…

mvy-siteimprove (approved):

PR #1314, opened 2023-01-18
Downgrade SIA-R21, introduce SIA-R110

Jym77:
Following https://github.com/Siteimprove/sanshikan/pull/276

R21 is now only a best practice, since `role="btn button"` is not a WCAG failure but definitely a code smell.
R110 is the less strict version, requiring at least one valid token (rather than only valid tokens).


rcj-siteimprove (approved):
Looks good 👍

PR #1316, opened 2023-01-19
SIA-R14: now filtering out unicode symbols and some control characters 

### LINKED ISSUE #1220: Investigate considering more characters as non-text content
Many Unicode characters are actually non-text content but symbols (e.g., arrows, …) and could be considered as non-text content, notably for the purpose of R14.

We should find a suitable class of Unicode characters to ignore.

dan-tripp-siteimprove:
Resolves #1220


packages/alfa-rules/test/sia-r14/rule.spec.tsx
```diff
@@ -45,6 +45,21 @@ test(`evaluate() passes a <button> element whose perceivable text content includ
   ]);
 });
 
+test(`evaluate() passes a <button> element whose perceivable text content includes symbols and
+      control characters and is included in its accessible name set by aria-label`, async (t) => {
+  const target = <button aria-label="Hello +-|­▼ world">Hello­←</button>;
```
Jym77, line 50:
```suggestion
  const target = <button aria-label="Hello ← world">Hello­ +-|­▼</button>;
```
Since the rule check that the label is included in the name, I think it's a better test to have more garbage in the label (i.e `<button aria-label="Hello +-|­▼ world">Hello­</button>` would already pass the current version of the rule)

Jym77 (changes requested):
Small improvement.

Jym77 (approved):

PR #1318, opened 2023-01-23
Record outcome mode (automatic / semi-automatic)

Jym77:
* `Outcome` has a `mode` parameter.
* `Interview.conduct` records whether the Oracle produced a non-`None` answer.
* Rule evaluation records in its outcome whether the oracle was used.
* Update many rules test cases when oracle is used 🙈 

---
* Refactor the outcome value (passed/failed/…)


rcj-siteimprove (approved):
👍

Issue #1342, opened 2023-02-09
Quality of Life improvements


Jym77:
A bunch of tooling changes to improve our own QoL.

* [x] Use [pr-commanrd](https://github.com/Siteimprove/pr-command) to allow easier `yarn document` or `yarn dedupe` in PRs. Can also work for `yarn dependency-graph` (which could be added as test in CI).
* [x] Use some sort of dependency checker for missing or extra dependencies. Depcheck seems to struggle with monosrepos and workspaces, [Knip](https://github.com/webpro/knip) might be a good candidate. Likely include a step in CI to check that.
* [x] Automatically check that any internal dependency in a `package.json` is also included as a `reference.path` in the corresponding `tsconfig.json`, and vice-versa. Likely needed as a CI step.
* [x] Check that each package has an API Extractor config.
* [x] Improve changelog flow. [changesets](https://github.com/changesets/changesets) looks like a promising candidate to keep the changelog together with the PRs and merge them at release time.
* [x] Improve release flow to reduce the number of manual steps, [changesets](https://github.com/changesets/changesets) would likely improve that.
* [x] Improve generation of top-level changelog.
* [x] Auto-generate per-package dependency graphs, curated as needed (removing duplicate edges, cut edges at directory boundaries, …)

PR #1344, opened 2023-02-09
Set up PR commands

Jym77:
Part of #1342.

Trying to set up PR commands. Not fully sure the yarn setup is correct, we'll update if needed.


Jym77:
!pr help

PR #1346, opened 2023-02-14
Add guide on writing a parser

rcj-siteimprove:
This PR adds one guide, the folder structure and a link to the guide folder on the root README.md

### DIFF
--- a/docs/guides/writing-a-parser.md
+++ b/docs/guides/writing-a-parser.md
@@ -1,0 +1,2 @@
+# Writing a parser
+
--- a/docs/guides/writing-a-parser.md
+++ b/docs/guides/writing-a-parser.md
@@ -30,7 +30,11 @@ meaning that the token can either have the literal value of `shipping` or `billi
 
 ## Implementing the parser
 
-Our strategy is to define individual "atomic" parsers for each of the tokens and then use the functions available in the `alfa-parser` package to combine them into the final parser in such a way that it encodes the grammar defined above. The individual parsers will be very simple since they will only check that a token has one of the valid literal values for that token. If it does, it will return an `Ok` result with the parsed token and the rest of the tokens ready to be parsed by the next parser. If it doesn't, it will return an `Err` result with a message and the parsing will terminate.
+Out aim is to write a parser that consumes an array of tokens (in our case strings) and returns a result of either `Ok`, with the result and the rest of the tokens, or an `Err` result with a message containing relevant information on why the parsing failed.
+
+The strategy is:
+- write individual atomic parsers to parse each indivial token using string matching
+- combine the atomic parser using the combinators from `alfa-parser` into the final parser in such a way that it encodes the grammar
 
 Start by adding the following imports
 ```TS
@@ -40,7 +44,7 @@ import { Slice } from "@siteimprove/alfa-slice";
 
 const { either, end, option, right } = Parser;
 ```
-The types from `alfa-result` will be used by the parser to signal success or failure and the `Slice` type is used by the parser to consume an array of tokens in a memory efficient way. The functions, or combinators, from the `Parser` namespace that we will be using have been stored as `const`s for convenience.
+The types from `alfa-result` will be used by the parser to signal success or failure and the `Slice` type is used by the parser to consume an array of tokens in a memory efficient way. The functions, or combinators, from the `Parser` namespace that will be used have been stored as `const`s for convenience.
 
 We will implement two of the "atomic" parsers and leave out the rest as they are very similar. In the actual implementation in [#1339][PR] this was done in a more general way where a function generates the parser from an array of literal values.
 
--- a/docs/guides/writing-a-parser.md
+++ b/docs/guides/writing-a-parser.md
@@ -34,7 +34,7 @@ Our aim is to write a parser that consumes an array of tokens (in our case strin
 
 The strategy is:
 - write individual atomic parsers to parse each indivial token using string matching
-- combine the atomic parser using the combinators from `alfa-parser` into the final parser in such a way that it encodes the grammar
+- combine the atomic parsers using the combinators from `alfa-parser` into the final parser in such a way that it encodes the grammar
 
 Start by adding the following imports
 ```TS
--- a/docs/guides/writing-a-parser.md
+++ b/docs/guides/writing-a-parser.md
@@ -1,2 +1,43 @@
 # Writing a parser
 
+This guide shows how to implement a simple parser using the [alfa-parser](https://github.com/Siteimprove/alfa/tree/main/packages/alfa-parser) package. The guide is based on the work done in https://github.com/Siteimprove/alfa/pull/1339.
+
+## Background
+
+The `alfa-parser` package contains functionality to make it easy to implement a parser. This package is used internally by Alfa to parse HTML, CSS etc., but it can be used for any parsing scenario.
+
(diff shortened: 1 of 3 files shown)


docs/guides/writing-a-parser.md
```diff
@@ -0,0 +1,108 @@
...
+```HTML
+<input autocomplete="section-blue billing home" />
+```
+This is invalid because the `home` token must be followed by a required token.
+
+We can express the grammar as follows
+```
+[section] [addressType] (unmodifiable | [modifier] modifiable) [webauthn] EOF
+```
+where a token name sorounded by square brackets `[]` means it is an optional token, tokens seperated by a vertical line `|` means exactly one of the tokens must be present and `EOF` is a special token signaling that no more tokens are allowed.
```
Jym77, line 23:
```suggestion
where a token name surrounded by square brackets `[]` means it is an optional token, tokens separated by a vertical line `|` means exactly one of the tokens must be present and `EOF` is a special token signalling that no more tokens are allowed.
```

docs/guides/writing-a-parser.md
```diff
@@ -0,0 +1,108 @@
...
+
+Here the possible values for the tokens are not listed, but as an example `addressType` is defined as follows
+```
+addressType = "shipping" | "billing"
```
Jym77, line 27:
```suggestion
addressType ::= "shipping" | "billing"
```
Keeping it closer to [BNF](https://en.wikipedia.org/wiki/Backus%E2%80%93Naur_form) syntax

docs/guides/writing-a-parser.md
```diff
@@ -0,0 +1,108 @@
+# Writing a parser
+
+This guide shows how to implement a simple parser using the [alfa-parser](https://github.com/Siteimprove/alfa/tree/main/packages/alfa-parser) package. The guide is based on the work done in [#1339][PR].
+
+## Background
+
+The `alfa-parser` package contains functionality to make it easy to implement a parser. This package is used internally by Alfa to parse HTML, CSS etc., but it can be used for any parsing scenario.
```
Jym77, line 7:
```suggestion
The `alfa-parser` package contains functionality to make it easy to implement a parser ([parsers combinators](https://en.wikipedia.org/wiki/Parser_combinator)). This package is used internally by Alfa to parse HTML, CSS etc., but it can be used for any parsing scenario.
```

docs/guides/writing-a-parser.md
```diff
@@ -0,0 +1,108 @@
...
+
+Here the possible values for the tokens are not listed, but as an example `addressType` is defined as follows
+```
+addressType = "shipping" | "billing"
+```
+meaning that the token can either have the literal value of `shipping` or `billing`.
+
+## Implementing the parser
+
+Our strategy is to define individual "atomic" parsers for each of the tokens and then use the functions available in the `alfa-parser` package to combine them into the final parser in such a way that it encodes the grammar defined above. The individual parsers will be very simple since they will only check that a token has one of the valid literal values for that token. If it does, it will return an `Ok` result with the parsed token and the rest of the tokens ready to be parsed by the next parser. If it doesn't, it will return an `Err` result with a message and the parsing will terminate.
```
Jym77, line 33: I'd rephrase this a bit in the different order:
* What is the `Parser` type (consume an input, here an array of tokens (string); return a `result` which is either an `Ok` with both the remaining input and the actual result, or an `Err` with any relevant information). (more or less the last two sentence)
* Writing parser strategy: atomic parsers + combinators.
* Here, atomic parser = string matching.

docs/guides/writing-a-parser.md
```diff
@@ -0,0 +1,108 @@
...
+
+Start by adding the following imports
+```TS
+import { Parser } from "@siteimprove/alfa-parser";
+import { Err, Ok } from "@siteimprove/alfa-result";
+import { Slice } from "@siteimprove/alfa-slice";
+
+const { either, end, option, right } = Parser;
+```
+The types from `alfa-result` will be used by the parser to signal success or failure and the `Slice` type is used by the parser to consume an array of tokens in a memory efficient way. The functions, or combinators, from the `Parser` namespace that we will be using have been stored as `const`s for convenience.
```
Jym77, line 43:
```suggestion
The types from `alfa-result` will be used by the parser to signal success or failure and the `Slice` type is used by the parser to consume an array of tokens in a memory efficient way. The functions, or combinators, from the `Parser` namespace that we will be used have been stored as `const`s for convenience.
```

docs/guides/writing-a-parser.md
```diff
@@ -0,0 +1,108 @@
...
+
+We will implement two of the "atomic" parsers and leave out the rest as they are very similar. In the actual implementation in [#1339][PR] this was done in a more general way where a function generates the parser from an array of literal values.
+
+```TS
+const segment: Parser<Slice<string>, string, string> = (input: Slice<string>) => {
```
Jym77, line 48:
```suggestion
const section: Parser<Slice<string>, string, string> = (input: Slice<string>) => {
```

docs/guides/writing-a-parser.md
```diff
@@ -0,0 +1,108 @@
...
+}
+```
+
+Now let's imagine we also implemented the other parsers in a similar fashion. We are now ready to combine all the parsers using the parser combinators `option`, `either`, `right` and `end` to get the final parser.
+
+```TS
+const parse = right(
+  option(section), // The attribute should optionally begin with a section token
+  right(
+    option(addressType), Next, optionally an addressType token
```
Jym77, line 81:
```suggestion
    option(addressType), // Next, optionally an addressType token
```

Jym77 (changes requested):
A few typo and one suggestion on a paragraph.

docs/guides/writing-a-parser.md
```diff
@@ -0,0 +1,112 @@
...
+
+Here the possible values for the tokens are not listed, but as an example `addressType` is defined as follows
+```
+addressType ::= "shipping" | "billing"
+```
+meaning that the token can either have the literal value of `shipping` or `billing`.
+
+## Implementing the parser
+
+Out aim is to write a parser that consumes an array of tokens (in our case strings) and returns a result of either `Ok`, with the result and the rest of the tokens, or an `Err` result with a message containing relevant information on why the parsing failed.
```
rcj-siteimprove, line 33:
```suggestion
Our aim is to write a parser that consumes an array of tokens (in our case strings) and returns a result of either `Ok`, with the result and the rest of the tokens, or an `Err` result with a message containing relevant information on why the parsing failed.
```

docs/guides/writing-a-parser.md
```diff
@@ -20,17 +20,21 @@ We can express the grammar as follows
...
 
 ## Implementing the parser
 
-Our strategy is to define individual "atomic" parsers for each of the tokens and then use the functions available in the `alfa-parser` package to combine them into the final parser in such a way that it encodes the grammar defined above. The individual parsers will be very simple since they will only check that a token has one of the valid literal values for that token. If it does, it will return an `Ok` result with the parsed token and the rest of the tokens ready to be parsed by the next parser. If it doesn't, it will return an `Err` result with a message and the parsing will terminate.
+Our aim is to write a parser that consumes an array of tokens (in our case strings) and returns a result of either `Ok`, with the result and the rest of the tokens, or an `Err` result with a message containing relevant information on why the parsing failed.
+
+The strategy is:
+- write individual atomic parsers to parse each indivial token using string matching
+- combine the atomic parser using the combinators from `alfa-parser` into the final parser in such a way that it encodes the grammar
```
Jym77, line 37:
```suggestion
- combine the atomic parsers using the combinators from `alfa-parser` into the final parser in such a way that it encodes the grammar
```

Jym77 (dismissed):
Small typo.

Jym77 (approved):

PR #1351, opened 2023-02-21
Parse `:host` pseudo-class selector

### LINKED ISSUE #1347: Add basic support for `:host` pseudo-class
When encountering a [`:host` pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:host), Alfa just fails to parse it and discard the full rule. This has caused trouble for, e.g., a `:host, :root` selector where Alfa discards the full rule after failing to parse the selector and therefore doesn't apply the declaration to the `:root` element.

We should add basic support of `:host`, have it match nothing at first, but at least parse it and keep the declarations.

- [ ] Follow-up on DEV-14044

rcj-siteimprove:
Resolves #1347

Updated the pseudo class parser to parse the `:host` selector.


rcj-siteimprove:
Accidentally branched from an old branch. Closing this PR and trying again.

PR #1352, opened 2023-02-21
Parse `:host` pseudo-class selector

### LINKED ISSUE #1347: Add basic support for `:host` pseudo-class

rcj-siteimprove:
Resolves #1347

Updated the pseudo class parser to parse the `:host` selector.

### DIFF
--- a/packages/alfa-selector/src/selector.ts
+++ b/packages/alfa-selector/src/selector.ts
@@ -1503,11 +1503,6 @@ export namespace Selector {
       super("host");
     }
 
-    public matches(_element: Element): boolean {
-      // The shadow DOM is currently not supported so there should not be any matches.
-      // If support for the shadow DOM is added, then this should be implemented.
-      return false;
-    }
   }
 
   /**
--- a/packages/alfa-selector/src/selector.ts
+++ b/packages/alfa-selector/src/selector.ts
@@ -884,6 +884,8 @@ export namespace Selector {
             return Result.of(Enabled.of());
           case "root":
             return Result.of(Root.of());
+          case "host":
+            return Result.of(Host.of());
           case "empty":
             return Result.of(Empty.of());
           case "first-child":
@@ -1486,6 +1488,25 @@ export namespace Selector {
     public matches(element: Element): boolean {
       // The root element is the element whose parent is NOT itself an element.
       return element.parent().every(not(isElement));
+    }
+  }
+
+  /**
+   * {@link https://drafts.csswg.org/css-scoping-1/#selectordef-host}
+   */
+  export class Host extends Pseudo.Class<"host"> {
+    public static of(): Host {
+      return new Host();
+    }
+
+    private constructor() {
+      super("host");
+    }
+
+    public matches(_element: Element): boolean {
+      // The shadow DOM is currently not supported so there should not be any matches.
+      // If support for the shadow DOM is added, then this should be implemented.
+      return false;
     }
   }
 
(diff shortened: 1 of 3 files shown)


packages/alfa-selector/src/selector.ts
```diff
@@ -1489,6 +1491,25 @@ export namespace Selector {
...
+
+    private constructor() {
+      super("host");
+    }
+
+    public matches(_element: Element): boolean {
+      // The shadow DOM is currently not supported so there should not be any matches.
+      // If support for the shadow DOM is added, then this should be implemented.
+      return false;
+    }
```
Jym77, line 1510:
```suggestion
```
We can just remove it as [the super-class already does so](https://github.com/Siteimprove/alfa/blob/04373c2c37feafdea0c8926d7ac1d69538eee368/packages/alfa-selector/src/selector.ts#L751-L753) I do not feel the comment is adding that much given that it is a fairly common pattern we have for selectors we don't really support.

packages/alfa-selector/test/selector.spec.tsx
```diff
@@ -692,6 +692,20 @@ test(".parse() parses a named pseudo-class selector", (t) => {
...
 
+test(".parse() parses :host pseudo-class selector", (t) => {
+  const parseResult = parse(":host");
+
+  if (parseResult.isErr()) {
+    t.fail(parseResult.getErr());
+    return;
+  }
+
+  t.deepEqual(parseResult.get().toJSON(), {
```
Jym77, line 703:
```suggestion
  t.deepEqual(parse(":host").get().toJSON(), {
```
We don't really need the first test. If the test throw an exception, that's almost as good as failing. Other tests in this file seem to use this pattern.

Jym77 (changes requested):
Simplifying a bit.

Jym77:
!pr help

siteimprove-builduser:
Hi there 👋 

 You have the following options: 
 - `!pr help` writes this comment
 - `!pr dedupe` Deduplicate packages
 - `!pr depgraph` Update dependency graph
 - `!pr document` Update packages documentation

packages/alfa-selector/src/selector.ts
```diff
@@ -1489,6 +1491,20 @@ export namespace Selector {
...
+   */
+  export class Host extends Pseudo.Class<"host"> {
+    public static of(): Host {
+      return new Host();
+    }
+
+    private constructor() {
+      super("host");
+    }
+
```
rcj-siteimprove, line 1505:
```suggestion
    }
```

Jym77 (approved):

rcj-siteimprove:
!pr document

Jym77:
!pr version

Jym77:
!pr version

Jym77:
!pr version

Jym77:
!pr version

Jym77:
!pr version

Jym77:
!pr version

Jym77:
!pr version

Jym77:
!pr document

Jym77 (approved):

PR #1354, opened 2023-02-21
Update whitespace parser to parse one or more

### LINKED ISSUE #1291: CSS variable substitution should collapse whitespace tokens
```css
--foo:   1
color: rbg(255,   255, 255,    var(--foo))
```
When lexing the value for `color`, adjacent whitespace are collapsed into one. Same goes when lexing `--foo`
However, when substituting `var(--foo)`, both the existing whitespace from `color` and the leading one from `--foo` are kept. This result in two `Token.Whitespace` that the parser for `color` does not expect, resulting in an invalid parsing 🙈 

* We could fix the parser for `color` to accept several whitespace in row, but it would only solve the problem for this property, not for others.
* We could tweak `Token.parseWhitespace` to allow several whitespace in a row, this seems a bit hackyk
* A maybe better but trickier way would be to tweak `substitute` so that it detects whitespace duplication and clean them up.
* Alternatively, `parseLonghand` (and `parseShorthand`) could do a cleaning step post-substitution to de-duplicate adjacent whitespace; this is possibly more costly as we'd need to go through the `Slice` one more time.

- [ ] Follow up on DEV-13826

rcj-siteimprove:
Resolves #1291 

Update `parseWhitespace` so that it accepts one or more whitespace tokens to allow for the possibility that the lexer didn't manage to collapse all whitespace prior to the parsing. This might affect performance due to the extra check for more whitespaces and since this affects all users of the parser. If the performance hit is deemed unacceptable, we will explore other solutions.

### DIFF
--- a/packages/alfa-css/src/syntax/token.ts
+++ b/packages/alfa-css/src/syntax/token.ts
@@ -8,6 +8,7 @@ import { Slice } from "@siteimprove/alfa-slice";
 
 import * as json from "@siteimprove/alfa-json";
 
+const { oneOrMore } = Parser;
 const { fromCharCode } = String;
 const { and } = Refinement;
 
@@ -811,7 +812,7 @@ export namespace Token {
 
   export const { of: whitespace, isWhitespace } = Whitespace;
 
-  export const parseWhitespace = parseToken(isWhitespace);
+  export const parseWhitespace = oneOrMore(parseToken(isWhitespace));
 
   export class Colon implements Equatable, Serializable<Colon.JSON> {
     private static _instance = new Colon();
--- a/packages/alfa-css/src/syntax/token.ts
+++ b/packages/alfa-css/src/syntax/token.ts
@@ -812,11 +812,9 @@ export namespace Token {
 
   export const { of: whitespace, isWhitespace } = Whitespace;
 
-  // If more than one whitespace is parsed, we only want one in the result.
-  // Calling `oneOrMore` ensures at least one, so indexing should be safe.
   export const parseWhitespace = map(
     oneOrMore(parseToken(isWhitespace)),
-    (whitespaces) => whitespaces[0]
+    ([first,]) => first
   );
 
   export class Colon implements Equatable, Serializable<Colon.JSON> {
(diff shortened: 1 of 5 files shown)


rcj-siteimprove:
!pr document

rcj-siteimprove:
!pr document

Jym77 (changes requested):
Actually thinking that we should upgrade the type of `oneOrMore` to return an array with at least one element (`[T, ...Array<T>]`). That would probably require a cast there, but it would improve the resulting type.

rcj-siteimprove:
> Actually thinking that we should upgrade the type of `oneOrMore` to return an array with at least one element (`[T, ...Array<T>]`). That would probably require a cast there, but it would improve the resulting type.

That sounds nice, I'll give it a go.

packages/alfa-parser/src/parser.ts
```diff
@@ -106,8 +106,9 @@ export namespace Parser {
 
   export function oneOrMore<I, T, E, A extends Array<unknown> = []>(
     parser: Parser<I, T, E, A>
-  ): Parser<I, Array<T>, E, A> {
-    return takeAtLeast(parser, 1);
+  ): Parser<I, [T, ...Array<T>], E, A> {
+    // The result contains at least one token, so this cast is safe
+    return map(takeAtLeast(parser, 1), (result) => result as [T, ...Array<T>]);
```
Jym77, line 111: Can't we
```suggestion
    return takeAtLeast(parser, 1) as Parser<I, [T, ...Array<T>], E, A>;
```
(we can even `as any as …` if needed, since we do have knowledge that the typesystem doesn't have)

packages/alfa-css/src/syntax/token.ts
```diff
@@ -811,7 +812,10 @@ export namespace Token {
 
   export const { of: whitespace, isWhitespace } = Whitespace;
 
-  export const parseWhitespace = parseToken(isWhitespace);
+  export const parseWhitespace = map(
+    oneOrMore(parseToken(isWhitespace)),
+    ([first,]) => first
```
Jym77, line 817: I think the comma is not needed.
```suggestion
    ([first]) => first
```

Jym77 (changes requested):
Small polish.

Jym77 (approved):

PR #1358, opened 2023-02-23
SIA-R16: Make `aria-controls` optional for role `combobox`

### LINKED ISSUE #1340: SIA-R16 should accept (closed) `combobox` without `aria-controls`.
Similar to what was done for R19 in #1313, and for the same reasons, `combobox` without `aria-controls` should be considered valid if `aria-expanded="false"`.

See also https://github.com/act-rules/act-rules.github.io/issues/2029

rcj-siteimprove:
Resolves #1340 

Element with role `combobox` and missing `aria-controls` now passes SIA-R16 regardless of the state of `aria-expanded`. In the future we might refine this so that it won't pass when the combobox is open.

### DIFF
--- a/packages/alfa-rules/src/sia-r16/rule.ts
+++ b/packages/alfa-rules/src/sia-r16/rule.ts
@@ -274,12 +274,8 @@ function isAttributeOptional(
   role: aria.Role.Name,
   attribute: aria.Attribute.Name
 ) {
-  if (role === "combobox" && attribute === "aria-controls") {
-    // combobox only requires aria-controls when opened, which we can't detect
-    // same as for sia-r19, @see https://github.com/Siteimprove/alfa/pull/1313
-    // should be refined to look at `aria-expanded` at some point
-    return true;
-  }
-
-  return false;
+  // combobox only requires aria-controls when opened, which we can't detect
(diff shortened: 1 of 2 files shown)


packages/alfa-rules/src/sia-r16/rule.ts
```diff
@@ -268,3 +269,17 @@ function isManagedAttribute(
...
+) {
+  if (role === "combobox" && attribute === "aria-controls") {
+    // combobox only requires aria-controls when opened, which we can't detect
+    // same as for sia-r19, @see https://github.com/Siteimprove/alfa/pull/1313
+    // should be refined to look at `aria-expanded` at some point
+    return true;
+  }
+
+  return false;
+}
```
Jym77, line 285:
```suggestion
    // combobox only requires aria-controls when opened, which we can't detect
    // same as for sia-r19, @see https://github.com/Siteimprove/alfa/pull/1313
    // should be refined to look at `aria-expanded` at some point
  return role === "combobox" && attribute === "aria-controls"
}
```

Jym77 (changes requested):
Streamlining conditional.

Jym77 (approved):

PR #1361, opened 2023-02-27
Write debugging guide

Jym77:


docs/guides/debugging.md
```diff
@@ -0,0 +1,220 @@
...
+
+(the file is saved as `.ts`, but, after building, it is the `.js` that must be run)
+
+The rule to check, and the file to load, are configured after the `import` block. The Alfa page is de-serialized from the JSON file. Then a simple audit of a given rule is run and simple statistics are shown, which allow to confirm that the numbers are the same…
+
+## Time travel
+
+Since we run into the Github repo, it is very easy to test with an older version of Alfa. As long as only the `scratches` directory has been changed, just `git checkout vX.Y.Z` will go back to any version (`yarn install` and `yarn build` need to be rerun). This makes it easy to check when a breaking change has been introduced. Be wary that by going back too far in time, the `scratches/page.ts` script itself might break if it's using more modern constructions…
+
+In case of regression, or change of behavior, finding the precise version where it was introduced and then searching for cues in the [Changelog](../../CHANGELOG.md) is often a good way to narrow dowrn the possible cause.
```
rcj-siteimprove, line 83: Made me think of [`git bisect`](https://git-scm.com/docs/git-bisect) which does a binary search of the history to find the commit where e.g. a bug was introduced. It can be used interactively where the user manually checks if the bug is present and it can even be automated which sounds like something we could do with the scratch file as long as it doesn't break. Just a thought, doesn't need to be included in the guide 🙂

rcj-siteimprove (dismissed):
Very nice guide! 👍

rcj-siteimprove (approved):

PR #1365, opened 2023-03-03
Fix bug in whitespace parser due to incorrect use of `Slice` type

rcj-siteimprove:
Fix to a bug uncovered after the release of [v0.60.0](https://github.com/Siteimprove/alfa/releases/tag/v0.60.0) where parsing whitespace could sometimes go into an infinite loop because of incorrect use of the `Slice` type. More specifically, when parsing an empty slice of an array, with more whitespace tokens in the array after the end of the slice, the parser would try to parse the whitespace after the slice and return the same empty array to be parsed again. This caused the bug because we now parse an unlimited number of whitespaces, instead of just one.

### DIFF
--- a/packages/alfa-css/src/syntax/token.ts
+++ b/packages/alfa-css/src/syntax/token.ts
@@ -1350,15 +1350,13 @@ export namespace Token {
 function parseToken<T extends Token>(
   refinement: Refinement<Token, T>
 ): Parser<Slice<Token>, T, string> {
-  return (input) => {
-    const result = input
+  return (input) =>
+    input
       .first()
-      .andThen((token) =>
+      .map((token) =>
         refinement(token)
-          ? Option.of(Ok.of<[Slice<Token>, T]>([input.rest(), token]))
-          : None
-      );
-
-    return result.isSome() ? result.get() : Err.of("Expected token");
-  };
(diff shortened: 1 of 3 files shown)


packages/alfa-css/src/syntax/token.ts
```diff
@@ -1348,12 +1351,14 @@ function parseToken<T extends Token>(
...
+    const result = input
+      .first()
+      .andThen((token) =>
+        refinement(token)
+          ? Option.of(Ok.of<[Slice<Token>, T]>([input.rest(), token]))
+          : None
+      );
 
-    return Err.of("Expected token");
+    return result.isSome() ? result.get() : Err.of("Expected token");
```
Jym77, line 1362:
```
return input
      .first()
      .map((token) =>
        refinement(token)
        ? Ok.of<[Slice<Token>, T]>([input.rest(), token])
          : Err.of("Mismatching token")
).getOr(Err.ef("No token left")
```

Jym77 (changes requested):
Polishing a bit.

packages/alfa-css/src/syntax/token.ts
rcj-siteimprove: Yeah, that looks nicer 👍

Jym77 (approved):

PR #1366, opened 2023-03-03
Cache descendants of trees

Jym77:
* Cache `Node.descendants()` in `alfa-tree`. This is a common use case; even if the sequence is lazy, actually building it requires recursing through the full tree, which takes time.
* Cache (and create) `Document.elementDescendants`. Many rules, and some other piece of code, gather all elements of a DOM tree, this saves a lot of time.

### DIFF
--- a/packages/alfa-aria/src/feature.ts
+++ b/packages/alfa-aria/src/feature.ts
@@ -160,7 +160,7 @@ const labels = Cache.empty<Node, Sequence<Element>>();
 const nameFromLabel = (element: Element, device: Device, state: Name.State) => {
   const root = element.root();
 
-  const elements = root.inclusiveDescendants().filter(isElement);
+  const elements = root.elementDescendants();
 
   const isFirstReference = element.id.some((id) =>
     ids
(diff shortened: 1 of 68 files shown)


packages/alfa-aria/src/feature.ts
```diff
@@ -160,7 +160,7 @@ const labels = Cache.empty<Node, Sequence<Element>>();
 const nameFromLabel = (element: Element, device: Device, state: Name.State) => {
   const root = element.root();
 
-  const elements = root.inclusiveDescendants().filter(isElement);
+  const elements = root.elementDescendants();
```
rcj-siteimprove, line 163: Is it on purpose that `.inclusiveDescendants()` is gone? It looks like it means the root element won't be included anymore, but I might be overlooking something

packages/alfa-tree/src/tree.ts
```diff
@@ -135,16 +135,25 @@ export abstract class Node<
...
+  // Caching it saves a lot of time by generating the sequence only once.
   public descendants(options?: Flags<F>): Sequence<Node<F>> {
-    return this.children(options).flatMap((child) =>
-      Sequence.of(
-        child,
-        Lazy.of(() => child.descendants(options))
-      )
-    );
+    if (this._descendants[options?.value ?? 0] === undefined) {
+      this._descendants[options?.value ?? 0] = this.children(options).flatMap(
```
rcj-siteimprove, line 147: Would it make sense to store `options?.value ?? 0` in a variable?

rcj-siteimprove (approved):
Very nice. Just have a few questions.

rcj-siteimprove (approved):

PR #1367, opened 2023-03-06
Remove `array` property from `Slice`

rcj-siteimprove:
This PR contains the following:
- The `array` and `offset` property of `Slice` has been removed as using them could result in unexpected behavior as described in #1365.
- Update and simplification of call sites using the `array` property
- Extraction of conditional parsing in `parseToken` into parser combinator `parseIf`


rcj-siteimprove:
!pr document

packages/alfa-css/src/syntax/block.ts
```diff
@@ -100,27 +102,38 @@ export namespace Block {
...
+
+        if (next.get().type === token.mirror.type) {
+          input = input.rest();
+          break;
+        }
+
+        const [remainder, component] = Component.consume(input).get();
+
+        input = remainder;
+        value.push(...component);
```
Jym77, line 130: We could probably rneplace that with a `takeUntil(Component.consume, close)` 🤔

packages/alfa-css/test/syntax/block.spec.ts
```diff
@@ -0,0 +1,70 @@
+import { Slice } from "@siteimprove/alfa-slice";
+import { test } from "@siteimprove/alfa-test";
+
+import { Block, Token } from "../../src/";
+
+const { consume } = Block;
+const { whitespace, openParenthesis, closeParenthesis } = Token;
+
+test(".consume() of empty returns error", (t) => {
+  t.deepEqual(consume(Slice.empty()).toJSON(), {
```
Jym77, line 10: Tests would be easier to write and read by using `lexer.lex`.

packages/alfa-rules/src/sia-r10/rule.ts
```diff
@@ -177,29 +177,27 @@ const webauthn = parserOf(["webauthn"]);
...
-    if (token !== undefined && tokens.includes(token)) {
-      return Result.of([input.slice(1), token]);
-    }
-
-    return Err.of(`Expected valid token, but got ${input.toJSON()}`);
-  };
+  return (input) =>
+    input
+      .first()
+      .filter((token) => tokens.includes(token))
```
Jym77, line 183: This looks like `parseToken(tokens => tokens.include(token))`

packages/alfa-style/src/style.ts
```diff
@@ -542,7 +542,7 @@ function substitute(
   let substituted = false;
 
   while (tokens.length > 0) {
-    const next = tokens.array[tokens.offset];
+    const next = tokens.first().getUnsafe(); // Safe, we just checked that `tokens` is not empty
```
Jym77, line 545:
```suggestion
    // `tokens` is not empty due to the previous test.
    const next = tokens.first().getUnsafe(); 
```

Jym77 (changes requested):
Using `lex` in the test will make it easier to maintain. The rest is less important polish.

packages/alfa-parser/src/parser.ts
```diff
@@ -343,6 +344,26 @@ export namespace Parser {
...
+      input
+        .first()
+        .map((token) =>
+          refinement(token)
+            ? Result.of<[Slice<I>, T], string>([input.rest(), token])
+            : Err.of("Mismatching token")
+        )
+        .getOr(Err.of("No token left"));
+  }
+
```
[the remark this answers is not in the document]
Jym77: A `Predicate<T>` is just a `Refinement<T, T>`, so that should work out of the box.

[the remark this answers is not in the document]
Jym77: True.
This is however fairly easy, since it is only `input => input.first().map(first => Ok.of([input.rest(), first])).getOr(Err.of("Empty input"))`.
If needed, it might be better to add a `Slice#split` helper that does the `input.first().map(first => [input.rest(), first])` bit, and then just call `input.split().map(Ok.of).getOr(Err.of("Empty input"))`. But that still feels like putting a bit too more dependency inside `Slice` (it only cares about `split` because of our usage of `Slice` is in `Parser` that needs a `[I, T]` output).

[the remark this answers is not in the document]
rcj-siteimprove: Yeah, and it was only a few places after all, so I have just duplicated it for now, then we can revisit later.

packages/alfa-css/src/syntax/block.ts
```diff
@@ -98,29 +99,53 @@ export namespace Block {
     | Token.CloseCurlyBracket;
 
   /**
-   * {@link https://drafts.csswg.org/css-syntax/#consume-a-simple-block}
+   * Consumes components delimited by parentheses, square brackets or curly
+   * brackets.
+   *
+   * @remarks
+   * The compiler doesn't check if the delimiter types match, that needs to be
+   * ensured by the caller of the function.
```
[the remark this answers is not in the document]
rcj-siteimprove: I'll add it to the list 😄

packages/alfa-css/src/syntax/token.ts
```diff
@@ -1346,16 +1346,14 @@ export namespace Token {
   export const parseCloseComment = parseToken(isCloseComment);
 }
 
-function parseToken<T extends Token>(
-  refinement: Refinement<Token, T>
-): Parser<Slice<Token>, T, string> {
-  return (input) =>
+function parseFirst(): Parser<Slice<Token>, Token, string> {
+  return (input: Slice<Token>) =>
```
Jym77, line 1350:
```suggestion
const parseFirst = (input: Slice<Token>) =>
```
(no need to create an extra continuation)

packages/alfa-css/src/syntax/token.ts
```diff
@@ -1346,16 +1346,14 @@ export namespace Token {
...
-        refinement(token)
-          ? Ok.of<[Slice<Token>, T]>([input.rest(), token])
-          : Err.of("Mismatching token")
-      )
+      .map((token) => Ok.of<[Slice<Token>, Token]>([input.rest(), token]))
       .getOr(Err.of("No token left"));
 }
+
+function parseToken<T extends Token>(refinement: Refinement<Token, T>) {
+  return parseIf(refinement, parseFirst(), (_) => "Mismatching token");
```
Jym77, line 1358:
```suggestion

function parseToken<T extends Token>(refinement: Refinement<Token, T>) {
  return parseIf(refinement, parseFirst, () => "Mismatching token");
```
(`_` doesn't have a special meaning in ES, so if we don't need more parameters, we shouldn't create one)

packages/alfa-rules/src/sia-r10/rule.ts
```diff
@@ -174,32 +174,30 @@ const modifiable = parserOf(modifiables);
...
 const webauthn = parserOf(["webauthn"]);
 
+function parseFirst(): Parser<Slice<string>, string, string> {
+  return (input: Slice<string>) =>
+    input
+      .first()
+      .map((token) => Ok.of<[Slice<string>, string]>([input.rest(), token]))
+      .getOr(Err.of("No token left"));
+}
+
```
Jym77, line 184:
```suggestion
const parseFirst = (input: Slice<string>) =>
    input
      .first()
      .map((token) => Ok.of<[Slice<string>, string]>([input.rest(), token]))
      .getOr(Err.of("No token left"));

```

packages/alfa-rules/src/sia-r10/rule.ts
```diff
@@ -174,32 +174,30 @@ const modifiable = parserOf(modifiables);
...
-
-    if (token !== undefined && tokens.includes(token)) {
-      return Result.of([input.slice(1), token]);
-    }
-
-    return Err.of(`Expected valid token, but got ${input.toJSON()}`);
-  };
+  return parseIf(
+    (token): token is string => tokens.includes(token),
+    parseFirst(),
```
Jym77, line 190:
```suggestion
    parseFirst,
```

packages/alfa-rules/src/sia-r10/rule.ts
```diff
@@ -174,32 +174,30 @@ const modifiable = parserOf(modifiables);
...
-      return Result.of([input.slice(1), token]);
-    }
-
-    return Err.of(
-      `Expected token beginning with \`section-\`, but got ${input}`
-    );
-  };
+  return parseIf(
+    (token): token is string => token.startsWith("section-"),
+    parseFirst(),
```
Jym77, line 198:
```suggestion
    parseFirst,
```

packages/alfa-xpath/src/syntax/token.ts
```diff
@@ -423,16 +424,14 @@ export namespace Token {
...
-    if (token !== undefined && refinement(token)) {
-      return Result.of([input.slice(1), token]);
-    }
+function parseFirst(): Parser<Slice<Token>, Token, string> {
+  return (input: Slice<Token>) =>
+    input
+      .first()
+      .map((token) => Ok.of<[Slice<Token>, Token]>([input.rest(), token]))
+      .getOr(Err.of("No token left"));
+}
```
Jym77, line 433:
```suggestion
const parseFirst = (input: Slice<Token>) =>
    input
      .first()
      .map((token) => Ok.of<[Slice<Token>, Token]>([input.rest(), token]))
      .getOr(Err.of("No token left"));
```

packages/alfa-xpath/src/syntax/token.ts
```diff
@@ -423,16 +424,14 @@ export namespace Token {
...
 
-    return Err.of("Expected token");
-  };
+function parseToken<T extends Token>(refinement: Refinement<Token, T>) {
+  return parseIf(refinement, parseFirst(), (_) => "Mismatching token");
```
Jym77, line 436:
```suggestion
  return parseIf(refinement, parseFirst, (_) => "Mismatching token");
```

Jym77 (changes requested):
Polishing a bit. Looks good.

rcj-siteimprove:
!pr extract

Jym77 (approved):

PR #1370, opened 2023-03-07
Extract integrations

Jym77:
Moving integrations packages to https://github.com/Siteimprove/alfa-integrations
* These are the packages that heavily depend on external dependencies (the original tool they integrate with), extracting them reduces footprint of the main repository (and faster `yarn install`).
* These are "leaves" packages that rarely get updated but get built every time (faster `yarn build`, more focused view for developpers).

This is still the main repository for Alfa, issues related to integrations will still be tracked here for now.


rcj-siteimprove (approved):
Looks good 🎉

PR #290, opened 2020-07-03
Set up API Extractor for generating API documentation

kasperisager:
This pull request sets up API Extractor for generating API documentation as Markdown files. API Extractor was chosen as it seemed the most mature and feature-complete of the options available as outlined in #135. Additionally, it's backed by @microsoft and used for large projects such as the SharePoint Framework (https://docs.microsoft.com/en-us/javascript/api/sp-core-library).

In addition to generating API documentation, API Extractor also generates API report files that extract and track the public API surface exposed across packages: https://github.com/Siteimprove/alfa/blob/api-extractor/docs/review/api. During CI, API Extractor can then fail the build if the public API changes.

There are still some outstanding issues in API Extractor that we need resolutions to before this pull request can advance:

- [x] https://github.com/microsoft/rushstack/issues/519
- [x] https://github.com/microsoft/rushstack/issues/1029
- [x] https://github.com/microsoft/rushstack/issues/1050
- [ ] https://github.com/microsoft/rushstack/issues/1308
- [ ] https://github.com/microsoft/rushstack/issues/1830
- [ ] https://github.com/microsoft/rushstack/issues/1921
- [ ] https://github.com/microsoft/rushstack/issues/1957
- [ ] https://github.com/microsoft/rushstack/issues/1958
- [x] https://github.com/microsoft/rushstack/issues/2140

Closes #135.

PR #1372, opened 2023-03-09
Generate API documentation

### LINKED ISSUE #1371: Documentation is not generated
The [API documentation](https://github.com/Siteimprove/alfa/blob/main/docs/api/index.md#packages) has not been generated since #290 🙈 
Need to figure out the missing step and add it to the build flow.

Jym77:
Closes #1371.

Since #290, we have ways to extract the API autombatically, but forgot to generate the actual documentation (there was a script, that was never called 🙈).
This add a documentation config, and instruction to generate it in release cycle.
Additionally, this cleans up a bit the `scripts` folder.


Jym77:
!pr dedupe

rcj-siteimprove (approved):

PR #1375, opened 2023-03-14
Use `hasComputedStyle` and `hasCascadedStyle` in more places

rcj-siteimprove:
Spreading the use of `hasComputedStyle` and `hasCascadedStyle`. The following were modified to use one or both of them:
- sia-r71
- sia-r72
- sia-r74
- sia-r80
- sia-r85
- is-inert.ts
- is-transparent.ts


packages/alfa-rules/src/sia-r71/rule.ts
```diff
@@ -13,13 +13,19 @@ import { Scope } from "../tags";
...
 export default Rule.Atomic.of<Page, Element>({
   uri: "https://alfa.siteimprove.com/rules/sia-r71",
   requirements: [Criterion.of("1.4.8")],
   tags: [Scope.Component],
   evaluate({ device, document }) {
+    const isNotJustified = hasComputedStyle(
+      "text-align",
+      (align) => align.value !== "justify",
+      device
+    );
```
Jym77, line 27: That can live out of the builder, so it is instancied once when the file is loaded, and not every time the rule is evaluated.

packages/alfa-rules/src/sia-r72/rule.ts
```diff
@@ -12,12 +12,18 @@ import { Scope } from "../tags";
...
 
 export default Rule.Atomic.of<Page, Element>({
   uri: "https://alfa.siteimprove.com/rules/sia-r72",
   tags: [Scope.Component],
   evaluate({ device, document }) {
+    const isNotUpperCased = hasComputedStyle(
+      "text-transform",
+      (transform) => transform.value !== "uppercase",
+      device
+    );
```
Jym77, line 25: Can be moved out.

packages/alfa-rules/src/sia-r74/rule.ts
```diff
@@ -26,19 +26,14 @@ export default Rule.Atomic.of<Page, Element>({
...
-                  }
-                }),
+            hasCascadedStyle(
+              "font-size",
+              (fontSize) =>
+                (fontSize.type === "length" ||
+                  fontSize.type === "percentage") &&
+                fontSize.value > 0,
+              device
+            ),
```
Jym77, line 36: This used to return `true` if it was neither length nor percentage, bit now returns `false` 🤔 
(and some test might be missing if we're not catching this 😖 )

packages/alfa-rules/src/sia-r80/rule.ts
```diff
@@ -14,13 +14,20 @@ import { Scope } from "../tags";
...
   requirements: [Criterion.of("1.4.8")],
   tags: [Scope.Component],
   evaluate({ device, document }) {
+    const hasRelativeUnit = hasCascadedStyle(
+      "line-height",
+      (lineHeight) =>
+        lineHeight.type !== "length" || Unit.Length.isRelative(lineHeight.unit),
+      device
+    );
+
```
Jym77, line 30: Can be moved out.

packages/alfa-rules/src/sia-r85/rule.ts
```diff
@@ -12,12 +12,18 @@ import { Scope } from "../tags";
...
 export default Rule.Atomic.of<Page, Element>({
   uri: "https://alfa.siteimprove.com/rules/sia-r85",
   tags: [Scope.Component],
   evaluate({ device, document }) {
+    const isNotItalic = hasComputedStyle(
+      "font-style",
+      (style) => style.value !== "italic",
+      device
+    );
+
```
Jym77, line 26: Can be moved out.

packages/alfa-style/src/node/predicate/is-transparent.ts
```diff
@@ -26,9 +26,12 @@ export function isTransparent(
...
           isElement(node) &&
-          Style.from(node, device, context)
-            .computed("opacity")
-            .some((opacity) => opacity.value === 0)
+          hasComputedStyle(
+            "opacity",
+            (opacity) => opacity.value === 0,
+            device,
+            context
+          )(node)
```
Jym77, line 34: Can be grouped with `and`

packages/alfa-style/src/node/predicate/is-transparent.ts
```diff
@@ -37,11 +40,12 @@ export function isTransparent(
...
-                (color) => color.type === "color" && Color.isTransparent(color)
-              )
+            hasComputedStyle(
+              "color",
+              Color.isTransparent,
+              device,
+              context
+            )(parent)
           ) {
             return true;
```
Jym77, line 50: The tests on `parent` can be grouped witha `and`.
The loop can likely be replaced by a `node.parent(Node.flatTree).some(…)`, probably extracting the `isText`.
I suspect that the two bits can then grouped together, maybe extracting a `hasTransparentParent` helper and then something like `or( and(isElement, hasComputedStyle), and(isText, hasTransparentParent))` 🤔

Jym77 (changes requested):
There seems to be one change of behaviour in R74 🤔 
A bit of polishing on the rest.

packages/alfa-rules/src/sia-r71/rule.ts
rcj-siteimprove: It depends on `device` which is only available inside `evaluate`. We could have it as a function on the file level, but that would still initialize a new predicate on each invocation I think 🤔

Jym77: Ah, yes. We do tend to have these out of the `evaluate`, though (as functions `Device => Predicate<Element>`).

packages/alfa-rules/src/sia-r74/rule.ts
rcj-siteimprove: Right, good catch, what it wanted was probably something like `foo ? fontSize.value > 0 : true`. I'll try and add a failing test and fix it 😄

packages/alfa-style/src/node/predicate/is-transparent.ts
```diff
@@ -23,33 +49,11 @@ export function isTransparent(
...
-
-          return isTransparent(parent);
-        }
-
-        return false;
-      });
+      .get(node, () =>
+        or(
+          and(isElement, isNotOpaque(device, context)),
+          and(isText, hasTransparentParent(device, context))
```
rcj-siteimprove, line 55: Not completely sure about this. There's a theoretical change in behavior because before, if `isText(node)`  was false, it would still recurse and call `isTransparent` on the parent. That won't happen anymore. Not sure if that's okay if `node` is always guarantied to be `Element` or `Text` 🤔

Jym77: Yes, we need to keep the recursion 🤔 
```html
<div style="opacity: 0">
  <span>Hello world</span>
</div>
```
The `<span>` is transparent, because of the (non-)opacity of the `<div>`. But it seems to fail the check.

Seemsl ike we can:
* rename `hasTransparentParent` => `hasTransparentTextColor`, removing the `isTransparent` call (and typing the input as `Text`). (actually, iirc, `hasComputedStyle` does goes to parent automatically when called on text node, so we should be able to simplify that a bit)
* add `hasTransparentParent` that just calls `parent.some(isTransparent)` (might end up inlinable with a simple arrow function, which I tend to like to keep the recursion inside a single function rather than jumping between two).
* call `hasTransparentParent` as the last item in the `or` here (only recurse in last recourse).

packages/alfa-style/src/node/predicate/is-transparent.ts
```diff
@@ -23,33 +43,12 @@ export function isTransparent(
...
-          return isTransparent(parent);
-        }
-
-        return false;
-      });
+      .get(node, () =>
+        or(
+          and(isElement, isNotOpaque(device, context)),
+          and(isText, hasTransparentTextColor(device, context)),
+          (node: Node<string>) => node.parent(Node.flatTree).some(isTransparent)
```
Jym77, line 50: **Question:** why is the type annotation needed?
```suggestion
          (node) => node.parent(Node.flatTree).some(isTransparent)
```

Jym77 (changes requested):
Mostly small polish to keep in line with other files.

packages/alfa-style/src/node/predicate/is-transparent.ts
rcj-siteimprove: The type will be `unknown` otherwise, I'm actually not sure why

Jym77: Ah, yes, the predicate could be anything and `Cache` only expect a fully evaluated `boolean`…
Does it work with `test(or(…), node)`? Then the typing of `test` should force the predicate to be about `Node` 🤔 

(in any case, `Node` should be enough, `Node<string>` is the default.

rcj-siteimprove: `test` didn't make a difference, but added it anyway to be consistent and used just `Node`

Jym77 (approved):

Issue #1037, opened 2022-01-21
Investigate adding `D extends Diagnostic` type parameter to `alfa-act`


Jym77:
Most classes in `alfa-act` wrap a `Diagnostic` (`Outcomes`, `Question`), or will produce one (`Rule`). We should add a type parameter `D extends Diagnostic` to these and make sure it is correctly propagated.

This would ensure better type safety for consumers. Among other, if we change the specific diagnostic produced by a rule, a consumer with old code of an `Outcome<… D …>` (produced by a `Rule<… D …>`) would cause type mismatch when trying to consume an `Outcome<… D' …>` and avoid runtime error by trying to access fields that have changed.

We might even be able to have several diagnostic parameters for rules and separate based on which outcome is produced.

PR #1376, opened 2023-03-15
Improve `Flattened` type

Jym77:
Part of #1037

* Add an `alfa-act/Question.Metadata` type that captures the expected shape of Metadata. Given that the shape is already enforced in the `Oracle` type, this ensures it at rule building time.
* Fix the inference of `Input`, `Target`, and `Subject` types in the `Flattened` export.
* Update TypeScript. We're currently locked out of 4.9 because of https://github.com/microsoft/TypeScript/issues/53234 😞 
* Add some `type` annotations in some `import` statements.
* Default empty question metadata to `{}` rather than `never`, this avoids `never` creeping in some weird places.
* Rules asking no questions now accept anything as oracle… since they'll never call it…


packages/alfa-act/src/audit.ts
rcj-siteimprove, line 1: What's the benefit of the type-imports? It looks like the transpiler should be clever enough to elide them without the type keyword.

Jym77: Yes, it is.
I find it valuable for devs to untangled stuff that look like circular imports but where one is actually a type import.
In other words, it's a comment whose accuracy is checked at build time 😄

rcj-siteimprove: Alright, that makes sense, just wanted to clarify the intent 😊

rcj-siteimprove (approved):

PR #1395, opened 2023-04-13
Add `Result#getUnsafe`

rcj-siteimprove:
Resolves one of the tasks on #1037 

Removed `Err#get` and `Ok#getErr` and replaced them by `Result#getUnsafe` and `Result#getErrUnsafe`. Then fixed all resulting compile errors.

It was necessary to introduce a breaking change to `Request#from`, `Response#from` and `Page#from` as they were unsafely getting the result of parsing a URL which might fail.


packages/alfa-http/test/request.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+  t.deepEqual(
+    Request.from({
+      url: "foo",
+      body: "",
+      headers: [],
+      method: "GET",
+    }).toJSON(),
+    {
+      type: "err",
+      error: "Invalid URL",
```
rcj-siteimprove, line 15: Annoyingly, the behavior is different between Node 14 and 16. In 14 the error message is "Invalid URL: foo" and in 16 it's just "Invalid URL". I'll need to make the tests able to cope with this.

.changeset/sour-dingos-rush.md
```diff
@@ -0,0 +1,18 @@
+---
```
Jym77, line 1: **Suggestion:**
I think we should have several changesets here
* One for `alfa-result`, which is breaking (should be major, down to minor because of pre-release). This should focus on the API change and advice (as quick migration fix) to replace `foo.get()` by `foo.getUnsafe()`.
* One for `alfa-web`. This is, technically, a different breaking change where the `from` functions now return `Result` instead of the object. This would advice to add `.getUnsafe()` call as a quick migration fix.
* Maybe one for all the rest. I'm not fully convinced it is needed since these do not have any API or functional change, so there is nothing that really need to be conveyed to consumers.

Splitting the changesets means that each packages CHANGELOG will only contain its own changes. That is, consumers of `alfa-result` don't care about the change in `alfa-web`, so the `alfa-result` changelog shouldn't list them.

packages/alfa-result/src/result.ts
```diff
@@ -59,8 +59,14 @@ export interface Result<T, E = T>
...
-  get(message?: string): T;
-  getErr(message?: string): E;
+  /**
+   * This may throw an exception. Use only when you can provide an external
+   * guarantee that it won't be used on Err. E.g. in tests, or when some
+   * condition exists that TypeScript cannot see (document it!)
+   *
+   * @internal
+   */
+  getUnsafe(message?: string): T;
```
Jym77, line 69: **Suggestion:** We should similarly have a `getErrUnsafe` to avoid type assertions and provide a fallback of the previous behavior.

packages/alfa-act/src/question.ts
```diff
@@ -188,9 +188,11 @@ export class Question<
       condition = conditionOrPredicateOrAnswer.isOk();
 
       if (condition) {
-        answer = conditionOrPredicateOrAnswer.get();
+        // the check ensures that the result is Ok
+        answer = conditionOrPredicateOrAnswer.getUnsafe();
       } else {
-        const error = conditionOrPredicateOrAnswer.getErr();
+        // since the condition is false, we know the result is an Err
+        const error = (conditionOrPredicateOrAnswer as Err<unknown>).getErr();
```
Jym77, line 195:
```suggestion
        const error = conditionOrPredicateOrAnswer.getErrUnsafe();
```

packages/alfa-css/src/syntax/declaration.ts
```diff
@@ -195,11 +196,13 @@ export namespace Declaration {
           const result = consume(Slice.of(value));
 
           if (result.isOk()) {
-            declarations.push(result.get()[1]);
+            // the check ensures the result is Ok
+            declarations.push(result.getUnsafe()[1]);
```
Jym77, line 200: **Thought:** This is weird 🤔 `.isOk` should correctly narrow the type.

packages/alfa-css/src/syntax/function.ts
```diff
@@ -144,7 +144,8 @@ export namespace Function {
           return result;
         }
 
-        const [remainder, value] = result.get();
+        // the previous check ensures the result is Ok
```
Jym77, line 147: **Thought:** Comparing with what was done for `Option`, it might be that changing the test to `!result.isOk()` would do the correct narrowing.

packages/alfa-css/src/syntax/lexer.ts
```diff
@@ -229,7 +229,8 @@ const consumeEscapedCodePoint: Parser.Infallible<
         break;
       }
 
-      const [, byte] = result.get();
+      // the previous check ensure the result is Ok
+      const [, byte] = result.getUnsafe();
```
Jym77, line 233: **Tought:** Same here, replacing `foo.isErr` by `!foo.isOk` should narrow type.

packages/alfa-url/src/url.ts
```diff
@@ -344,7 +346,8 @@ export namespace URL {
         return result;
       }
 
-      base = result.get();
+      // The previous check ensures that the result is Ok
+      base = result.getUnsafe();
```
Jym77, line 350: **Thought:** Can we change to `!isOk`?

packages/alfa-web/src/page.ts
```diff
@@ -119,12 +121,18 @@ export namespace Page {
...
-  export function from(json: JSON): Page {
-    return Page.of(
-      Request.from(json.request),
-      Response.from(json.response),
-      Document.from(json.document),
-      Device.from(json.device)
+  export function from(json: JSON): Result<Page, string> {
+    return Request.from(json.request).andThen((request) =>
+      Response.from(json.response).andThen((response) =>
+        Result.of(
```
Jym77, line 127:
```suggestion
      Response.from(json.response).map((response) =>
```
Wouldn't that work also?

packages/alfa-web/test/page.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+        },
+        scripting: {
+          enabled: true,
+        },
+        preferences: [],
+      },
+      document: {
+        type: "document",
+        children: [],
+        style: [],
```
Jym77, line 40: **Thought:** you may want to shorten the test by factoring in the device and document JSON, you can probably also use `Device.standard().toJSON()` since it doesn't really matter.

packages/alfa-xpath/src/syntax/lexer.ts
```diff
@@ -253,7 +254,8 @@ const lexCommentContents: Parser<Slice<number>, string> = (input) => {
           input = input.slice(1);
           value += ":";
 
-          const [remainder, result] = lexCommentContents(input).get();
+          // this function never returns Err
+          const [remainder, result] = lexCommentContents(input).getUnsafe();
```
Jym77, line 258: **Thought:** Should that be an infaillible parser, then?

Jym77 (changes requested):

packages/alfa-css/src/syntax/declaration.ts
rcj-siteimprove: Indeed it did, must have been a mouse slip 😅

packages/alfa-result/src/result.ts
rcj-siteimprove: Make sense, added.

packages/alfa-url/src/url.ts
```diff
@@ -46,14 +46,16 @@ export class URL implements Equatable, Hashable, Serializable<URL.JSON> {
...
+    return URL.parse("https://example.com").getUnsafe();
   }
 
   /**
    * {@link https://tools.ietf.org/html/rfc6694#section-3}
    */
   public static blank(): URL {
-    return URL.parse("about:blank").get();
+    // this is a valid URL
+    return URL.parse("about:blank").getUnsafe();
```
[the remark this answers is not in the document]
rcj-siteimprove: Yes, I think that would be nice. Added.

packages/alfa-xpath/src/syntax/lexer.ts
rcj-siteimprove: I ended up making it infallible and adding a `Parser.Infallible#toParser` function.

packages/alfa-rules/src/sia-r6/rule.ts
```diff
@@ -35,9 +35,10 @@ export default Rule.Atomic.of<Page, Element>({
 
       expectations(target) {
         // The last filter in applicability ensures that lang exists
+        // The applicability ensures it can be parsed
```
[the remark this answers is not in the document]
rcj-siteimprove: That makes sense, I'll add a follow up task on it

.changeset/rotten-walls-attack.md
```diff
@@ -0,0 +1,5 @@
+---
+"@siteimprove/alfa-web": minor
+---
+
+Changed `Request#from`, `Response#from` and `Page#from` to return `Result<...>` to reflect the fact that the function might fail on invalid input. As a guick migration add a `.getUnsafe()` call to the returned result which will retain the original behavior where an exception might be thrown.
```
Jym77, line 5:
```suggestion
Changed `Request#from`, `Response#from` and `Page#from` to return `Result<...>` to reflect the fact that the function might fail on invalid input. As a quick migration add a `.getUnsafe()` call to the returned result which will retain the original behavior where an exception might be thrown.
```

packages/alfa-style/src/property/box-shadow.ts
```diff
@@ -10,7 +10,7 @@ import {
   Token,
 } from "@siteimprove/alfa-css";
 import { Parser } from "@siteimprove/alfa-parser";
-import { Err, Result } from "@siteimprove/alfa-result";
+import { Err, Ok, Result } from "@siteimprove/alfa-result";
```
Jym77, line 13:
```suggestion
import { Err, Result } from "@siteimprove/alfa-result";
```

Jym77 (dismissed):

Jym77 (approved):

PR #1397, opened 2023-04-18
Style refactor: streamlining type parameters and extracting variables mangling

Jym77:
* Rename type parameters to be more explicit.
* Extract all the variables mangling to a separate file.

This does not require changeset since the API change is just the name of type parameters.


rcj-siteimprove (approved):
Looks good

PR #1400, opened 2023-04-20
Export `Diagnostic` subclasses from `alfa-rules`

rcj-siteimprove:
Resolves one of the tasks on #1037 

This change exposes diagnostic subclasses like e.g. `WithNextHeading` on the `Diagnostic` class from `alfa-rules`. This means they can now be referenced like in the following:
```
import { Diagnostic } from "@siteimprove/alfa-rules";

function foo(d: Diagnostic.Languages) {
  // ...
}
```


packages/alfa-rules/src/common/act/diagnostic.ts
```diff
@@ -1,73 +1,67 @@
...
+  export const { MatchingClasses } = matchingClasses;
+  export const { WithDeclaration } = withDeclaration;
+  export const { WithNextHeading } = withNextHeading;
+  export const { ClippingAncestors } = clippingAncestors;
+  export const { Contrast, TextSpacing, WithBadElements, WithRole } =
+    diagnostic;
 
+  export const { isDistinguishingStyles: isDistinguishingStylesDeprecated } =
+    DistinguishingStyles;
+  export const { isLanguages } = Languages;
```
Jym77, line 45: **Question:** Can't we group the exports?
`export const { isLanguages, Languages } = languages`
and so on.

Jym77 (changes requested):
This requires a changeset.

Jym77 (approved):

PR #1402, opened 2023-04-21
Add `Maybe#toOption` and move out of `Option` namespace

rcj-siteimprove:
Resolves part of #1037 

Added `Maybe#toOption` and moved type out of `Option` namespace. This is a step in the direction of cleaning how we use the `Maybe` type in the rules.


Jym77:
!pr extract

Jym77 (dismissed):

Jym77 (approved):

PR #1403, opened 2023-04-21
SIA-R6: Make values computed in applicability reusable

### LINKED ISSUE #1399: Extract and reuse language attribute parsing in sia-r6
**Thought (non-blocking):** Wow! Looks like we should save that parsed value when computing it rather than compute it twice and heavily `getUnsafe` 😅

_Originally posted by @Jym77 in https://github.com/Siteimprove/alfa/pull/1395#discussion_r1166509369_

rcj-siteimprove:
Resolves #1399 

This caches one computation and gets rid of one `.getUnsafe` call

### DIFF
--- a/packages/alfa-rules/src/sia-r6/rule.ts
+++ b/packages/alfa-rules/src/sia-r6/rule.ts
@@ -9,40 +9,52 @@ import { Page } from "@siteimprove/alfa-web";
 
 import { expectation } from "../common/act/expectation";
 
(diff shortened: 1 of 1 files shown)


Jym77 (approved):

Jym77 (changes requested):
We actually only have one target and don't need a map.

Jym77 (approved):

PR #1405, opened 2023-05-01
Typescript 5

Jym77:
Part of #1037.

Update to TS 5. There is still a type that cannot really be handled. This is partly due to some design choices in TS. So we have to help it and assert the correct type.


Jym77:
!pr extract

rcj-siteimprove (approved):
Looks good

PR #1406, opened 2023-05-01
Abstract Math calculation in style properties

Jym77:
Part of #1037 and #1202.

Create an abstraction for length, length-percentage, and number that can also be the result of calculations. The abstraction is easy to use in individual properties since it appears as a CSS value with a handy resolver to absolutize it.

### DIFF
--- a/packages/alfa-css/src/value/math-expression/math.ts
+++ b/packages/alfa-css/src/value/math-expression/math.ts
@@ -2,7 +2,7 @@ import { Hash } from "@siteimprove/alfa-hash";
 import { Option, None } from "@siteimprove/alfa-option";
 import { Parser } from "@siteimprove/alfa-parser";
 import { Slice } from "@siteimprove/alfa-slice";
-import { Result } from "@siteimprove/alfa-result";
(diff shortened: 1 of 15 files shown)


packages/alfa-css/src/value/math-expression/math.ts
```diff
@@ -113,34 +112,61 @@ export class Math<
...
-      ? expression.toNumber()
-      : None;
+    try {
+      const expression = this._expression.reduce({
+        length: () => {
+          throw new Error(`Missing length resolver for ${this}`);
+        },
+        percentage: () => {
+          throw new Error(`Missing percentage resolver for ${this}`);
+        },
```
Jym77, line 141: This is not super clean. It might be better to push the optionality of the resolver further down and return `Result`.
OTOH, `Expression#reduce` should just expect a correct reducer and do its magic…
Anyway, this is hopefully good enough for now and we can revisit it later if neeed.

Jym77:
!pr extract

rcj-siteimprove (approved):
Looks good 😃

PR #1411, opened 2023-05-03
Accept calculations in more style properties

Jym77:
Part of #1037 and #1202.

This proves out to not be a very scalable solution, especially since


Jym77:
!pr extract

rcj-siteimprove (approved):
🎉

PR #1412, opened 2023-05-04
`alfa-css`: cleanup an reorganisation

Jym77:
Reorganise a bit the `alfa-css` package, essentially considering each directory as its own package with exxports through an `index.ts"` file.

The intention is that `Numeric` will be duplicated as a `Value`, including the calculations.
This will probably require `unit` to be lifted one level in the directory hierarchy. This will probably require `Value` to be also duplicated, with the "true" `Value` having some sort of `resolve` endpoint to resolve any calculation left inside the value.


packages/alfa-css/src/calculation/math-expression/expression.ts
```diff
@@ -4,7 +4,7 @@ import { Serializable } from "@siteimprove/alfa-json";
 import { None, Option } from "@siteimprove/alfa-option";
 
 import { Unit } from "../unit/unit";
-import { Length, Number, Numeric, Percentage } from "../numeric";
+import { Length, Number, Numeric, Percentage } from "../numeric/index";
```
rcj-siteimprove, line 7: **question (non-blocking):** I think `index` can still be omitted, is it added on purpose?

rcj-siteimprove (dismissed):
Nice cleanup, just found one typo and a question

packages/alfa-css/src/calculation/math-expression/expression.ts
Jym77: 🙈 IDE messing up when moving and renaming files.

rcj-siteimprove (approved):

PR #1413, opened 2023-05-04
Query API for `alfa-dom`

rcj-siteimprove:
Part of #1037 

First attempt at creating a query API for querying the DOM. For now, two common queries has been extracted and put into the new `Query` namespace in `alfa-dom` and the results are now cached which should give a slight performance improvement.


packages/alfa-dom/src/node/document.ts
```diff
@@ -38,22 +41,13 @@ export class Document extends Node<"document"> {
...
     options: Node.Traversal = Node.Traversal.empty
   ): Sequence<Element> {
-    if (this._elementDescendants[options.value] === undefined) {
-      this._elementDescendants[options.value] = this.descendants(
-        options
-      ).filter(Element.isElement);
-    }
-
-    return this._elementDescendants[options.value];
+    return getElementDescendants(this, options);
```
rcj-siteimprove, line 50: **note:** Not sure if we should replace all usages of `elementDescendants` with `getElementDescendant` instead of having this indirection. If so, maybe that can be done in a separate PR to limit the blast radius of this one. OTOH it's also convenient to have this instance function so that we can do `foo.elementDescendants()` 🤔

packages/alfa-aria/src/node.ts
```diff
@@ -207,17 +208,9 @@ export namespace Node {
...
-    // The collected references are added to the map in reverse order to ensure
-    // that the first occurrence of a given ID is what ends up in the map in
-    // event of duplicates.
-    const ids = Map.from(
-      elements
-        .collect((element) => element.id.map((id) => [id, element] as const))
-        .reverse()
-    );
+
+    const elements = getElementDescendants(root);
```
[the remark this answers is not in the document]
Jym77: I don't think we need a special function (especially since it duplicate the cache, …) We can just do `getElementDescendants(element).append(element)` if need be.

packages/alfa-dom/src/node/document.ts
```diff
@@ -1,10 +1,13 @@
 import { None, Option } from "@siteimprove/alfa-option";
-import { Sequence } from "@siteimprove/alfa-sequence/src/sequence";
 import { Trampoline } from "@siteimprove/alfa-trampoline";
 
+import { Sequence } from "@siteimprove/alfa-sequence";
```
Jym77, line 4:
```suggestion
import { Sequence } from "@siteimprove/alfa-sequence";
import { Trampoline } from "@siteimprove/alfa-trampoline";
```

packages/alfa-dom/src/node/document.ts
Jym77: I think hat ultimately we should. `foo.elementDescendants` wasn't that well thought through.
Let's keep it for another PR indeed.

packages/alfa-dom/src/node/query/element-descendants.ts
```diff
@@ -0,0 +1,23 @@
...
+ */
+export function getElementDescendants(
+  node: Node,
+  options: Node.Traversal = Node.Traversal.empty
+): Sequence<Element> {
+  const optionsMap = _elementDescendantsCache.get(node, () => []);
+  if (optionsMap[options.value] === undefined) {
+    optionsMap[options.value] = node
+      .descendants(options)
+      .filter(Element.isElement);
```
Jym77, line 19: **Issue:** Does that actually update the array inside the cache 🤔

packages/alfa-dom/src/node/query/element-id-map.ts
```diff
@@ -0,0 +1,32 @@
...
+  return buildElementIdMap(node);
+}
+
+function buildElementIdMap(node: Node): Map<string, Element> {
+  // Build a map from ID -> element to allow fast resolution of ID references.
+  // The collected references are added to the map in reverse order to ensure
+  // that the first occurrence of a given ID is what ends up in the map in
+  // event of duplicates.
+  return Map.from(
+    Query.getElementDescendants(node)
```
Jym77, line 28: **Suggestion:** We're inside the same directory and should import the function directly, rather than the namespace.

packages/alfa-dom/src/node/query/index.ts
```diff
@@ -0,0 +1,17 @@
+import * as elementDescendants from "./element-descendants";
+import * as elementIdMap from "./element-id-map";
+
+/**
+ * @public
+ */
+export namespace Query {
+  /**
+   * @public
+   */
```
Jym77, line 10:
```suggestion
```
I think they inherit the one from the namespace.

packages/alfa-dom/src/node/query/index.ts
```diff
@@ -0,0 +1,17 @@
...
+  export const getElementDescendants = elementDescendants.getElementDescendants;
+
+  /**
+   * @public
+   */
```
Jym77, line 15:
```suggestion
```

packages/alfa-dom/test/node/query/element-descendants.spec.tsx
```diff
@@ -0,0 +1,15 @@
+import { test } from "@siteimprove/alfa-test";
+import { h } from "../../../h";
+
+import { Query } from "../../../src";
+
+const { getElementDescendants } = Query;
+
+test("#getElementDescendants() returns descendants non-inclusively", (t) => {
```
Jym77, line 8: **Nitpick:** This technically doesn't test non-inclusivity since `doc` is not an `Element` and would be filtered away anyway.

packages/alfa-dom/src/node/query/element-id-map.ts
```diff
@@ -0,0 +1,32 @@
...
+import { Query } from ".";
+import { Node } from "../../node";
+import { Document } from "../document";
+import { Element } from "../element";
+
+const elementMapCache = Cache.empty<Document, Map<string, Element>>();
+
+/**
+ * @public
+ */
```
Jym77, line 13:
```suggestion
 * @remarks Since `id` are scoped to trees, and do not cross shadow or content boundaries,
 * we never need traversal options.
 */
```

Jym77 (changes requested):
Great work. Some small polish.

packages/alfa-dom/src/node/document.ts
rcj-siteimprove: Cool, I'll add it to #1037

packages/alfa-dom/test/node/query/element-descendants.spec.tsx
rcj-siteimprove: Ah, yes, good catch

packages/alfa-dom/src/node/query/element-descendants.ts
Jym77: > It seems to yes when I step through the code, I guess because it's just a reference to the physical array or something?

👍 

> Not sure however if it's the best approach. Maybe it's better to explicitly have a 4 separate caches for each of the possible enum values?

No. There are actually 8 enum values (this is a bits array, they can be combined: https://github.com/Siteimprove/alfa/blob/main/packages/alfa-dom/src/node.ts#L319). The `.value` helper is precisely here so that we abstract the flag implementation away and get a guaranteed different integer for each flag combination. See [alfa-flags](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-flags/src/flags.ts) for the dirty details.

packages/alfa-aria/src/node.ts
```diff
@@ -15,16 +15,18 @@ import * as dom from "@siteimprove/alfa-dom";
...
 
 import { Attribute } from "./attribute";
+import { Feature } from "./feature";
 import { Name } from "./name";
 import { Role } from "./role";
-import { Feature } from "./feature";
 
 import { Container, Element, Inert, Text } from ".";
 
+import { Lazy } from "@siteimprove/alfa-lazy";
```
Jym77, line 24: **chore:** Should be moved with the other imports (or removed).

packages/alfa-aria/src/node.ts
```diff
@@ -207,17 +209,13 @@ export namespace Node {
...
-    // event of duplicates.
-    const ids = Map.from(
-      elements
-        .collect((element) => element.id.map((id) => [id, element] as const))
-        .reverse()
-    );
+
+    const exclusiveDescendants = getElementDescendants(root);
+    const elements = dom.Element.isElement(root)
+      ? Sequence.of(root, Lazy.force(exclusiveDescendants))
```
Jym77, line 215: **Question:** I'm not sure why we need to `Lazy.force` it now? 
```suggestion
      ? exclusiveDescendants.prepend(root)
```

packages/alfa-dom/test/node/query/element-id-map.spec.tsx
```diff
@@ -0,0 +1,16 @@
+import { test } from "@siteimprove/alfa-test";
+import { h } from "../../../h";
+
+import { Query } from "../../../src";
+
+const { getElementIdMap } = Query;
+
+test("#getElementIdMap() returns map from ids to elements", (t) => {
```
Jym77, line 8: **Suggestion:** can we add a test to check that when `id` are duplicated we only return the first one (in tree order)?

Jym77 (changes requested):
We should also use the new cached map for `aria-labelledby` resolution at https://github.com/Siteimprove/alfa/blob/b1e42bee45a3cb5d78f50c6af509eeab29519101/packages/alfa-aria/src/name.ts#L890-L903 (and it should also simplify that code by not needing to sort since we'll poke the map in the correct order)

packages/alfa-dom/test/node/query/element-id-map.spec.tsx
rcj-siteimprove: > We should also use the new cached map for `aria-labelledby` resolution at
> 
> https://github.com/Siteimprove/alfa/blob/b1e42bee45a3cb5d78f50c6af509eeab29519101/packages/alfa-aria/src/name.ts#L890-L903
> 
> (and it should also simplify that code by not needing to sort since we'll poke the map in the correct order)

There's also a follow up task on #1037 on getting rid of all usages of `.elementDescendants()` so it would be picked up there too. Should we keep it as part of that follow up task?

rcj-siteimprove:
> We should also use the new cached map for `aria-labelledby` resolution at
> 
> https://github.com/Siteimprove/alfa/blob/b1e42bee45a3cb5d78f50c6af509eeab29519101/packages/alfa-aria/src/name.ts#L890-L903
> 
> (and it should also simplify that code by not needing to sort since we'll poke the map in the correct order)

There's also a follow up task on #1037 on getting rid of all usages of `.elementDescendants()` so it would be picked up there too. Should we keep it as part of that follow up task?

Jym77 (dismissed):

PR #1415, opened 2023-05-08
Clean up deprecated rules

Jym77:
Remove the old deprecated rules.
Replace R18, R66, R69 by their v2.
R66v2 and R69v2 are already used by Dory and the extension; R18v2 is already used by the extension.


rcj-siteimprove (approved):
Looks good

PR #1416, opened 2023-05-09
Introduce new CSS length with CSS calculations included

Jym77:
Part of #1202 

Prepare the architecture for supporting CSS calculations in all numeric types, including inside "composed" type such as RGB colors.

This set up calculated lengths as a value type in `alfa-css`, and use them in one property in `alfa-style`.
Ultimately, the goal is to use that in all places, and stop exporting `calculation/numeric/length`, then do the same for other numeric type (plus the mixed `dimension-percentage` ones. This PR only showcases the solution and is big enough already…


Jym77:
!pr extract

rcj-siteimprove (dismissed):
Great work. Just found a typo in the changeset.

rcj-siteimprove (approved):

rcj-siteimprove (approved):

PR #1425, opened 2023-05-25
Replace `.elementDescendants()` with `.getElementDescendants()`

rcj-siteimprove:
Follows up on #1413, part of #1037.

### DIFF
--- a/packages/alfa-rules/src/sia-r81/rule.ts
+++ b/packages/alfa-rules/src/sia-r81/rule.ts
@@ -1,7 +1,7 @@
 import { Diagnostic, Rule } from "@siteimprove/alfa-act";
 import { DOM, Node } from "@siteimprove/alfa-aria";
 import { Device } from "@siteimprove/alfa-device";
-import { Element, Namespace } from "@siteimprove/alfa-dom";
+import { Element, Namespace, Query } from "@siteimprove/alfa-dom";
 import { Iterable } from "@siteimprove/alfa-iterable";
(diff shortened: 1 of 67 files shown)


Jym77:
!pr extract

packages/alfa-rules/src/sia-r81/rule.ts
```diff
@@ -143,10 +143,9 @@ function linkContext(element: Element, device: Device): Set<dom.Node> {
   }
 
   for (const describedby of element.attribute("aria-describedby")) {
-    for (const reference of element
-      .root()
-      .elementDescendants()
-      .filter(hasId(equals(...describedby.tokens())))) {
+    for (const reference of getElementDescendants(element.root()).filter(
+      hasId(equals(...describedby.tokens()))
+    )) {
```
Jym77, line 148: **Suggestion:** This actually looks like it could use `getIdMap` 🤔 
It seems to be fetching all elemenjt whose `id` is one of the `describedby.tokens`.

Jym77 (changes requested):
Looks like there is an (unrelated) possible optimisation with using `getIdMap`. I guess we should as well do it now.

packages/alfa-rules/src/sia-r81/rule.ts
rcj-siteimprove: Cool, good catch. I'll look into it 😃

Jym77 (approved):

PR #1427, opened 2023-06-05
Add `alfa-rectangle` package

rcj-siteimprove:
This PR adds a new package `alfa-rectangle` with a `Rectangle` class for working with rectangles.

In addition I have added a new, optional, `Rectangle` property to the `Element` class. It's not being passed in or used for anything yet.


rcj-siteimprove:
!pr extract

packages/alfa-dom/src/node/element.ts
```diff
@@ -32,15 +33,17 @@ export class Element<N extends string = string>
     name: N,
     attributes: Iterable<Attribute> = [],
     children: Iterable<Node> = [],
-    style: Option<Block> = None
+    style: Option<Block> = None,
+    rectangle: Option<Rectangle> = None
```
Jym77, line 37: **Note:** we start having a lot of optional arguments on that function and it might be worth adding (or even replacing) an API with an `options` (or `extra`) object to selectively pass the ones we want.
Probably not for now, anyway.

packages/alfa-dom/src/node/element.ts
```diff
@@ -49,6 +52,7 @@ export class Element<N extends string = string>
   private readonly _name: N;
   private readonly _attributes: Map<string, Attribute>;
   private readonly _style: Option<Block>;
+  private readonly _rectangle: Option<Rectangle>;
```
Jym77, line 55: **Suggestion:** I think this should be called "bounding box" or something like that (i.e. a semantic name, rather than a technical one).
OTOH, this is not exactly any of the CSS boxes 🙈 https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect
Maybe `box` is generic enough?

packages/alfa-dom/package.json
```diff
@@ -49,6 +49,7 @@
     "@siteimprove/alfa-media": "workspace:^0.63.3",
     "@siteimprove/alfa-option": "workspace:^0.63.3",
     "@siteimprove/alfa-predicate": "workspace:^0.63.3",
+    "@siteimprove/alfa-rectangle": "workspace:^0.63.3",
```
Jym77, line 52: **Chore:** it also needs to be added to `tsconfig.json`.

packages/alfa-rectangle/src/rectangle.ts
```diff
@@ -0,0 +1,157 @@
...
+    return this._x;
+  }
+
+  get right() {
+    return this._x + this._width;
+  }
+
+  get bottom() {
+    return this._y + this._height;
+  }
```
Jym77, line 58: **Issue:** these should handle negative height/width. And we should probably provide x/y getters also.
See the [DOMRect API](https://developer.mozilla.org/en-US/docs/Web/API/DOMRect).

packages/alfa-rectangle/src/rectangle.ts
```diff
@@ -0,0 +1,157 @@
...
+  private readonly _height: number;
+
+  private constructor(x: number, y: number, width: number, height: number) {
+    this._x = x;
+    this._y = y;
+    this._width = width;
+    this._height = height;
+  }
+
+  get width() {
```
Jym77, line 36: **Chore:** we usually have both explicit `public` modifier, and explicit return type.
```suggestion
  public get width(): number {
```

packages/alfa-rectangle/src/rectangle.ts
```diff
@@ -0,0 +1,157 @@
...
+        value.height === this.height)
+    );
+  }
+
+  hash(hash: Hash): void {
+    hash
+      .writeFloat32(this._x)
+      .writeFloat32(this._y)
+      .writeFloat32(this._width)
+      .writeFloat32(this._height);
```
Jym77, line 107: **Suggestion:** we can just use the generic `writeNumber`, and let `Hash` handle optimisations if needed.
```suggestion
      .writeNumber(this._x)
      .writeNumber(this._y)
      .writeNumber(this._width)
      .writeNumber(this._height);
```

packages/alfa-rectangle/src/rectangle.ts
```diff
@@ -0,0 +1,157 @@
...
+
+    for (const rectangle of rectangles) {
+      l = min(l, rectangle.left);
+      t = min(t, rectangle.top);
+      r = max(r, rectangle.right);
+      b = max(b, rectangle.bottom);
+    }
+
+    return Rectangle.of(l, t, r - l, b - t);
+  }
```
Jym77, line 156: **Suggestion:** we could have a `#union(other)` method on the class, and this would just be a `reduce` on the list.
This also suggest we may want the (reverse) infinite rectangle as a static instance (like `.empty()` on other classes). Maybe also the same for `intersection` (and the infinite rectangle). Intersection is a likely use case for layout (is text on top of image?)

packages/alfa-rectangle/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+    "$schema": "http://json.schemastore.org/tsconfig",
+    "extends": "../tsconfig.json",
+    "files": [
+        "src/index.ts",
+        "src/rectangle.ts",
+        "test/rectangle.spec.ts"
+    ]
+  }
+
```
Jym77, line 10: **Chore:** we need to add ˛references to the other packages used by this one.

packages/alfa-rectangle/src/rectangle.ts
```diff
@@ -0,0 +1,157 @@
...
+
+  hash(hash: Hash): void {
+    hash
+      .writeFloat32(this._x)
+      .writeFloat32(this._y)
+      .writeFloat32(this._width)
+      .writeFloat32(this._height);
+  }
+
+  toJSON(): Rectangle.JSON {
```
Jym77, line 110: **Note:** DOM `Node` are also serialisable to SARIF and EARL. OTOH, it is not clear we'll need to add the box to these serialisations, since it doesn't add that much (semantics) to the result.

Jym77 (changes requested):

Jym77:
!pr depgraph

rcj-siteimprove:
In the latest push
- I changed the implementation of `.equals` so that different representations of the same rectangles are equal (different corners chosen as the origin)
- I added `empty` and `full` rectangles
- I had to take infinities into account in the `top`, `left`, `right` and `bottom` properties. Not totally sure about this.
- Finished `#union` and `#intersection`.
- Added a bunch more tests

rcj-siteimprove:
!pr extract

packages/alfa-rectangle/src/rectangle.ts
```diff
@@ -86,28 +144,76 @@ export class Rectangle
...
-  equals(value: unknown): value is this;
-  equals(value: unknown): boolean {
+  public union(other: Rectangle): Rectangle {
+    if (this.contains(other)) {
+      return this;
+    }
+
+    if (other.contains(this)) {
+      return other;
+    }
```
Jym77, line 154: **Question:** Are these early escapes really needed? It seems that the computations below should give the same result anyway, and these don't save that much time since `.contains` itself makes more or less the same comparisons.

packages/alfa-rectangle/src/rectangle.ts
```diff
@@ -86,28 +144,76 @@ export class Rectangle
...
+      return Rectangle.empty();
+    }
+
+    if (this.contains(other)) {
+      return other;
+    }
+
+    if (other.contains(this)) {
+      return this;
+    }
```
Jym77, line 180: **Question:** are these early escape really needed?

Jym77 (changes requested):
Looks good. Some minor clean up.

packages/alfa-rectangle/src/rectangle.ts
rcj-siteimprove: Right. My reason for doing it was that it reuses the existing object rather than allocating a new one which would have to be garbage collected etc. I don't have a strong opinion though on whether it's worth the extra lines of code and potential confusion.

packages/alfa-rectangle/src/rectangle.ts
rcj-siteimprove: I think I had to add the first check due to how the intersection is calculated below which for disjoint rectangles would actually otherwise return a rectangle sitting between the disjoint rectangles. That might on the other hand also be a sign that it's not the best way to calculate it. I'm not sure.

Jym77 (approved):

PR #1440, opened 2023-06-22
Element accept optional rectangle

rcj-siteimprove:
Resolves https://siteimprove-wgs.atlassian.net/browse/II-6415

All ways of constructing an `Element` now accepts an optional `box` parameter corresponding to the bounding client rect.

In particular support is added to the jsx/tsx syntax so that the box can be specified as a special attribute for testing purposes like so
```tsx
const element = <div box={{ x: 12, y: 7, width: 3, height: 9 }}></div>;
````

In addition, constructing the element from a native DOM element, when Alfa is running in a browser, will also get the box by calling `getBoundingClientRect`.

### CODE REVIEW

File: docs/review/api/alfa-style.api.md
Comment (rcj-siteimprove) [inline docs/review/api/alfa-style.api.md:1]: **note:** this was updated after running `yarn extract`, not sure why

**Comment by rcj-siteimprove** [comment]
!pr extract

File: docs/review/api/alfa-style.api.md
Comment (rcj-siteimprove) [reply docs/review/api/alfa-style.api.md:1]: It seems to be failing in CI, running another extract with pr command to see if that helps

File: .changeset/grumpy-hounds-yawn.md [hunk 1]
```diff
@@ -0,0 +1,7 @@
+---
+"@siteimprove/alfa-dom": minor
+---
+
+**Added:** Support for optional `box` everwhere when constructing an `Element`
+
+Also renamed public property `rectangle` to `box` which was overlooked in the last version
```
Jym77, line 7: This is now breaking since we released the previous version 🙈

packages/alfa-dom/src/jsx.ts
```diff
@@ -16,6 +17,7 @@ export function jsx<N extends string = string>(
 ): Element<N> {
   const attributes: Record<string, string | boolean> = {};
   const style: Record<string, string> = {};
+  let box: Rectangle | undefined = undefined;
```
Jym77, line 20:
```suggestion
  let box: Rectangle;
```
I think this is actually enough 🤔

packages/alfa-dom/test/jsx.spec.tsx
```diff
@@ -0,0 +1,19 @@
...
+    y: 7,
+    width: 3,
+    height: 9,
+  });
+});
+
+test("Element without box attribute has box property equal to `None`", (t) => {
+  const element = <div></div>;
+
+  t(element.box.isNone);
```
Jym77, line 18:
```suggestion
  t.deepEqual(element.box.isNone(), true);
```
I think the current test is just making a truthy test on the function, which is passing because functions are not falsy.

Jym77 (changes requested):
Small polish.

packages/alfa-dom/test/jsx.spec.tsx
rcj-siteimprove: Good catch 🙈

packages/alfa-dom/src/jsx.ts
rcj-siteimprove: It is actually complaining because it's not assigned to in all code paths

Jym77 (approved):

PR #1441, opened 2023-06-26
Removed `alfa-highlight`

Jym77:
Since #764, `alfa-highlight` was just a wrapper for `chalk` which should be used directly instead.


Jym77:
!pr extract

rcj-siteimprove (approved):

PR #1450, opened 2023-07-12
Use layout to improve visibility checks

Jym77:

PR #1451, opened 2023-07-12
Use layout to improve SIA-R83

### LINKED ISSUE #974: SIA-R83 could ignore clipping on elements that can be determined as distant containers
In #932 we had R83 ignore clipping done by `<body>` under the assumption that this is often a distant container with enough room inside to let the text grow and that is was often flagging small text incorrectly (e.g. flagging a breadcrumb).

We do regularly encounter similar cases of R83 flagging clipping caused by a distant container where the actual text is small and has room to grow inside the container up to several times its original size. While we can't detect these without layout information, we can try and get heuristic of "this is a distant container".

- Elements with a role of `main` can likely be assumed as being distant containers.

---
- [ ] Follow-up to DEV-12495

Jym77:
Resolves #974

### DIFF
--- a/packages/alfa-rules/test/sia-r83/rule.spec.tsx
+++ b/packages/alfa-rules/test/sia-r83/rule.spec.tsx
@@ -459,7 +459,6 @@ test(`evaluate() passes a text node with fixed height and another property
   ]);
 });
 
-/* here */
 test(`evaluates() passes a text node horizontally overflowing its small
       parent and not clipped by its wide grand-parent`, async (t) => {
   const target = h.text("Hello world");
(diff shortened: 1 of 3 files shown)


Jym77:
Note: will wait for #1450 to be done, so that we can reuse `hasBox` here instead.

packages/alfa-rules/test/sia-r83/rule.spec.tsx
```diff
@@ -459,6 +459,64 @@ test(`evaluate() passes a text node with fixed height and antoher property
   ]);
 });
 
+/* here */
```
rcj-siteimprove, line 462: **question:** are these comments some remnants that should be removed? (also on line 491)

rcj-siteimprove (approved):
Looks good

rcj-siteimprove (approved):

PR #1456, opened 2023-07-19
Fix false positive bug in R75

rcj-siteimprove:
In v0.65.0, there were two changes in behavior in R75.
In
```html
<div style="background: #eee; font-size: 0px">
    <div style="font-size: 14px">world</div>
</div>
```
R75 was previously not applicable to the first `div` because the applicability explicitly checked for `font-size: 0px`. In v0.65.0 that check was removed and we instead expected `isVisible` to filter out elements with `font-size: 0px`. It doesn't however because the element is still be visible so the rule now applies to more elements than before. We decided that was acceptable in this case because the rules passes this element since the non-whitespace text content is large enough. We still rewrote the applicability so that it checks for "elements with visible text" in stead of "visible elements with text" as that is closer to the intended applicability as stated on https://alfa.siteimprove.com/rules/sia-r75.

On the other hand, in
```html
<div style="background: #eee; font-size: 0px">
    hello
    <div style="font-size: 14px">world</div>
</div>
```
the first `div` would incorrectly be flagged as an error due to the fact that it now passed the applicability test and contained non-whitespace text `hello`. So the "real" bug, which was just uncovered by the change to the applicability, was actually in `.expectations()` since that didn't filter out the invisible `hello` when checking for font-size >= 9px.

### DIFF
--- a/packages/alfa-rules/src/sia-r75/rule.ts
+++ b/packages/alfa-rules/src/sia-r75/rule.ts
@@ -40,14 +40,15 @@ export default Rule.Atomic.of<Page, Element>({
           and(
             hasNamespace(Namespace.HTML),
             not(hasName("sup", "sub")),
-            (node) =>
+            not((node) =>
               visibleTextCache
                 .get(node, () =>
                   node
                     .descendants(Node.fullTree)
                     .filter(and(isText, isVisible(device)))
                 )
-                .some(() => true),
+                .isEmpty()
+            ),
             hasCascadedStyle(`font-size`, () => true, device)
           )
         );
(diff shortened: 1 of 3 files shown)


packages/alfa-rules/src/sia-r75/rule.ts
```diff
@@ -28,16 +32,22 @@ export default Rule.Atomic.of<Page, Element>({
...
-            // If the font-size ultimately computes to size 0, the element is not
-            // visible.
+            (node) =>
+              visibleTextCache
+                .get(node, () =>
+                  node
+                    .descendants(Node.fullTree)
+                    .filter(and(isText, isVisible(device)))
+                )
+                .some(() => true),
```
Jym77, line 50:
```suggestion
            not((node) =>
              visibleTextCache
                .get(node, () =>
                  node
                    .descendants(Node.fullTree)
                    .filter(and(isText, isVisible(device)))
                )
                .isEmpty()),
```
Makes the intention clearer.

Jym77 (changes requested):
Minor polish.

PR #1464, opened 2023-09-01
Use layout to improve handling of interposed elements

rcj-siteimprove:
With this PR we now try to automatically answer the question of which interposed elements can be ignored using layout.

This will affect SIA-R66 and SIA-R69 so that they won't need to ask about ignored interposed descendants if layout is available.

In addition we prepare for getting device dependent boxes by adding a helper function `getBoundingBox(Element, Device)` to `alfa-style` even though we do not yet store information about which device was used to collect the boxes.

### DIFF
--- a/packages/alfa-rules/src/common/expectation/contrast.ts
+++ b/packages/alfa-rules/src/common/expectation/contrast.ts
@@ -249,7 +249,6 @@ function getIntersectors(
 
   const elementBox = getBoundingBox(element, device);
 
-  // If just one of the elements involved doesn't have a bounding box, we regard that as an invalid state for deciding intersectors and return `None`
   if (
     !elementBox.isSome() ||
     Iterable.some(candidates, (candidate) =>
--- a/packages/alfa-rules/src/common/expectation/contrast.ts
+++ b/packages/alfa-rules/src/common/expectation/contrast.ts
@@ -5,7 +5,8 @@ import { Element, Node, Text } from "@siteimprove/alfa-dom";
 import { Iterable } from "@siteimprove/alfa-iterable";
 import { Set } from "@siteimprove/alfa-set";
 
-import { Option } from "@siteimprove/alfa-option";
+import { None, Option } from "@siteimprove/alfa-option";
+import { getBoundingBox } from "@siteimprove/alfa-style/src/element/element";
 import { expectation } from "../act/expectation";
 import { Group } from "../act/group";
 import { Question } from "../act/question";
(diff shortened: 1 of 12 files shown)


rcj-siteimprove:
!pr extract

packages/alfa-style/src/element/helpers/get-bounding-box.ts
```diff
@@ -0,0 +1,18 @@
...
+/**
+ * @public
+ * Gets the bounding box, corresponding to a specific device, of an element
+ *
+ * @privateRemarks
+ * We don't use the passed in device yet, but later we should use it to ensure the device used to collect the bounding box corresponds to the current device
+ */
+export function getBoundingBox(
+  element: Element,
+  device: Device
```
Jym77, line 15: **Suggestion:** This should also accept a `Context` parameter (defaulting to `Context.empty()`), and return `None` if it is not `Context.empty` (since we assume layout is grabbed on empty contexts).

packages/alfa-rules/src/common/expectation/contrast.ts
```diff
@@ -223,3 +230,42 @@ export function contrast(foreground: RGB, background: RGB): number {
...
+    return None;
+  }
+
+  return Option.of(
+    Iterable.filter(
+      candidates,
+      (canditate) =>
+        elementBox
+          .get()
+          .intersects(getBoundingBox(canditate, device).getUnsafe()) // Precense of the box is guaranteed by the above check
```
Jym77, line 268:
```suggestion
          .intersects(getBoundingBox(canditate, device).getUnsafe()) // Presence of the box is guaranteed by the above check
```

packages/alfa-rules/src/common/expectation/contrast.ts
```diff
@@ -223,3 +230,42 @@ export function contrast(foreground: RGB, background: RGB): number {
 
   return round(contrast * 100) / 100;
 }
+
+/**
+ * Finds elements from a collection of candidate that intersect with a given element
+ *
+ * @remarks
+ * If `candidates` is non-empty and just one element doesn't have layout `None` is returned
```
Jym77, line 238:
```suggestion
 * If either the element or one of the `candidates` doesn't have layout, we can't fully decide intersection and return `None`.
```
**Note:** it almost feels like this should return a `Result<Iterable<Element>, string>` rather than an `Option` 🤔

packages/alfa-rules/src/common/expectation/contrast.ts
```diff
@@ -223,3 +230,42 @@ export function contrast(foreground: RGB, background: RGB): number {
...
+  device: Device
+): Option<Iterable<Element>> {
+  // If the collection of candidates is empty we don't need layout to determine that there are no intersectors
+  if (Iterable.isEmpty(candidates)) {
+    return Option.of(candidates);
+  }
+
+  const elementBox = getBoundingBox(element, device);
+
+  // If just one of the elements involved doesn't have a bounding box, we regard that as an invalid state for deciding intersectors and return `None`
```
Jym77, line 252:
```suggestion
```
I feel we do not need to duplicate the comment since it's already part of the TSdoc.

Jym77 (approved):
Some minor polish.

packages/alfa-rules/src/common/expectation/contrast.ts
rcj-siteimprove: I'll add a note about it on https://github.com/Siteimprove/alfa/issues/1037

Jym77 (approved):

PR #1474, opened 2023-10-02
Store original device layout was scraped with

rcj-siteimprove:
We should store the device used to scrape the layout to avoid using a layout that no longer applies if another device is afterwards used when auditing.

This PR includes:
* Passing the device along when constructing any part of the document so that it can be stored on the elements
* Weakly storing bounding box by device in a `Cache` on the element
* Replacing direct box-accessing by function to get it by device
* Enabling sending the device to tests through jsx
* Updating the tests that depends on the box
* Adding type parameter for serialization options to `Serializer` interface
* Elements are serialized with serialization options containing the device for which the box should be serialized


rcj-siteimprove:
!pr extract

rcj-siteimprove:
!pr debgraph

rcj-siteimprove:
!pr depgraph

rcj-siteimprove:
!pr extract

packages/alfa-dom/src/h.ts
```diff
@@ -37,9 +38,10 @@ export function h<N extends string = string>(
   attributes?: Array<Attribute> | Record<string, string | boolean>,
   children?: Array<Node | string>,
   style?: Array<Declaration> | Record<string, string>,
-  box?: Rectangle
+  box?: Rectangle,
+  device?: Device
```
Jym77, line 42: **Suggestion:** This (and the same in other functions and in JSX) can default to `Device.standard()`, that way we don't need to provide a device in tests that also use `Device.standard()` by default.

packages/alfa-dom/src/jsx.ts
```diff
@@ -33,7 +35,11 @@ export function jsx<N extends string = string>(
         continue;
 
       case "box":
-        box = Rectangle.from(value as Rectangle.JSON);
+        const deviceAndBox = value as { device: Device } & Rectangle.JSON;
+
+        box = Rectangle.from(deviceAndBox);
+        device = deviceAndBox.device;
```
Jym77, line 41: **Question:** This seems needlessly complex to add the device to `deviceAndBox` and extract it directly.
Isn't it sufficient to just pass it when calling `h`?

packages/alfa-dom/src/node.ts
```diff
@@ -243,6 +244,11 @@ export interface Node {
 export namespace Node {
   export interface JSON<T extends string = string> extends tree.Node.JSON<T> {}
 
+  export interface SerializationOptions
+  {
+    device: Device
+  }
+
```
Jym77, line 251:
```suggestion
  export interface SerializationOptions  {
    device: Device
  }

```

packages/alfa-dom/src/node.ts
```diff
@@ -319,33 +325,36 @@ export namespace Node {
     Traversal.nested
   );
 
-  export function from(json: Element.JSON): Element;
+  export function from(json: Element.JSON, device: Option<Device>): Element;
```
Jym77, line 328:
```suggestion
  export function from(json: Element.JSON, device?: Device): Element;
```
We should make these optional argument (and build the `Option` if needed), in order to not completely break the API.

packages/alfa-dom/src/node/document.ts
```diff
@@ -109,10 +110,13 @@ export namespace Document {
   /**
    * @internal
    */
-  export function fromDocument(json: JSON): Trampoline<Document> {
-    return Trampoline.traverse(json.children ?? [], Node.fromNode).map(
-      (children) => Document.of(children, json.style.map(Sheet.from))
-    );
+  export function fromDocument(
+    json: JSON,
+    device: Option<Device>
```
Jym77, line 115:
```suggestion
    device?: Device
```

packages/alfa-dom/src/node/element.ts
```diff
@@ -312,7 +319,7 @@ export class Element<N extends string = string>
       style: this._style.map((style) => style.toJSON()).getOr(null),
       shadow: this._shadow.map((shadow) => shadow.toJSON()).getOr(null),
       content: this._content.map((content) => content.toJSON()).getOr(null),
-      box: this._box.map((box) => box.toJSON()).getOr(null),
+      box: Option.from(options).flatMap(o => this._boxes.get(o.device)).map(box => box.toJSON()).getOr(null)
```
Jym77, line 322: I think the logic of checking the options for a device would be a bit clearer with optional chaining:
```suggestion
      box: options?.device === undefined ? null : this._boxes.get(options.device).map(box => box.toJSON()).getOr(null)
```

packages/alfa-style/src/element/helpers/get-bounding-box.ts
```diff
@@ -7,9 +7,6 @@ import { Context } from "@siteimprove/alfa-selector";
 /**
  * @public
  * Gets the bounding box, corresponding to a specific device, of an element
- *
- * @privateRemarks
- * We don't use the passed in device yet, but later we should use it to ensure the device used to collect the bounding box corresponds to the current device
  */
 export function getBoundingBox(
```
Jym77, line 11: **Question:** Are we still using this placeholder, or can it be totally removed?

Jym77 (changes requested):
Looks good, some clean up and possible improvements.

packages/alfa-style/src/element/helpers/get-bounding-box.ts
rcj-siteimprove: Nope, I have removed it.

rcj-siteimprove:
!pr extract

packages/alfa-dom/src/jsx.ts
rcj-siteimprove: I am not sure I completely get what you mean by that.

Is it the fact that `device` is put inside the `box` attribute like so `<div box={{ device, x: 8, ... }}>` and would it be better to have it as a separate attribute like so `<div device={{ device }} box={{ x: 8, ... }}>`? 

In any case we need to extract the device somehow inside the `jsx` function so that it can be passed on to `h` as far as I can tell.

It would of course be nice to only have to pass the device in one time like so `h.document(<div box={{...}}></div>, device)`, but the problem is that the elements have already been constructed by `jsx` before `h.document` is called and it's too late to pass in the device because the boxes does not get stored on the elements if there is no device at construction time 🤔

rcj-siteimprove:
!pr extract

Jym77 (approved):
Looks good 👍

PR #1485, opened 2023-10-18
Consider that images scale to specified dimensions, rather than overflowing.

### LINKED ISSUE #1476: Investigate visibility of tracking pixel
It seems Alfa is detecting a tracking pixel as visible despite having `height` and `width` set at 0.

- [x] Follow up on CST-1486

Jym77:
Resolves #1476

Currently very focused on images, since tracking pixels sounds like a common use case. We will expand that further if we need to and encounter other replaced elements that behave in the same way.


packages/alfa-style/src/node/predicate/is-clipped.ts
```diff
@@ -18,6 +18,10 @@ const { isElement, hasBox } = Element;
 const { not, or, test } = Predicate;
 const { and } = Refinement;
 
+// function show(msg: string): (_: any, res: boolean) => void {
+//   return (_, res) => console.log(`${msg}: ${res}`);
+// }
+
```
rcj-siteimprove, line 24: **Question:** Should this be cleaned?

Issue #381, opened 2020-09-14
Provide extended diagnostics for rules


kasperisager:
This issue will track progress on providing extended outcome diagnostics for the rules that would benefit from it.

- [ ] **R16:** Include information on _which_ required states or properties are missing.

PR #1502, opened 2023-11-13
Send accessible name diagnostic

rcj-siteimprove:
Part of #381 

Accessible name will be sent as part of the diagnostics for R15, R39, R41 and R81


rcj-siteimprove:
!pr extract

Jym77 (approved):

PR #1505, opened 2023-11-14
Factor heading diagnostics

rcj-siteimprove:
`WithPreviousHeading` and `WithNextHeading` have been factored into `WithOtherHeading`.


packages/alfa-rules/src/common/diagnostic/with-other-heading.ts
```diff
@@ -0,0 +1,138 @@
+import { Diagnostic } from "@siteimprove/alfa-act";
+import { Element } from "@siteimprove/alfa-dom";
+import { Hash } from "@siteimprove/alfa-hash";
+import { Option } from "@siteimprove/alfa-option";
+
+/**
+ * @internal
+ */
+export type HeadingPosition = "previous" | "next" | "unknown";
```
Jym77, line 9: **Suggestion:** I think we can stay with only the first two for now, and add more if need arise. Whatever we change later, it's pretty sure that `unknown` won't be the text we want.

packages/alfa-rules/src/sia-r53/rule.ts
```diff
@@ -31,97 +32,84 @@ export default Rule.Atomic.of<Page, Element>({
...
+              device,
+              tee(
+                (currentLevel) =>
+                  hasHeadingLevel(
+                    device,
+                    tee(
+                      (previousLevel) => previousLevel >= currentLevel - 1,
+                      (p) => {
+                        previousLevel = p;
+                      },
```
Jym77, line 55: **Nitpick:** I think we can have prettier inline that by using parenthesis instead of brackets 🤔

packages/alfa-rules/src/sia-r53/rule.ts
```diff
@@ -31,97 +32,84 @@ export default Rule.Atomic.of<Page, Element>({
...
+                    ),
+                  )(previousHeading),
+                (c) => {
+                  currentLevel = c;
+                },
```
Jym77, line 60: Same here.

Jym77 (approved):

Issue #1486, opened 2023-10-18
Support `:where` pseudo-class


Jym77:
- [ ] Follow up on CI-4502

PR #1514, opened 2023-11-28
Compute selector specificity at parse time.

Jym77:
Part of #1486

Move the computation of selectors' specificity at the time the `Selector` object is built (usually at parse time). This allows for specific selectors to "cheat" with specificity calculation without making the shared code unreadable.

There is an unrelated change to `alfa-test` to make it easier to write ramdom tests, and this is refactored from `alfa-map` which was the only package doing random tests so far.


Jym77:
!pr extract

rcj-siteimprove (approved):
Looks good

PR #1523, opened 2023-12-04
Clone-replace for `alfa-dom` `Node`s

rcj-siteimprove:
We introduce new functions for cloning a DOM-tree with the option of replacing elements based on a predicate. A clone will be a new instance structurally equal to the original, but physically different.

The main use case is for a situation where we want to audit a document with some elements replaced to see how the modification impacts the result and since trees in Alfa are immutable, we do that by cloning and replacing.


rcj-siteimprove:
!pr extract

packages/alfa-dom/src/node/element.ts
```diff
@@ -447,6 +448,59 @@ export namespace Element {
...
+                declaration.value,
+                declaration.important,
+              ),
+            ),
+          );
+        }),
+        deviceOption.flatMap((d) => element.getBoundingBox(d)),
+        deviceOption,
+        element.externalId,
+        element.extraData,
```
Jym77, line 489: **Issue:** Also realising that this will actually share the same `extraData` object, which might create some confusion.
```typescript
const foo = h.element(…, { extraStuff: "foo" });

const bar = Node.clone(foo);
bar.extraData.extraStuff = "bar";
```
This also change `foo.extraData.ExtraStuff`.

On on side, it is not that bad because Alfa doesn't care, but it feels like this can still create confusion in users. We could:
* Keep it that way. Make it clear that `extraData` is copied with referencial equality, with all that it implies.
* Discard `extraData` when cloning, make it clear that it is the behaviour.
* Offer some API to pass an `extraData` cloning function.

The third option seems a bit crazy. Given that we do not currently have any actual use for `extraData`, I'd rather go for the second one. We can revisit later if needed.

---
Also, we should make it clear in the function's documentation that `externalId` is going to be copied to the clone. This is likely the desired behavior (at least it is for our use case), but it could be confusing if undocumented.

packages/alfa-dom/src/node.ts
```diff
@@ -380,6 +380,108 @@ export namespace Node {
     }
   }
 
+  export function clone(
+    node: Element,
+    newElements?: Element[],
+    predicate?: Predicate<Element>,
+    device?: Device,
+  ): Element;
```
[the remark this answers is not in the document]
rcj-siteimprove: For the record we agreed offline to not do this right now as it caused other difficulties.

packages/alfa-dom/src/node/document.ts
```diff
@@ -125,6 +127,30 @@ export namespace Document {
...
+    device?: Device,
+  ): Trampoline<Document> {
+    return Trampoline.traverse(document.children(), (child) => {
+      if (Element.isElement(child) && predicate(child)) {
+        return Trampoline.done(newElements);
+      }
+
+      return Node.cloneNode(child, newElements, predicate, device).map(
+        (node) => [node],
+      );
```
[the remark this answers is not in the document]
rcj-siteimprove: Regarding the suggestion, I think the challenge is that the mapper should return some kind of `Trampoline<U>` so I think we do need to lift the single `node` into `[node]` inside the `.map()` function, otherwise we end up with an array of trampolines it seems 🤔

Regarding the iterators, might be possible, but not totally sure how well it would play together with `Trampoline`

packages/alfa-dom/src/node.ts
```diff
@@ -380,6 +380,108 @@ export namespace Node {
...
+    return cloneNode(node, newElements, predicate, device).run();
+  }
+
+  export function cloneNode(
+    node: Node,
+    newElements: Element[] = [],
+    predicate: Predicate<Element> = () => false,
+    device?: Device,
+  ): Trampoline<Node> {
+    if (Element.isElement(node)) {
```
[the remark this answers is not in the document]
rcj-siteimprove: Done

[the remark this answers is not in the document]
rcj-siteimprove: Not sure about this one, let's have a talk about it.

packages/alfa-dom/test/node.spec.tsx
```diff
@@ -61,3 +64,51 @@ test(`#tabOrder() correctly handles shadow roots with slotted elements after the
...
+    [
+      h.element(
+        "p",
+        [h.attribute("title", "foo")],
+        [h.text("hello")],
+        [],
+        Namespace.HTML,
+        Rectangle.of(1, 2, 3, 4),
+        device,
+      ),
```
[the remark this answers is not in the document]
rcj-siteimprove: Done

packages/alfa-dom/test/node.spec.tsx
```diff
@@ -61,3 +64,51 @@ test(`#tabOrder() correctly handles shadow roots with slotted elements after the
...
+    ],
+    [h.sheet([h.rule.style("p", { background: "green" })])],
+    "bar",
+    { extraStuff: "baz" },
+  );
+
+  const clonedDoc = Node.clone(doc);
+
+  t.deepEqual(clonedDoc.toJSON(), doc.toJSON());
```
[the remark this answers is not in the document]
rcj-siteimprove: Done

rcj-siteimprove:
!pr extract

packages/alfa-dom/src/node/element.ts
rcj-siteimprove: I went with the second suggestion and discarded the `extraData`.

rcj-siteimprove:
!pr extract

rcj-siteimprove:
Ready for re-review, though still with a few un-resolved comments that could perhaps be split into new tasks, if they are non-blocking for this.

packages/alfa-dom/src/node/element.ts
```diff
@@ -447,6 +448,67 @@ export namespace Element {
...
+          clonedElement._attachShadow(
+            Shadow.of(
+              shadow
+                .children()
+                .map((child) => Node.clone(child, options, device)),
+              shadow.style,
+              shadow.mode,
+              shadow.externalId,
+              shadow.extraData,
+            ),
```
Jym77, line 500: **Suggestion:** we should probably have a `Shadow.clone` instead of inlining it here.

packages/alfa-dom/test/node.spec.tsx
```diff
@@ -61,3 +63,75 @@ test(`#tabOrder() correctly handles shadow roots with slotted elements after the
...
+  const clonedDoc = Node.clone(doc);
+
+  t.notEqual(clonedDoc, doc);
+
+  t.deepEqual(clonedDoc.toJSON(), doc.toJSON());
+});
+
+test(`Node.clone() clones shadow`, (t) => {
+  const div = <div>hello</div>;
+  const shadow = h.shadow([<div>foo</div>]);
```
Jym77, line 89: **Note:** Our JSX syntax (actually, `h`) will do shadow magic upon encountering a children that is a shadow tree. So:
```
const div = <div>{shadow}</div>
```
should be enough.
https://github.com/Siteimprove/alfa/blob/main/packages/alfa-dom/src/h.ts#L103-L127
https://github.com/Siteimprove/alfa/blob/main/packages/alfa-dom/test/h.spec.ts#L39

packages/alfa-dom/test/node.spec.tsx
```diff
@@ -61,3 +63,75 @@ test(`#tabOrder() correctly handles shadow roots with slotted elements after the
...
+
+  t.deepEqual(
+    div.shadow.getUnsafe().toJSON(),
+    clonedDiv.shadow.getUnsafe().toJSON(),
+  );
+});
+
+test(`Node.clone() clones content`, (t) => {
+  const div = <div>hello</div>;
+  const content = h.document([<div>foo</div>]);
```
Jym77, line 105: Same remark with content documents.

Jym77 (changes requested):
Looks good. Some minor polish and streamlining.

packages/alfa-dom/test/node.spec.tsx
rcj-siteimprove: Done

rcj-siteimprove:
!pr extract

packages/alfa-dom/test/node.spec.tsx
```diff
@@ -101,8 +99,8 @@ test(`Node.clone() clones shadow`, (t) => {
 });
 
 test(`Node.clone() clones content`, (t) => {
-  const div = <div>hello</div>;
   const content = h.document([<div>foo</div>]);
+  const div = <div>{content}</div>;
 
   div._attachContent(content);
```
Jym77, line 105:
```suggestion
```

packages/alfa-dom/src/node/element.ts
```diff
@@ -488,21 +490,13 @@ export namespace Element {
...
-              shadow
-                .children()
-                .map((child) => Node.clone(child, options, device)),
-              shadow.style,
-              shadow.mode,
-              shadow.externalId,
-              shadow.extraData,
-            ),
-          );
+          clonedElement._attachShadow(Shadow.clone(shadow, options, device));
```
Jym77, line 493: **Nitpick:** We can inline `shadow`, to match what is done with content document.

Jym77 (approved):

PR #1524, opened 2023-12-04
Accept "of selector" syntax in `:nth-[last]-child` pseudo-classes

### LINKED ISSUE #1511: Support "of selector" syntax in `:nth-(last-)child` selectors
We currently only support `:nth-child(An+B)` but not the `nth-child(An+B of selector)` syntax. Same for `:nth-last-child`.
This also affects calculation of selectivity.
https://www.w3.org/TR/selectors/#the-nth-child-pseudo
https://www.w3.org/TR/selectors/#specificity

Jym77:
Resolves #1511

### DIFF
--- a/.changeset/breezy-horses-smell.md
+++ b/.changeset/breezy-horses-smell.md
@@ -1,0 +1,5 @@
+---
+"@siteimprove/alfa-selector": minor
+---
+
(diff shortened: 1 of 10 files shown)


rcj-siteimprove (approved):
Looks good

PR #1538, opened 2023-12-18
Assume audio is playing and that it has controls based on attributes

### LINKED ISSUE #1326: Media rules (and Applicability) could auto-detect embedded play button and `autoplay` attribute
When the `<audio>` element has `controls`, Alfa could assume that it includes a play button and avoid asking the question.

Alfa could use the `autoplay` attribute and assume the audio is playing.

(R49 is already doing that)

rcj-siteimprove:
Resolves #1326

The shared applicability for audio rules has been updated to not ask if audio is playing when the autoplay attribute is present and to not ask where play button is when controls attribute is present.


packages/alfa-rules/test/sia-r23/rule.spec.tsx
```diff
@@ -200,6 +200,35 @@ test(`evaluate() cannot tell if questions are left unanswered`, async (t) => {
...
+    await evaluate(
+      R23,
+      { document },
+      oracle({
+        "is-audio-streaming": false,
+        "is-playing": false,
+      }),
+    ),
+    [cantTell(R23, target, undefined, Outcome.Mode.SemiAuto)],
+  );
```
Jym77, line 229: **Issue:** These should be examples where the outcome is **not** `cantTell` to emphasize that we can get a definitive outcome without answering that question.
(it would be Inapplicable without he change, but I feel it would be less confusing to reread if we see a Passed or Failed result in this case)

Jym77 (changes requested):
* We should do the same for the video rules.
* We should also add examples for the composite rules.

packages/alfa-rules/test/sia-r23/rule.spec.tsx
rcj-siteimprove: Done

Jym77 (approved):

PR #1541, opened 2023-12-20
Align R18 with specification

### LINKED ISSUE #1475: R18 and <input type="file">
These fail R18, and I think that they should pass:
```
Fails R18: <input type="file" aria-required="true" aria-label="field 1"><br>
<br>
Fails R18: <input type="file" aria-invalid="false" aria-label="field 2"><br>
```
I think they should pass because of [this ACT rule change](https://github.com/act-rules/act-rules.github.io/issues/1573).  And [ARIA in HTML for &lt;input type="file"&gt;](https://w3c.github.io/html-aria/#el-input-file) says that the state and property allowances are "Global aria-* attributes, aria-disabled, aria-invalid and aria-required attributes."

rcj-siteimprove:
Resolves #1475


.changeset/violet-coats-tie.md
```diff
@@ -0,0 +1,5 @@
+---
+"@siteimprove/alfa-rules": minor
```
Jym77, line 2:
```suggestion
"@siteimprove/alfa-rules": patch
```

Jym77 (changes requested):

Jym77 (approved):

PR #1546, opened 2024-01-08
`@siteimprove/alfa-dom`: Break circular import

Jym77:
Breaking a circular import `index.ts` -> `node/element.ts` -> `index.ts` by importing directly from the local file.
This seems to resolve the problems when rebuilding the package 🤞


rcj-siteimprove (approved):
Good catch 🎉

PR #1547, opened 2024-01-09
Fix tree traversal of flat tree

Jym77:
Fix a bug in tree traversal of flat tree for the rare case of comments being direct child of the shadow root.

We had an exception for elements and text to skip over the shadow root and return the host instead when querying the parent in the flat tree. This is correct because in the flat tree the shadow tree has been inserted by "merging" the host and the root. However, that exception was missing for comments, creating an issue in these cases.

Additionally fixed a missing tree traversal option in `getNodeBetween`.

### DIFF
--- a/.changeset/dirty-lizards-explode.md
+++ b/.changeset/dirty-lizards-explode.md
@@ -1,0 +1,5 @@
+---
+"@siteimprove/alfa-dom": patch
+---
+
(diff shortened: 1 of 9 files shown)


rcj-siteimprove (approved):
It looks good, there seems to be an issue in the CI though

Jym77:
!pr extract

rcj-siteimprove (approved):

Issue #1533, opened 2023-12-13
Handle Shadow DOM related pseudo-classes and -elements


Jym77:
`:host`, `:host-context` and `::slotted` are not correctly handled.
This will require some tweak to the cascade building to handle context.
It may require quite a tweak to style building since `style` attribute only comes at priority 3 in cascade sorting, thus not trumping context (priority 2). Maybe we'll need to move the `style` attribute processing to cascade time rather than style time.

PR #1550, opened 2024-01-15
Add `style` attribute and importance to cascade sort

### LINKED ISSUE #1532: `!important` User-Agent rules should not be overwritten by `!important` author rules.
```typescript
// non-important UA rule can be overwritten by non-important author rule
const div = <div class="foo">Hello</div>;
// important (test!) UA rule cannot be overwritten by important author rule
const noscript = <noscript class="bar">World</noscript>;
const document2 = h.document(
  [div, noscript],
  [
    h.sheet([
      h.rule.style(".foo", { display: "inline" }),
      h.rule.style(".bar", { display: "inline !important" }),
    ]),
  ],
);

console.dir(Style.from(div, device).computed("display").toJSON(), {
  depth: null,
}); // inline <- good
console.dir(Style.from(noscript, device).computed("display").toJSON(), {
  depth: null,
}); // inline <- wrong
```

Part of the difficulty is that a same rule can contain both `!important` and not important declarations (e.g. `div { display: inline; color: red !important}`). This require splitting up the rule into its components at cascade time. Currently, importance of declarations is handled when building style (via the `shouldOverride` checker), this should likely be modved at cascade time.

Given that we only have 3 `important` UA rules, and that they are quite specific on `hidden` elements and `<noscript>`, none of them seem likely that author will try and override them (and even less rely on the fact that their attempt fail), this is likely minor and can be ignored for now.

Jym77:
Resolves #1532 

With [CSS cascade level 5](https://drafts.csswg.org/css-cascade-5/#cascade-sort) (compared to [level 4](https://drafts.csswg.org/css-cascade-4/#cascade-sort)):
* Importance reverse sorting not just for origin, but also for encapsulation context. Origin was OK to ignore since we have very few important user agent rules and they are unlikely to be temptatively overridden by author rules (and even less relying on the fact that it doesn't work). Encapsulation context, however, is required to solve #1533.
* Style attribute now has its own step in the cascade sorting, making it more natural to handle it there.

With this PR, both importance and style attribute are now part of the Cascade. This streamline `Style` building since all declarations are provided, and in decreasing cascade order. So, only parsing of the value is really required. Cascade becomes a bit more complex and, notably, elements with `style` attribute creates new branches in the rules tree, thus decreasing the possibility of sharing. Given that `style` attribute declarations have higher precedence than style sheet declarations, they should be close to the leaves and this shouldn't be a big issue.

`Style.from()` still handles the `style` attribute for elements that are not part of a `Document` (hence have no access to style sheet and no associated cascade). This should be mostly a use case for our own tests.


Jym77:
!pr extract

rcj-siteimprove (approved):

PR #1551, opened 2024-01-16
Make punctuation-only texts inapplicable in R69

### LINKED ISSUE #1166: SIA-R69: Consider making non-text content not applicable
Decorative text like `|` , `&`, etc. should not be flagged as an issue.

rcj-siteimprove:
Resolves #1166

We just ignore text that only consists of punctuation.

### DIFF
--- a/packages/alfa-rules/src/sia-r69/rule.ts
+++ b/packages/alfa-rules/src/sia-r69/rule.ts
@@ -39,5 +39,5 @@ export default Rule.Atomic.of<
 });
 
 function isOnlyPunctuation(text: Text): boolean {
-  return text.data.replace(/\p{P}|\p{S}|\p{Cf}/gu, "").length === 0;
+  return /^[\p{P}\p{S}\p{Cf}]+$/gu.test(text.data);
 }
--- a/packages/alfa-rules/src/sia-r69/rule.ts
+++ b/packages/alfa-rules/src/sia-r69/rule.ts
@@ -1,5 +1,6 @@
 import { Rule } from "@siteimprove/alfa-act";
 import { Element, Text } from "@siteimprove/alfa-dom";
+import { Predicate } from "@siteimprove/alfa-predicate";
 import { Criterion } from "@siteimprove/alfa-wcag";
 import { Page } from "@siteimprove/alfa-web";
 
@@ -10,6 +11,8 @@ import { nonDisabledTexts } from "../common/applicability/non-disabled-texts";
 import { hasSufficientContrast } from "../common/expectation/contrast";
 
 import { Scope, Stability, Version } from "../tags";
+
+const { not } = Predicate;
 
 export default Rule.Atomic.of<
   Page,
@@ -23,7 +26,9 @@ export default Rule.Atomic.of<
   evaluate({ device, document }) {
     return {
       applicability() {
-        return nonDisabledTexts(document, device);
+        return nonDisabledTexts(document, device).filter(
+          not(isOnlyPunctuation),
+        );
       },
 
       expectations(target) {
@@ -32,3 +37,7 @@ export default Rule.Atomic.of<
     };
   },
 });
+
+function isOnlyPunctuation(text: Text): boolean {
+  return text.data.replace(/\p{P}|\p{S}|\p{Cf}/gu, "").length === 0;
+}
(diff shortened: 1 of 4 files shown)


packages/alfa-rules/src/sia-r69/rule.ts
```diff
@@ -23,7 +26,9 @@ export default Rule.Atomic.of<
   evaluate({ device, document }) {
     return {
       applicability() {
-        return nonDisabledTexts(document, device);
+        return nonDisabledTexts(document, device).filter(
+          not(isOnlyPunctuation),
+        );
```
Jym77, line 31:
```suggestion
        return nonDisabledTexts(document, device).reject(
          isOnlyPunctuation,
        );
```
should work.

packages/alfa-rules/test/sia-r69/rule.spec.tsx
```diff
@@ -395,6 +395,40 @@ test(`evaluate() is inapplicable to the text that is part of a label of a disabl
   t.deepEqual(await evaluate(R69, { document }), [inapplicable(R69)]);
 });
 
+test("evaluate() is inapplicable to text that would otherwise pass if it was not only punctuation", async (t) => {
+  const target = h.text(",./;'[]\\-=?<>:\"{}|_+!@#$%^&*()");
+
+  const document = h.document([
+    <html
+      style={{
+        backgroundImage: "url('foo.png')",
```
Jym77, line 404: **Suggestion:** That technically would make it a CantTell if it was applicable. `backgroundColor` would match the description better.

packages/alfa-rules/test/sia-r69/rule.spec.tsx
```diff
@@ -395,6 +395,40 @@ test(`evaluate() is inapplicable to the text that is part of a label of a disabl
...
+  t.deepEqual(await evaluate(R69, { document }), [inapplicable(R69)]);
+});
+
+test("evaluate() is inapplicable to text that would otherwise fail if it was not only punctuation", async (t) => {
+  const target = h.text(",./;'[]\\-=?<>:\"{}|_+!@#$%^&*()");
+
+  const document = h.document([
+    <html
+      style={{
+        backgroundImage: "url('foo.png')",
```
Jym77, line 421: Same here. Which gets even stranger since the test case is now exactly the same.

Jym77 (changes requested):

Jym77:
@dan-tripp-siteimprove FYI.

packages/alfa-rules/test/sia-r69/rule.spec.tsx
rcj-siteimprove: I based the tests on two other tests that were passing and failing resp., but I forgot to copy the oracle as well. Now I have based the tests on two other tests that are not using an oracle at all to keep it simple. I've confirmed that adding a single letter to the text in both tests results in passing and failing outcome resp.

packages/alfa-rules/test/sia-r69/rule.spec.tsx
```diff
@@ -418,7 +418,8 @@ test("evaluate() is inapplicable to text that would otherwise fail if it was not
   const document = h.document([
     <html
       style={{
-        backgroundImage: "url('foo.png')",
+        backgroundImage: "linear-gradient(#000 50%, transparent 50%)",
+        backgroundColor: "#000",
         color: "#000",
```
Jym77, line 423: Wouldn't that be inapplicable as invisible because it's black-on-black?

Jym77 (approved):

packages/alfa-rules/test/sia-r69/rule.spec.tsx
rcj-siteimprove: For the record, this is answered in above answer

dan-tripp-siteimprove:
Hi @rcj-siteimprove - today I noticed that this PR did it for R69 and not R66.  [Test page here](https://dan-tripp-siteimprove.github.io/contrast-pipe-char-test-page--2025-01-10--QHBRssM.html).  IMO R66 should get the same treatment that this PR gave to R69, because WCAG uses the same language (text = "expressing something in human language") for both success criteria.  If you agree, may I submit a PR to this effect?  :)

Jym77:
Yes R66 should have the same treatment. Sorry we missed that 🙈 
We should move that to the common Applicability, I guess.

dan-tripp-siteimprove:
Cool.  I would like to submit a PR for this.   (Though I'm not sure when.)  Thank you both.

PR #1553, opened 2024-01-18
Support `:host` pseudo class

Jym77:
Part of #1533 

* Include encapsulation context in CSS cascade precedence.
* Correctly parse `:host` selector.
* Include "shadow rules" (i.e., `:host`, for now) from the shadow tree style sheet into shadow hosts' cascade.
* Exclude "shadow rules" from the cascade of the shadow tree they are defined on.

Given the nature of `:host`, full matching cannot be done within the selector itself but has to be delegated to the cascade (and selector map).

### DIFF
--- a/.changeset/curly-apes-repeat.md
+++ b/.changeset/curly-apes-repeat.md
@@ -1,0 +1,9 @@
+---
+"@siteimprove/alfa-cascade": minor
+---
+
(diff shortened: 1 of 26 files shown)


rcj-siteimprove (approved):
Looks good

PR #1559, opened 2024-01-24
Support `::slotted` pseudo-element

### LINKED ISSUE #1533: Handle Shadow DOM related pseudo-classes and -elements

Jym77:
* Support `::slotted` pseudo-element.
* Fix matching of `:host(-context)` as leftmost selector in a complex selector.
* Switch encapsulation contexts from boolean to numbers since a single element may be affected by declarations from more than two trees.

Resolves #1533


Jym77:
!pr extract

Jym77:
!pr extract

rcj-siteimprove (approved):
Looks good

PR #1560, opened 2024-01-26
Fix ownership of rules

Jym77:
Randomly found that bug 🤷 
This notably caused https://github.com/Siteimprove/alfa/blob/main/packages/alfa-cascade/src/user-agent.ts#L40-L44 to not be considered as a User Agent rule due to the nesting, and therefore lose cascade versus important Author rules.


rcj-siteimprove (dismissed):
Good catch 👍🏼

Jym77:
!pr extract

rcj-siteimprove (approved):

PR #1564, opened 2024-01-30
Implement SIA-R111 (Target size)

### LINKED ISSUE #1552: Implement SIA-R111: Target Size
See https://github.com/act-rules/act-rules.github.io/pull/1919

We should make a simple first version which probably will have many false negatives (but shouldn't have any false positives).

The ideas for the first version:
- No z-index analysis to find correct shape, we just use bounding box
- Ignore scrolling that will affect shape

rcj-siteimprove:
Resolves #1552


packages/alfa-rules/src/sia-r111/rule.ts
```diff
@@ -0,0 +1,63 @@
...
+import { expectation } from "../common/act/expectation";
+
+const { getElementDescendants } = Query;
+const { and } = Refinement;
+const { hasRole } = DOM;
+const { hasComputedStyle, isFocusable } = Style;
+
+export default Rule.Atomic.of<Page, Element>({
+  uri: "https://alfa.siteimprove.com/rules/sia-r111",
+  requirements: [], // TODO: Not sure how to determine these
```
Jym77, line 18:
```suggestion
  requirements: [Criterion.of("2.5.5")], 
```

packages/alfa-rules/test/sia-r111/rule.spec.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const document = h.document([target]);
+
+  t.deepEqual(await evaluate(R111, { document, device }), [
+    failed(R111, target, {
+      1: Outcomes.HasInsufficientSize,
+    }),
+  ]);
+});
+
+test("evaluate() is inapplicable to disabled button", async (t) => {
```
Jym77, line 52: **Issue:** we should have an example with `pointer-events: none`.

Jym77 (changes requested):
Well done!

Jym77 (approved):

PR #1566, opened 2024-02-05
Fix spacing in name computation

### LINKED ISSUE #1203: Accessible name from `aria-label` is not computed correctly in case of inline elements
```HTML
<span role="presentation" aria-label="Eli">
    <span aria-label="Garaventa">Zambino</span>
</span>
<span>the weird.</span>
```

The accessible name should be:
_My name is Eli the weird._
It is instead: _My name is Elithe weird._

[Link to Testable Statement](https://www.w3.org/wiki/AccName_1.1_Testable_Statements#Name_from_content)

Jym77:
Spaces were too eagerly trimmed when computing name from  descendants, resulting in them disappearing from the concatenation when they should still be here.

Also split the huge `name.ts` into its `Source` and `State` components. The actual changes are in adding the `spaceBefore` and `spaceAfter` properties, and handling them in the various `Name.from*` functions.

Resolves #1203, resolves 1054.


packages/alfa-aria/test/name.spec.tsx
```diff
@@ -1746,3 +1909,101 @@ test(".from() does not recurse into content documents", (t) => {
...
+  const target = (
+    <button>
+      <span>Hello</span>
+      <span> World</span>
+    </button>
+  );
+  h.document([target]);
+
+  t.equal(getName(target).value, "Hello World");
+});
```
Jym77, line 2009: @2biazdk FYI

rcj-siteimprove (approved):
Nice 🙌🏼

PR #1572, opened 2024-02-13
Add package `alfa-string`

Jym77:
We had a bit too many copies of some of the common string manipulations…


rcj-siteimprove (approved):
Nice 👍🏼

PR #1579, opened 2024-02-20
R111 improvements

### LINKED ISSUE #1576: R111 improvements
- [x] Inline target inapplicable
- [x] Pass user agent controlled targets
- [x] Send rectangle as part of extended diagnostics
- [x] Make rule inapplicable to invisible targets.

https://www.w3.org/WAI/WCAG21/Understanding/target-size.html

https://github.com/act-rules/act-rules.github.io/blob/f92d3bb4770468b2d9f60ef005bc7d3af1a293cd/_rules/target-size-enhanced-gi8qkf.md

https://github.com/act-rules/act-rules.github.io/blob/f92d3bb4770468b2d9f60ef005bc7d3af1a293cd/pages/glossary/rendered-on-a-line.md

rcj-siteimprove:
Resolves #1576

### DIFF
--- a/packages/alfa-rules/src/sia-r111/rule.ts
+++ b/packages/alfa-rules/src/sia-r111/rule.ts
@@ -46,7 +46,7 @@ export default Rule.Atomic.of<Page, Element>({
         return {
           1: expectation(
             test(
-              or(hasSufficientSize(44, device), isUserAgentControlled),
+              or(isUserAgentControlled, hasSufficientSize(44, device)),
               target,
             ),
             () => Outcomes.HasSufficientSize(box),
--- a/packages/alfa-rules/src/sia-r111/rule.ts
+++ b/packages/alfa-rules/src/sia-r111/rule.ts
@@ -7,6 +7,8 @@ import { Err, Ok } from "@siteimprove/alfa-result";
 import { Style } from "@siteimprove/alfa-style";
 import { Criterion } from "@siteimprove/alfa-wcag";
 import { Page } from "@siteimprove/alfa-web";
+import { Predicate } from "@siteimprove/alfa-predicate";
+import { Device } from "@siteimprove/alfa-device";
 
 import { expectation } from "../common/act/expectation";
 
@@ -14,6 +16,7 @@ import { WithBoundingBox } from "../common/diagnostic";
 
 const { getElementDescendants } = Query;
 const { and } = Refinement;
+const { or, test } = Predicate;
 const { hasRole } = DOM;
 const { hasComputedStyle, isFocusable } = Style;
 
@@ -38,12 +41,14 @@ export default Rule.Atomic.of<Page, Element>({
       },
 
       expectations(target) {
-        // Existence of bounding box is guaranteed by applicability
+        // Existence of a bounding box is guaranteed by applicability
         const box = target.getBoundingBox(device).getUnsafe();
-
         return {
           1: expectation(
-            box.width >= 44 && box.height >= 44,
+            test(
+              or(hasSufficientSize(44, device), isUserAgentControlled),
+              target,
+            ),
             () => Outcomes.HasSufficientSize(box),
             () => Outcomes.HasInsufficientSize(box),
           ),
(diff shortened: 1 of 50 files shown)


packages/alfa-rules/test/sia-r111/rule.spec.tsx
```diff
@@ -23,7 +23,26 @@ test("evaluate() passes button with clickable area of exactly 44x44 pixels", asy
...
+      type="checkbox"
+      box={{ device, x: 8, y: 8, width: 13, height: 13 }}
+    ></input>
+  );
+
+  const document = h.document([target]);
+
+  t.deepEqual(await evaluate(R111, { document, device }), [
+    passed(R111, target, {
+      1: Outcomes.HasSufficientSize(target.getBoundingBox(device).getUnsafe()),
```
Jym77, line 45: **Suggestion:** We should have a specific `Outcomes.IsUserAGentControlled` to return here.

packages/alfa-rules/src/sia-r111/rule.ts
```diff
@@ -54,11 +62,22 @@ export default Rule.Atomic.of<Page, Element>({
...
-    Diagnostic.of("Target has sufficient size"),
-  );
+  export const HasSufficientSize = (box: Rectangle) =>
+    Ok.of(WithBoundingBox.of("Target has sufficient size", box));
+
+  export const HasInsufficientSize = (box: Rectangle) =>
+    Err.of(WithBoundingBox.of("Target has insufficient size", box));
+}
+
+// This predicate is assumed to only be used on elements with bounding boxes which should be guaranteed by applicability
```
Jym77, line 72: **Chore:** make this into a real TSdoc string, so that it shows up in IDE's contextual help.

Jym77 (changes requested):
Looks good.
It will also need changeset (it adds the UA controlled exception).

Jym77 (dismissed):

packages/alfa-rules/src/sia-r111/rule.ts
```diff
@@ -54,11 +61,29 @@ export default Rule.Atomic.of<Page, Element>({
...
-    Diagnostic.of("Target has sufficient size"),
-  );
+  export const HasSufficientSize = (name: string, box: Rectangle) =>
+    Ok.of(WithBoundingBox.of("Target has sufficient size", name, box));
+
+  export const HasInsufficientSize = (name: string, box: Rectangle) =>
+    Err.of(WithBoundingBox.of("Target has insufficient size", name, box));
+
+  export const IsUserAgentControlled = (name: string) =>
+    Ok.of(WithName.of("Target is user agent controlled", name));
```
Jym77, line 71: This might get a bit annoying downstream if we provide incompatible diagnostics without a differenciating property.

I think for know it might be easier to always send a `WithBouindingBox`. We need to work out a better system for incompatible diagnostics, but that requires more sync and contracts with downstream usage.

Jym77 (changes requested):

Jym77:
> @Jym77 This should be ready for re-review. Not sure if we want the new diagnostic to inherit from `WithName`, was not sure how to best implement "multiple diagnostics". Let me know if you prefer another approach.

I think it's good for now. We'll need to revisit a bit the full diagnostic architecture at some point…

packages/alfa-rules/src/sia-r111/rule.ts
rcj-siteimprove: Done

Jym77 (dismissed):

packages/alfa-rules/test/sia-r111/rule.spec.tsx
```diff
@@ -23,7 +23,29 @@ test("evaluate() passes button with clickable area of exactly 44x44 pixels", asy
...
+      type="checkbox"
+      box={{ device, x: 8, y: 8, width: 13, height: 13 }}
+    ></input>
+  );
+
+  const document = h.document([target]);
+
+  t.deepEqual(await evaluate(R111, { document, device }), [
+    passed(R111, target, {
+      1: Outcomes.IsUserAgentControlled(""),
```
Jym77, line 48:
```suggestion
      1: Outcomes.IsUserAgentControlled("", target.getBoundingBox(device).getUnsafe()),
```

Jym77 (approved):

PR #1589, opened 2024-03-18
Implement R113 - Target Size (Minimum)

rcj-siteimprove:
Adds the new rule for checking minimum size and distance as defined here: https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html

Collision detection between a circle and a rectangle is still not done. It will either be done in this PR or in a follow up.


packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -0,0 +1,145 @@
...
+function hasSufficientSpacing(
+  document: Document,
+  device: Device,
+): Predicate<Element> {
+  return (target) => {
+    // The target might have been cached while computing the spacing for another target
+    const cachedTarget = spacingCache
+      .get(document, Cache.empty)
+      .get(device, Cache.empty)
+      .get(target);
```
Jym77, line 98: We need to do all the computation inside the `ifMissing` bit of the last `.get` to ensure the result is indeed cached.

packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    if (cachedTarget.isSome()) {
+      return cachedTarget.get();
+    }
+
+    // Existence of a bounding box is guaranteed by applicability
+    const box = target.getBoundingBox(device).getUnsafe();
+
+    for (const otherTarget of targetsOfPointerEvents(document, device)) {
```
Jym77, line 107: This seems costly (quadratic number of comparisons, most of them are clearly useless because the targets are very far away).
We'll probably need at some point to maybe gather the targets in a different structure to speed up the process (probably taking inspiration in Computational geometry algorithms (maybe Voronoi diagrams?))
E.g., sorting targets by x-position (or y) of the center and stopping comparison as soon as they are more than 12px apart should speed up circle intersections.

packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -0,0 +1,145 @@
...
+      if (target === otherTarget) {
+        continue;
+      }
+
+      // Existence of a bounding box is guaranteed by applicability
+      const other = otherTarget.getBoundingBox(device).getUnsafe();
+
+      // If the box of the target doesn't intersect with the circumscribed square of the other target, we know they are far enough apart
```
Jym77, line 115: Won't that leave a possibility that the circle of the target still intersects the box of the other?

packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -0,0 +1,145 @@
...
+        )
+      ) {
+        continue;
+      }
+
+      // TODO: Check if the 24px diameter circle of the target intersect with the bounding box of the other target
+
+      // If the other target is undersized, the 24px diameter circle of the target must not intersect with the 24px diameter circle of the other target
+      if (
+        (other.width < 24 || other.height < 24) &&
```
Jym77, line 128: We could get a small optimization by storing the list of undersized targets.

packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -0,0 +1,145 @@
...
+        (box.center.x - other.center.x) ** 2 +
+          (box.center.y - other.center.y) ** 2 <=
+          24 ** 2
```
Jym77, line 131: We can also get a small optimization by pre-computing that once and for all.

packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -0,0 +1,145 @@
...
+        // Existence of a bounding box is guaranteed by applicability
+        const box = target.getBoundingBox(device).getUnsafe();
+        const name = WithName.getName(target, device).getOr("");
+        return {
+          1: expectation(
+            isUserAgentControlled()(target),
+            () => Outcomes.IsUserAgentControlled(name, box),
+            hasSufficientSizeOrSpacing(document, device)(target)
+              ? () => Outcomes.HasSufficientSizeOrSpacing(name, box)
+              : () => Outcomes.HasInsufficientSizeAndSpacing(name, box),
```
Jym77, line 41:
```suggestion
            expectation(hasSufficientSizeOrSpacing(document, device)(target),
              () => Outcomes.HasSufficientSizeOrSpacing(name, box),
              () => Outcomes.HasInsufficientSizeAndSpacing(name, box))
```
I think we should also split the test further (size first then spacing) which would make different outcomes for both cases.

Jym77 (approved):
Looks good. Some possible optimisations, but let's not block this now.

rcj-siteimprove:
It was agreed offline that it's okay to merge this, make a release and address the comments on a follow up release as soon as possible. This is in order to unblock dependent work.

PR #1592, opened 2024-04-02
R113 follow up

rcj-siteimprove:
Follows up on #1589

We send more granular information about how the rule succeeded in the extended diagnostics and the spacing computation should now be correctly implemented.

The spacing still has a quadratic runtime, but I'm working on finding a solution using some kind of search tree. I might add it to this PR or in another follow up.

The image shows how the distance to the corners are calculated:
image


packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -50,96 +58,133 @@ export default Rule.Atomic.of<Page, Element>({
...
 
-    for (const otherTarget of targetsOfPointerEvents(document, device)) {
-      if (target === otherTarget) {
+    const undersizedTargets = undersizedCache
+      .get(document, Cache.empty)
+      .get(device, () =>
+        targetsOfPointerEvents(document, device).reject(
+          hasSufficientSize(24, device),
+        ),
+      );
```
Jym77, line 118: **Question:** I somehow feel that this "belongs" to inside the `evaluate` function of the rule, without being fully able to pinpoint a clear reason. Maybe this is because it reuses the targets and redoes the same test that part of the expectation does.
We could start the `evaluate` function by gathering all targets and then splitting them in two depending on size. OTOH, `hasSufficientSize` seems simple enough that re-computing it is simpler than a look-up (for deciding if the target has indeed sufficient size). For the undersized targets, they are used a lot and should indeed be stored.

packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -50,96 +58,133 @@ export default Rule.Atomic.of<Page, Element>({
...
+  const dx = Math.abs(cx - center.x);
+  const dy = Math.abs(cy - center.y);
+
+  if (dx > halfWidth + r || dy > halfHeight + r) {
+    // The circle center is outside the padded rectangle
+    return false;
+  }
+
+  // The circle center is inside the padded rectangle
+  if (dx <= rect.width / 2 || dy <= rect.height / 2) {
```
Jym77, line 175:
```suggestion
  if (dx <= halfWidth || dy <= halfHeight) {
```

packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -50,96 +58,133 @@ export default Rule.Atomic.of<Page, Element>({
...
-
         return false;
       }
     }
 
     return true;
   };
 }
+
+function circleIntersectsRect(
```
Jym77, line 151: **Suggestion:** I feel this function is complex enough to have its own tests.

Also, if it's possible to have an ASCII art explanation of the geometry, that would be awesome…

packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -50,96 +58,133 @@ export default Rule.Atomic.of<Page, Element>({
...
         )
       ) {
-        continue;
+        return false;
       }
 
-      // TODO: Check if the 24px diameter circle of the target intersect with the bounding box of the other target
-
-      // If the other target is undersized, the 24px diameter circle of the target must not intersect with the 24px diameter circle of the other target
       if (
```
Jym77, line 139:
```suggestion
      // If the candidate is undersized, the 24px diameter circle of the target must not intersect with the 24px diameter circle of the candidate
      if (
```

Jym77 (changes requested):

packages/alfa-rules/src/sia-r113/rule.ts
rcj-siteimprove: This was talked about out of band. We will revisit it when we need to make updates to the applicability of the rule.

rcj-siteimprove:
!pr extract

packages/alfa-rectangle/src/rectangle.ts
```diff
@@ -196,6 +196,70 @@ export class Rectangle
     );
   }
 
+  /**
+   * @remarks
+   * Checks if the rectangle intersects a given circle.
```
Jym77, line 201:
```suggestion
   * Checks if the rectangle intersects a given circle.
   *
   * @remarks
```
(so that the "title" of the comment is what the function does)

packages/alfa-rectangle/src/rectangle.ts
```diff
@@ -196,6 +196,70 @@ export class Rectangle
...
+      return false;
+    }
+
+    // The circle center is inside the padded rectangle
+    if (dx <= halfWidth || dy <= halfHeight) {
+      // The circle lies at most a radius away from the rectangle in the x or y directions
+      return true;
+    }
+
+    // The circle center lies in one of the corners of the padded rectangle.
```
Jym77, line 245:
```suggestion
    // 3. The circle center lies in one of the corners of the padded rectangle.
```

packages/alfa-rectangle/src/rectangle.ts
```diff
@@ -196,6 +196,70 @@ export class Rectangle
...
+    // If the distance from the circle center to the closest corner of the rectangle
+    // is less than the radius of the circle, the circle intersects the rectangle.
+    return (dx - halfWidth) ** 2 + (dy - halfHeight) ** 2 <= r ** 2;
+  }
+
+  /**
+   * @remarks
+   * Computes the squared distance between the centers of two rectangles.
```
Jym77, line 253:
```suggestion
   * Computes the squared distance between the centers of two rectangles.
   *
   * @remarks
```

packages/alfa-rules/src/common/diagnostic/with-bounding-box.ts
```diff
@@ -12,35 +18,92 @@ export class WithBoundingBox extends WithName {
...
   }
 
+  public get condition(): Either<
+    WithBoundingBox.UACondition,
+    WithBoundingBox.SizeAndSpacingCondition
+  > {
+    return this._condition;
+  }
+
+  public get tooCloseNeighbors(): Sequence<Element> {
```
Jym77, line 103:
```suggestion
  public get tooCloseNeighbors(): Iterable<Element> {
```
We usually expose stuff as `Iterable` to avoid exposing implementation details (i.e.  we can switch to another kind of iterable if need be without breaking consumers).

packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -32,13 +33,31 @@ export default Rule.Atomic.of<Page, Element>({
...
+            () =>
+              expectation(
+                hasSufficientSize(24, device)(target),
+                () => Outcomes.HasSufficientSize(name, box),
+                () =>
+                  expectation(
+                    hasSufficientSpacing(document, device)(
+                      target,
+                      tooCloseNeighbors,
+                    ),
```
Jym77, line 51: **Suggestion:** We can replace `hasSufficientSpacing` by something that produce a sequence or other kind of iterable (typically `yield`ing them upon discovery) and store that value locally.

That function can be called **before** the call to `expectation` in the arrow function:
```typescript
expectation(hasSufficientSize,
() => …
() => { const tooCloseNeighbours = … ; return expectation(tooCloseNeighbours.isEmpty(), …, …) }
```

packages/alfa-rules/src/sia-r113/rule.ts
```diff
@@ -50,96 +69,120 @@ export default Rule.Atomic.of<Page, Element>({
...
+  Cache<Device, Sequence<Element>>
 >();
 
 /**
  * @remarks
- * Spacing is calculated by
- * 1. drawing a 24px diameter circle around the center of the bounding box of the target,
- * 2. checking if the circle intersects with the bounding box of any other target, or
- * 3. if the circle intersects with the 24px diameter circle of another undersized target.
+ * This generator functions yields all elements that have insufficient spacing to the target.
```
Jym77, line 144: **Chore:** move comment title before the `@remarks` tag.
(can't make suggestion on that one, not sure why 🤔)

packages/alfa-rules/src/sia-r111/rule.ts
```diff
@@ -45,12 +46,36 @@ export default Rule.Atomic.of<Page, Element>({
  * @public
  */
 export namespace Outcomes {
```
Jym77, line 48: **Note:** Since the outcomes are shared, they can be refactored in `src/common/outcomes`.
Not urgent.

Jym77 (changes requested):
Looks good. Some smallish polish.

packages/alfa-rules/src/sia-r111/rule.ts
rcj-siteimprove: I'll add it to this #1593

rcj-siteimprove:
!pr extract

Jym77 (approved):

Issue #1593, opened 2024-04-04
SIA-R111 and SIA-R113 improvements


Jym77:
- [ ] Make inline targets inapplicable. 
- [ ] Make invisible targets inapplicable.
- [ ] Make `<area>` elements inapplicable (due to the often weird shape).
- [ ] Improve UA controlled exception to keep those whose size has been modified by author.
- [ ] Consider `<label>` as part of the size of their labelled element; this is probably OK until rpevious point is done, since in most of the case the labellable element would be an `<input>` that is currently ignored. Should we merge rectangles? Probably only if they are "close enough", but how to detect that? Implicit labels should be good, for explicit ones we need to have some heuristic of "closeness".
- [ ] Optimize collision detection
- [ ] Factor duplicated `Outcome` namespaces of R111 and R113

rcj-siteimprove:
Even though inline and invisible targets should be inapplicable, they should still be regarded when calculating spacing of an applicable target. In other words an applicable target must still not be too close to e.g. a link in a text paragraph or some invisible button. Is that correctly understood @Jym77? What about the `<area>` elements, should they also be regarded when calculating spacing of an applicable target?

Jym77:
Yes, all interactive content should be considered for spacing (main use case: touchscreen with low granularity where the contact point might cover more than one element).

MaksimShakavin:
Hello @rcj-siteimprove @Jym77 , 
I created a bug https://github.com/Siteimprove/alfa/issues/1639 which is probably connected to this issue. If you think that it duplicates the topics covered in the issue of may be already fixed but not released, feel free to close it.

PR #1594, opened 2024-04-08
SIA-R111 and SIA-R113 applicability improvements

rcj-siteimprove:
Part of #1593

### DIFF
--- a/.changeset/slimy-socks-cry.md
+++ b/.changeset/slimy-socks-cry.md
@@ -2,4 +2,4 @@
 "@siteimprove/alfa-rules": patch
 ---
 
-**Fixed:** R113 now is no longer applicable to invisible targets and targets inside a paragraph
+**Fixed:** R113 is no longer applicable to invisible targets and targets inside a paragraph
--- a/.changeset/slimy-socks-cry.md
+++ b/.changeset/slimy-socks-cry.md
@@ -1,0 +1,5 @@
+---
+"@siteimprove/alfa-rules": patch
+---
+
+**Fixed:** R113 now is no longer applicable to invisible targets and targets inside a paragraph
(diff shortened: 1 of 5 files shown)


.changeset/slimy-socks-cry.md
```diff
@@ -0,0 +1,5 @@
+---
+"@siteimprove/alfa-rules": patch
+---
+
+**Fixed:** R113 is no longer applicable to invisible targets and targets inside a paragraph
```
Jym77, line 5:
```suggestion
**Fixed:** R111 and R113 are no longer applicable to invisible targets and targets inside a paragraph
```

packages/alfa-rules/src/common/applicability/targets-of-pointer-events.ts
```diff
@@ -1,38 +1,102 @@
...
-  return cache.get(document, Cache.empty).get(device, () =>
+  return allTargetsCache.get(document, Cache.empty).get(device, () =>
     getElementDescendants(document, Node.fullTree).filter(
       and(
         hasComputedStyle(
           "pointer-events",
           (keyword) => keyword.value !== "none",
           device,
         ),
-        isFocusable(device),
```
Jym77, line 32: I think we should keep the focusable test here (and keep the same predicate as used in the Applicability).

Actually also wondering whether we can turn things a bit around and start from the tabbable elements (using `taborder` from `alfa-dom`) and filter out based on role, … This would let us cache tabbable elements for R65 🤔 Probably not for this PR, though.

Jym77 (changes requested):
Changeset needs to include R111.

Jym77 (approved):

PR #1598, opened 2024-04-18
R111 and R113: Improve inline detection

rcj-siteimprove:
Part of #1593


packages/alfa-rules/src/common/applicability/targets-of-pointer-events.ts
```diff
@@ -95,3 +117,18 @@ function isTarget(device: Device): Predicate<Element> {
...
+    }
+
+    const children = element.children(Node.flatTree);
+    if (children.some(and(isText, isVisible(device)))) {
+      return true;
+    }
+
+    return children.filter(isElement).some(hasNonTargetText(device));
+  };
+}
```
Jym77, line 134: **Issue:** This may recurse deep in the tree in case of nested `<div>` that are typical of frameworks. It is probably OK-ish due to stopping at the first text, but would be better cached for performance.

Another possibility to avoid the double traversal of the tree is to handle it as part of `visit`. `visit` would build two maps: container -> targets, and container -> non-target. Then the applicability returns all values of the first map (targets) whose key is associated with an empty list in the second.

Jym77 (changes requested):
Small performance improvement.

Jym77:
Can you try and update the definition of "rendered on a line" in Sanshikan to match the code?

Jym77 (approved):

PR #1603, opened 2024-04-25
Support `layer` and `supports` condition in `@import` CSS rules

### LINKED ISSUE #945: Investigate how to support Cascade layers (`@layer`)
https://www.w3.org/TR/css-cascade-5/#layering
https://www.bram.us/2021/09/15/the-future-of-css-cascade-layers-css-at-layer/

- [x] Add `@layer` in `alfa-dom`
- [x] Support them during cascade
- [x] add parsing in `Native`
- [x] add layers syntax to `@import` rules

---
- [x] Follow up on CST-890
- [x] Follow up on II-6406

Jym77:
Resolves #945

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "minimist": "^1.2.6",
     "package-dependency-graph": "^1.14.4",
     "prettier": "^3.0.0",
-    "typescript": "^5.2.2"
+    "typescript": "^5.4.5"
   },
   "packageManager": "yarn@4.1.1"
 }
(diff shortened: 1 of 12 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
     "minimist": "^1.2.6",
     "package-dependency-graph": "^1.14.4",
     "prettier": "^3.0.0",
-    "typescript": "^5.2.2"
+    "typescript": "^5.4.5"
```
Jym77, line 39: TS update is needed for the latest `CSSImportRule` interface, as `supportsText` is fairly recent.

rcj-siteimprove (approved):

PR #1605, opened 2024-04-26
Fix serialization of boxes

rcj-siteimprove:
I discovered that the options containing the device used to serialize the boxes of elements where not passed down in all cases when recursively serializing descendants, causing the box to only be serialized on the top elements. This PR fixes that.

### DIFF
--- a/packages/alfa-dom/test/node.spec.tsx
+++ b/packages/alfa-dom/test/node.spec.tsx
@@ -175,13 +175,10 @@ test(`Node.clone() correctly replaces elements based on predicate`, (t) => {
 test(`#toJSON() serializes boxes of all descendants when device is passed in`, (t) => {
   const device = Device.standard();
 
-  const box1 = { device, x: 8, y: 8, width: 100, height: 100 };
-  const box2 = { device, x: 16, y: 16, width: 50, height: 50 };
-
   const doc = h.document([
-    <div box={box1}>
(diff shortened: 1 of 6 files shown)


rcj-siteimprove:
!pr extract

packages/alfa-dom/test/node.spec.tsx
```diff
@@ -167,3 +168,35 @@ test(`Node.clone() correctly replaces elements based on predicate`, (t) => {
     children: [...newElements.map((e) => e.toJSON()), bar.toJSON()],
   });
 });
+
+test(`#toJSON() serializes boxes of all descendants when device is passed in`, (t) => {
```
Jym77, line 172: **Suggestion:** It seems that DOM children were already correctly handled 🤔 (was this example failing?) This PR is fixing the recursion for Shadow DOM and content document, we should add tests for these instead (also with `<slot>`).

Jym77 (changes requested):

packages/alfa-dom/test/node.spec.tsx
rcj-siteimprove: 1. I don't think they were. I think passing the `options` to `...super.toJSON(option)` on line 302 in element.ts is was what causes the option to be passed to the children. When I remove it again it also fails this test.
2. I agree there should also be tests for shadow DOM and content and `<slot>`

PR #1610, opened 2024-05-07
Depgraphs

### LINKED ISSUE #1342: Quality of Life improvements

Jym77:
Resolves #1342

Auto-generate and curate internal dependency graph for each package.
The results are a bit suboptimal, but we won't get anything better with auto-generated graphs…

There is currently no possibility to generate graphs one at a time, and no automatic test that graphs are up-to-date (they do take some time to run). So, we need to manually call the script when needed.


Jym77:
!pr extract

rcj-siteimprove (approved):
Nice work 💪🏼

PR #1611, opened 2024-05-08
Optimize accessible name calculation

rcj-siteimprove:
This PR fixes two optimization problems in the call hierarchy of `Name.fromDescendants`:

1. `Name.fromSteps` would potentially call the same functions two times unnecessarily due to an `.orElse`. This was very bad because the collection of functions passed in would contain up to two functions which would call `fromDescendants` recursively making the traversal grow extremely fast for some certain inputs.
    - Fixing this brought the number of recursive calls for some test input down from ~8.000.000 to ~3.000 and the run time from ~50.000 ms to ~200 ms
3. `Name.fromElement` would potentially recurse into `Name.fromDescendants` two times with identical arguments
    - Fixing this additionally brought the number of recursive calls for the same test input from ~3.000 down to ~100 and the run time to ~50 ms.

~~In the process of troubleshooting, I in-lined and thus eliminated the `Name.fromSteps` call in `Name.fromElement`. Despite making `fromElement` more complex, I decided to keep this in-lined because I think passing functions that is part of a recursive loop as arguments into another function which then just invokes them anyway, makes the flow very hard to follow and makes these kinds inefficiencies hard to spot.~~

~~However `Name.fromSteps` is called from other places, so it has nonetheless been optimized with a cache to avoid unnecessary re-computation. We could consider in-lining this the few other places it's used using the same approach and eliminate it completely.~~

I have taken care not to change functionality and the test suite didn't detect any regressions, but we might still want to have this regression tested.


packages/alfa-aria/src/name/name.ts
```diff
@@ -593,12 +603,19 @@ export namespace Name {
...
+
+    return Array.collectFirst(steps, (step, index) => {
+      const result = step();
+      results[index] = result;
+      return result.reject((name) => name.value === "");
+    }).orElse(() =>
+      Array.collectFirst(steps, (step, index) => {
+        const result = results[index] === undefined ? step() : results[index];
+        return result.reject((name) => name.isEmpty());
+      }),
```
Jym77, line 618:
```suggestion
      Array.collectFirst(steps, (step, index) => (results[index] ?? step()).reject((name) => name.isEmpty());
      }),
```

Jym77 (changes requested):
Small improvement.

Jym77 (dismissed):

Jym77 (approved):

Jym77 (approved):

PR #1618, opened 2024-05-29
Add verbosity option for JSON serialization of tree nodes

rcj-siteimprove:
We introduce a verbosity option for JSON serialization of tree nodes in general and in particular for DOM tree nodes. The purpose is to avoid redundant serialization of the same nodes over and over again when they are referenced multiple times. This is for instance the case in the result of an audit where the same DOM node might be the target of many different rules. We enable this by generating a random UUID on construction of each DOM node, which we call a `serializationId`.

Serializing with `Minimal` or `Low` verbosity will then only output the `serializationId` and the type. 

Serializing with `Medium` verbosity or not supplying a verbosity will result in an output containing all properties except the `serializationId` - the same as the output before verbosity was introduced. This makes the change backwards compatible as doing nothing will get you the same result as before.

Serializing with `High` verbosity will result in an output with all properties _and_ the `serializationId`. This option is meant to be used for, and is necessary for, establishing the map from ID to fully serialized DOM node. It's the callers responsibility to do this and to look up the nodes by ID when the full object needs to be reconstructed from the `serializationId`.

Additionally, we now depend on `crypto` to be present globally, meaning that we are no longer compatible with Node 18.


rcj-siteimprove:
!pr extract

rcj-siteimprove:
@Jym77, apparently `crypto` is not available as a global variable in Node 18, only after 19 making the tests fail for that version and not the others. It's possible to run the node 18 tests with the flag `--experimental-global-webcrypto`, but I'm not sure if that is something we want to do. Otherwise we need to import it everywhere we use it. Not sure if that will break something else. What do you think we should do?

Jym77:
> @Jym77, apparently `crypto` is not available as a global variable in Node 18, only after 19 making the tests fail for that version and not the others. ~It's possible to run the node 18 tests with the flag `--experimental-global-webcrypto`, but I'm not sure if that is something we want to do~. Otherwise we need to import it everywhere we use it. Not sure if that will break something else. What do you think we should do?
> 
> Edit: the flag mentioned doesn't seem to work anyway, so it seems we need to import crypto.

18 is already in LTS mode: https://nodejs.org/en/about/previous-releases

Maybe we can just move on and switch to supporting 20+ (update the actions to use 20, 22 instead of 18, 20). This does require a breaking changeset, though…

.changeset/orange-hairs-fold.md
```diff
@@ -0,0 +1,9 @@
+---
+"@siteimprove/alfa-rules": minor
+"@siteimprove/alfa-json": minor
+"@siteimprove/alfa-tree": minor
+"@siteimprove/alfa-act": minor
+"@siteimprove/alfa-dom": minor
+---
+
+**Added:** Verbosity option has been added to JSON serialization of alfa-tree `Node` and to `Outcome` and `Group`.
```
Jym77, line 9: We need the symmetrical changeset that `Node` builders now accept an optional `serializationID`.

packages/alfa-dom/test/h.spec.ts
```diff
@@ -10,16 +10,19 @@ import { Element } from "../src/node/element";
 import { Type } from "../src/node/type";
```
Jym77, line 11: We should add tests that:
* `serializationID` is correctly propagated if provided.
* `serializationID` is created if not provided (just test presence, of course).

packages/alfa-dom/test/node.spec.tsx
```diff
@@ -303,3 +305,239 @@ test(`#toJSON() serializes box of descendant inside content`, (t) => {
...
+        undefined,
+        undefined,
+        undefined,
+        elmId,
+      ),
+    ],
+    undefined,
+    undefined,
+    docId,
+  );
```
Jym77, line 334: **Suggestion** (not blocking): this can probably be refactored out of the individual cases to avoid repetition.

packages/alfa-dom/test/node.spec.tsx
```diff
@@ -303,3 +305,239 @@ test(`#toJSON() serializes box of descendant inside content`, (t) => {
...
+
+  const options = {
+    device,
+    verbosity: json.Serializable.Verbosity.Minimal,
+  } as const;
```
Jym77, line 339: **Suggestion** (not blocking): It should be possible to group `Minimal` and `Low` with a `for (const verbosity of …)` loop in the test (same for nothing/`Medium`).

packages/alfa-tree/src/tree.ts
```diff
@@ -1,3 +1,5 @@
+/// <reference lib="dom" />
+
```
Jym77, line 2:
```suggestion
```
Is this a leftover from a `console.log`?

Jym77 (changes requested):
Looks good.

packages/alfa-tree/src/tree.ts
rcj-siteimprove: No, it's required for `crypto`. Is there a better way to include it?

rcj-siteimprove:
!pr extract

Jym77 (approved):

PR #1622, opened 2024-06-10
Accept and pass down options in serialization of collection-like types (and `Page`  and releated)

rcj-siteimprove:
This is a follow up on #1618 where verbosity option was added for serialization of DOM nodes.

The serialization options are now propagated through `Page` and the collection-like types like `Array`, `Sequence` etc.

A test has been added to R41. It doesn't test the rule, but uses it to show case that the options are passed down through `Array` and `Group`.

### DIFF
--- a/packages/alfa-rules/test/sia-r41/rule.spec.tsx
+++ b/packages/alfa-rules/test/sia-r41/rule.spec.tsx
@@ -1,4 +1,3 @@
-import { Outcome } from "@siteimprove/alfa-act";
 import { h } from "@siteimprove/alfa-dom";
 import { test } from "@siteimprove/alfa-test";
 
@@ -6,8 +5,11 @@ import R41, { Outcomes } from "../../src/sia-r41/rule";
 
 import { Group } from "../../src/common/act/group";
 
+import { Outcome } from "@siteimprove/alfa-act";
 import { Response } from "@siteimprove/alfa-http";
+import { Serializable } from "@siteimprove/alfa-json";
 import { URL } from "@siteimprove/alfa-url";
+
 import { WithName } from "../../src/common/diagnostic";
 import { evaluate } from "../common/evaluate";
 import { oracle } from "../common/oracle";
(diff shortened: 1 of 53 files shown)


packages/alfa-web/src/page.ts
```diff
@@ -64,11 +64,11 @@ export class Page
     return this._device;
   }
 
-  public toJSON(): Page.JSON {
+  public toJSON(options?: Node.SerializationOptions): Page.JSON {
     return {
       request: this._request.toJSON(),
       response: this._response.toJSON(),
```
Jym77, line 70: Since we're touching this file for this reason, we should as well future-proof it.
```suggestion
      request: this._request.toJSON(options),
      response: this._response.toJSON(options),
```

packages/alfa-web/src/page.ts
```diff
@@ -64,11 +64,11 @@ export class Page
...
-      document: this._document.toJSON({ device: this._device }),
+      document: this._document.toJSON(options ?? { device: this._device }),
       device: this._device.toJSON(),
```
Jym77, line 72:
```suggestion
      device: this._device.toJSON(options),
```

Jym77 (changes requested):

rcj-siteimprove:
@Jym77 I expanded to scope of this to also include the collection-like types. There are still more types where the options should be passed through, but I think this is a good start.

packages/alfa-rules/test/sia-r41/rule.spec.tsx
```diff
@@ -171,3 +173,44 @@ test(`evaluate() can't tell if two links that have the same name references
     ),
   ]);
 });
+
+test(`toJSON() with minimal verbosity produces target with correct serialization ids`, async (t) => {
```
Jym77, line 177: I'm not sure this is relevant for this rule 🤔 Is it a test for serialisation of `Outcome` (we're missing these tests in `alfa-act` 🙈 )? Anyway, doesn't hurt.

Jym77 (approved):

rcj-siteimprove:
> Sorry, didn't thought it would explode like that 😖

No, worries, it was not that bad 😅

packages/alfa-rules/test/sia-r41/rule.spec.tsx
rcj-siteimprove: Yeah it's a bit awkward, but I wanted to have sort of an end-to-end and chose R41 because it show cases that the options are passed through both `Array`, `Outcome` and `Group` before reaching the `Node`.

There should also be tests of the individual units, but I thought having this would be good for now. When I get time, I can add some more tests in the individual packages.

PR #1625, opened 2024-06-18
Accept dimension divisions in calculations

Jym77:
Accept divisions of dimensions when they cancel out in a calculation, e.g. `calc(1em / 1px)` should reduce to a number.

This means that any kind of `Numeric` may use a `resolver` for lengths (and percentage). Keeping these as optional for now to keep backward compatibility and avoid the weird effect or `calc(1 + 1)` **requiring** a length resolver…


Jym77:
!pr extract

Jym77:
!pr extract

Jym77:
!pr extract

rcj-siteimprove (approved):

Issue #1621, opened 2024-06-10
Modernize Typescript configuration


Jym77:
Set up from 2017 doesn't work that well in modern development, see, e.g., [Choosing compiler options](https://www.typescriptlang.org/docs/handbook/modules/guides/choosing-compiler-options.html#im-writing-a-library).
Alfa:
* [x] Change TS `module` from `commonjs` to `node16`. This is the recommended setting for libraries. It will imply updating all internal imports to include extension.
* [x] Add `verbatimModuleSyntax` to TS config.
* [x] Change TS `target` to `es2022`. Features are supported by all browsers (except IE) since several versions ago.
* [x] Generate files in `dist` rather than `src`. Seems to be better practice now 🤔 
* [x] Switch JS generation from CJS to ESM. ESM is part of the language specification, so should be used instead.
* [x] Clean up `jsxImportSource`.
* [x] Clean up `import {h}`, `import … from "…/native"` and others using `dist`. Clean up `package.json` exporting `dist` entry points.
* [x] setup correct `exports` in `alfa-web` (`native`)
* [ ] ~Consider using correct imports from self-module rather than `../dist` in tests.~
* [ ] ~Clean up knip config ignoring a file in `alfa-cascade`, likely due to a `dist` setup?~
* [x] Update knip?
* [x] Update chalk

Alfa-integrations:
* [x] Change TS `module` from `commonjs` to `node16`. This is the recommended setting for libraries. It will imply updating all internal imports to include extension.
* [x] Add `verbatimModuleSyntax` to TS config.
* [x] Change TS `target` to `es2022`. Features are supported by all browsers (except IE) since several versions ago.
* [x] Generate files in `dist` rather than `src`. Seems to be better practice now 🤔 
* [x] Switch JS generation from CJS to ESM. ESM is part of the language specification, so should be used instead.
* [x] Clean up `jsxImportSource`.
* [x] Clean up `import {h}`, `import … from "…/native"` and others using `dist`. Clean up `package.json` exporting `dist` entry points.
* [ ] ~Consider using correct imports from self-module rather than `../dist` in tests.~
* [x] Update knip? => we need to have `alfa-cli` generates its files in another directory, which actually just means moving them to `src` and generate in `bin` where they already are.
* [x] Re-enable TS config validation in `validate-structure`.

Alfa-companion:
* [x] Change TS `module` from `commonjs` to `node16`. This is the recommended setting for libraries. It will imply updating all internal imports to include extension.
[...]

PR #1628, opened 2024-06-20
Generate output files in `dist` directory

Jym77:
Part of #1621.


config/knip.ts
```diff
@@ -11,12 +11,23 @@ const config: KnipConfig = {
       ignoreDependencies: ["prettier"],
     },
     "packages/*": { entry, project },
-    "packages/alfa-device": { entry: [...entry, "native.ts!"], project },
+    "packages/alfa-cascade": {
+      entry,
+      project,
+      ignore: ["src/ancestor-filter.ts"],
```
Jym77, line 17: Seems to be a problem with knip not detecting that something is used in a test file.
Hopefully will be clearable once we reach a more final step.

Jym77:
!pr extract

Jym77:
!pr extract

packages/alfa-act/package.json
```diff
@@ -11,11 +11,11 @@
     "directory": "packages/alfa-act"
   },
   "bugs": "https://github.com/siteimprove/alfa/issues",
-  "main": "src/index.js",
-  "types": "src/index.d.ts",
+  "main": "dist/index.js",
```
rcj-siteimprove, line 14: **Question:** Are we going to switch to `exports` at some point? It's apparently the "modern" alternative to `main`.
https://nodejs.org/api/packages.html#package-entry-points

rcj-siteimprove (dismissed):

packages/alfa-act/package.json
Jym77: Yes, probably at some point… It does restrict what external consumer can use, which is actually rather what we want :-)

rcj-siteimprove (approved):

Jym77:
!pr extract

rcj-siteimprove (approved):

PR #1637, opened 2024-06-28
Set up `verbatimModuleSyntax`

Jym77:
Part of #1621

Set up `verbatimModuleSyntax` as recommended for libraries.
This effectively required to use `import type` where possible, which in turned required to update dependency graphs 🙈 

The `type` indications were added via `typescript-eslint`, but the running time is a bit slow and the default configuration opinionated in different ways than Alfa is, so I didn't activate it here. Would be the topic for another PR…


Jym77:
!pr extract

PR #1640, opened 2024-07-01
Cleaning

Jym77:
End of #1621 for Alfa.

Mostly cleans up the spurious `/dist` in imports that were needed due to the module resolution, plus a bit of assorted cleanup.
(and fixing a typo in all changelogs 🙈)

Ideally, we'd still want to have `exports` fields in `package.json` (this will break how Dory reads the version number, we may need to provide it directly instead); which will enable us to use self-imports in tests rather than referring to `dist`, but we should probably take that at a later point.


rcj-siteimprove (approved):
🧹

PR #1643, opened 2024-07-01
Move changeset generator to CJS

Jym77:
It seems that changeset config doesn't manage to load its own ESM version and thus needs to load more CJS code.
Fortunately, it also seems that only one file is really impacted 🤞


Jym77:
!pr extract

rcj-siteimprove (dismissed):

rcj-siteimprove (approved):

PR #1648, opened 2024-07-08
Include descendants in clickable area computation

rcj-siteimprove:
A target might have an undersized bounding box while the clickable area is sufficiently sized if there is an element descendant (like an image) that is sufficiently sized.

In general we need the target size rules to rely on the clickable area and not the bounding box of a target. So we try to calculate the clickable area of an element which as the union the bounding boxes of an element and all its descendants., i.e. the smallest rectangle that contains all the bounding boxes. This can overshoot the actual clickable area if it's not rectangular.

### DIFF
--- a/packages/alfa-dom/src/node/query/inclusive-element-descendants.ts
+++ b/packages/alfa-dom/src/node/query/inclusive-element-descendants.ts
@@ -1,0 +1,23 @@
+import { Cache } from "@siteimprove/alfa-cache";
+import type { Sequence } from "@siteimprove/alfa-sequence";
+import { Node } from "../../node.js";
+import { Element } from "../element.js";
+
+const cache = Cache.empty<Element, Array<Sequence<Element>>>();
+
+/**
+ * @public
+ */
+export function getInclusiveElementDescendants(
+  node: Element,
+  options: Node.Traversal = Node.Traversal.empty,
+): Sequence<Element> {
+  const optionsMap = cache.get(node, () => []);
+  if (optionsMap[options.value] === undefined) {
+    optionsMap[options.value] = node
+      .inclusiveDescendants(options)
+      .filter(Element.isElement);
+  }
+
+  return optionsMap[options.value];
+}
--- a/packages/alfa-dom/src/node/query/inclusive-element-descendants.ts
+++ b/packages/alfa-dom/src/node/query/inclusive-element-descendants.ts
@@ -12,12 +12,5 @@ export function getInclusiveElementDescendants(
   node: Element,
   options: Node.Traversal = Node.Traversal.empty,
 ): Sequence<Element> {
-  const optionsMap = cache.get(node, () => []);
-  if (optionsMap[options.value] === undefined) {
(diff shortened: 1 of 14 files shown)


packages/alfa-rules/src/common/dom/get-clickable-box.ts
```diff
@@ -0,0 +1,41 @@
...
+const cache = Cache.empty<Device, Cache<Element, Option<Rectangle>>>();
+
+/**
+ * Gets the bounding box of the clickable area of an element
+ * or None if the element or one of its descendants doesn't have a bounding box.
+ *
+ * @remarks
+ * This function assumes that the element can receive pointer events, i.e. is clickable.
+ * If called on an element that is not clickable, the function will still return the
+ * the area that would be clickable if the element was could receive pointer events.
```
Jym77, line 18:
```suggestion
 * the area that would be clickable if the element could receive pointer events.
```

packages/alfa-rules/src/common/dom/get-clickable-box.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export function getClickableBox(
+  device: Device,
+  element: Element,
+): Option<Rectangle> {
+  return cache.get(device, Cache.empty).get(element, () => {
+    let boxes: Array<Rectangle> = [];
+    for (let box of element
+      .inclusiveDescendants(Node.flatTree)
+      .filter(isElement)
```
Jym77, line 30:
```suggestion
    for (let box of Query.getElementsDescendants(element)
```
(might need to re-add `element` to the list)

Jym77 (dismissed):
Looks good. Some minor polish.

packages/alfa-rules/src/common/dom/get-clickable-box.ts
rcj-siteimprove: I have added the function `Query.getInclusiveElementDescendants` for this.

packages/alfa-dom/src/node/query/inclusive-element-descendants.ts
```diff
@@ -0,0 +1,23 @@
...
+  options: Node.Traversal = Node.Traversal.empty,
+): Sequence<Element> {
+  const optionsMap = cache.get(node, () => []);
+  if (optionsMap[options.value] === undefined) {
+    optionsMap[options.value] = node
+      .inclusiveDescendants(options)
+      .filter(Element.isElement);
+  }
+
+  return optionsMap[options.value];
```
Jym77, line 22:
```suggestion
return getElementDescendants(node, options).prepend(node)
```
That way, we'll share the cache.

Jym77 (changes requested):

packages/alfa-dom/src/node/query/inclusive-element-descendants.ts
rcj-siteimprove: Nice 👍🏼

Jym77 (approved):

PR #1658, opened 2024-07-22
Export version for easier access

Jym77:
Got one more case where Alfa version was needed (to upload external page reports) and didn't want to go through one more hack… This time we have a correct solution…

It turns out it is surprisingly difficult to get your own version 🙈 
* The package.json variable `process.env.npm_package_version` is only available when called through npm, yarn, … (which makes sense), hence not when called from a browser or directly from node. This makes it hard to use for us. Additionally, it is not clear which package is picked depending on which script is run from where.
* Importing `package.json` seems like bad practice as it means all clients would have it and it risk exposing too much in some cases. Additionally, import assertions are currently messed up between TS 5 and Node 20. And loading and parsing this file is not what we want to do at runtime anyway…
* Pre-generating it require updating it **after** new versions are created.

This should solve the third point, with both a (fairly simple) generation script and a hook in the release flow to trigger it. Lifecycle hooks are very little in yarn so we cannot just rely on them but still want to avoid name collision with npm lifecycle hooks…


rcj-siteimprove (dismissed):
Looks good

Jym77:
!pr extract

rcj-siteimprove (approved):

PR #1660, opened 2024-07-24
Add z-index property

rcj-siteimprove:
The z-index property will be needed in a later PR when we start working on stacking contexts.


rcj-siteimprove:
!pr extract

packages/alfa-style/src/property/z-index.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const { either } = Parser;
+
+const parse = either(Keyword.parse("auto"), Integer.parse);
+
+/**
+ * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/z-index}
+ * @internal
+ */
+export default Longhand.of(Keyword.of("auto"), parse, (value) => value, {
```
Jym77, line 14: **Chore:** Let's make specified and computed types explicit like in the other properties.

Jym77, line 14: **question:** Do we need to resolve calculated values (for integers)? 
I think it is not technically needed (since `z-index: calc(1em / 1px)` sounds like madness), but we should have the computed type as `Integer.Canonical` which may require a call to `.resolve` anyway (using the default length resolver based on the style, same as in other properties).

Jym77 (changes requested):
Looks good. Some minor fixes.

Also, do not merge before I've fixed the publication again, to avoid mixing relevant changes in it 🙈

rcj-siteimprove:
!pr extract

packages/alfa-style/src/property/z-index.ts
```diff
@@ -0,0 +1,24 @@
...
+ * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/z-index}
+ * @internal
+ */
+export default Longhand.of<Specified, Computed>(
+  Keyword.of("auto"),
+  parse,
+  (value) => value.map((zIndex) => zIndex.resolve()),
+  {
+    inherits: false,
+  },
```
Jym77, line 23: **Nitpick**
```suggestion
  { inherits: false },
```

packages/alfa-style/src/property/z-index.ts
```diff
@@ -0,0 +1,24 @@
...
+type Computed = Keyword<"auto"> | Integer.Canonical;
+
+/**
+ * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/z-index}
+ * @internal
+ */
+export default Longhand.of<Specified, Computed>(
+  Keyword.of("auto"),
+  parse,
+  (value) => value.map((zIndex) => zIndex.resolve()),
```
Jym77, line 20: **Note:** This should be good, but it might break in the weird case of a length division.
```suggestion
  (value, style) => value.map((zIndex) => zIndex.resolve(Resolver.length(style))),
```
would be the safe bet.

Jym77 (dismissed):

packages/alfa-style/src/property/z-index.ts
rcj-siteimprove: Just out of curiosity, is a length division when there units involved? Units should be forbidden in z-index values and that seems to be true even if they cancel out; I tested `{ z-index: "calc(1px / 1px)"; }` on https://jigsaw.w3.org/css-validator/validator and it gave a parsing error and in chrome the computed value becomes `auto`.

rcj-siteimprove: It also gave a parsing error when the syntax was valid 😅; `.foo { z-index: calc(1px / 1px); }`

rcj-siteimprove: We probably actually need a resolver that _doesn't_ cancel units in this case to conform to the spec 🙈 But it might be too much of a corner case 🤷🏼

Jym77: Based on [Type checking](https://drafts.csswg.org/css-values/#calc-type-checking) and [Operations on types](https://drafts.css-houdini.org/css-typed-om-1/#cssnumericvalue-multiply-two-types), I think that these do technically "match" the `<number>`/`<integer>` types. And since, according to [Math expressions](https://drafts.csswg.org/css-values/#math):

> A [math function](https://drafts.csswg.org/css-values/#math-function) represents a numeric value, one of (…)[<number>](https://drafts.csswg.org/css-values/#number-value), [<integer>](https://drafts.csswg.org/css-values/#integer-value) (…), and can be used wherever such a value would be valid.

I understand the specs as not forbidding it…
I totally get why browsers think it's madness and refuse this, however.

Jym77 (approved):

PR #1669, opened 2024-08-20
Serialize elements in `WithBadElements`

rcj-siteimprove:
When `WithBadElements` diagnostic is serialized, we now serialize the erroneous elements the same way we do most other places in stead of returning their xpaths. This give the clients control over how the elements should be serialized (full object or only serializationId).

### DIFF
--- a/.changeset/hungry-eyes-care.md
+++ b/.changeset/hungry-eyes-care.md
@@ -1,0 +1,5 @@
+---
+"@siteimprove/alfa-rules": minor
+---
+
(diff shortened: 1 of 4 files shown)


mvy-siteimprove (approved):

rcj-siteimprove:
@Jym77 due to time constraints we agreed on bypassing the protections and merge this. @mvy-siteimprove has reviewed. Feel free to add comments that should be addressed after the fact.

PR #1675, opened 2024-08-29
Add properties `container-type`, `mix-blend-mode` and `isolation`

rcj-siteimprove:
These properties will be needed for stacking context calculations.

### DIFF
--- a/packages/alfa-style/src/longhands.ts
+++ b/packages/alfa-style/src/longhands.ts
@@ -83,6 +83,7 @@ import MarginRight from "./property/margin-right.js";
 import MarginTop from "./property/margin-top.js";
 import MinHeight from "./property/min-height.js";
 import MinWidth from "./property/min-width.js";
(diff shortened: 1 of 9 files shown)


rcj-siteimprove:
!pr extract

.changeset/great-dodos-cheer.md
```diff
@@ -0,0 +1,5 @@
+---
+"@siteimprove/alfa-style": minor
+---
+
+**Added:** Style properties `container-type`, `mix-blend-mode` and `isolation` were added.
```
Jym77, line 5:
```suggestion
**Added:** Style properties `container-type`, `mix-blend-mode` and `isolation` are now supported.
```

Jym77 (commented):
Looks good.
If we're not in a rush, I'd rather not merge it now so we can keep the 0.89 mess with the npm publication isolated and only put actual changes in 0.90 😅

Jym77 (approved):

PR #1679, opened 2024-09-04
Add support for `contain` property

rcj-siteimprove:
This property will be needed for stacking context calculations.

### DIFF
--- a/packages/alfa-css/src/value/contain.ts
+++ b/packages/alfa-css/src/value/contain.ts
@@ -215,8 +215,8 @@ export namespace ContainFlags {
     return Result.of([
       input,
       ContainFlags.of(
-        size !== undefined && size.value === "size",
-        size !== undefined && size.value === "inline-size",
+        size?.value === "size",
+        size?.value === "inline-size",
         layout !== undefined,
         style !== undefined,
         paint !== undefined,
--- a/packages/alfa-css/src/value/contain.ts
+++ b/packages/alfa-css/src/value/contain.ts
@@ -1,0 +1,237 @@
+import type { Hash } from "@siteimprove/alfa-hash";
+import { Err, Result } from "@siteimprove/alfa-result";
+import { Parser } from "@siteimprove/alfa-parser";
+
+import { Keyword } from "./keyword.js";
+import type { Resolvable } from "./resolvable.js";
+import { Value } from "./value.js";
+
+import { Token, type Parser as CSSParser } from "../syntax/index.js";
+import { Slice } from "@siteimprove/alfa-slice";
+
+const { either } = Parser;
+
+/**
+ * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/contain#formal_syntax}
+ *
+ * @public
+ */
+export type Contain =
+  | Keyword<"none">
+  | Keyword<"strict">
+  | Keyword<"content">
+  | ContainFlags;
+
+/**
+ * @public
+ */
+export class ContainFlags
+  extends Value<"contain-flags", false>
+  implements Resolvable<ContainFlags, never>
+{
+  public static of(
+    size: boolean,
+    inlineSize: boolean,
+    layout: boolean,
+    style: boolean,
+    paint: boolean,
+  ) {
+    return new ContainFlags(size, inlineSize, layout, style, paint);
+  }
+
+  private readonly _size: boolean;
+  private readonly _inlineSize: boolean;
+  private readonly _layout: boolean;
+  private readonly _style: boolean;
+  private readonly _paint: boolean;
+
+  private constructor(
+    size: boolean,
+    inlineSize: boolean,
+    layout: boolean,
+    style: boolean,
+    paint: boolean,
+  ) {
+    super("contain-flags", false);
+    this._size = size;
+    this._inlineSize = inlineSize;
+    this._layout = layout;
+    this._style = style;
+    this._paint = paint;
+  }
+
+  public get size(): boolean {
+    return this._size;
+  }
+
+  public get inlineSize(): boolean {
+    return this._inlineSize;
+  }
+
+  public get layout(): boolean {
+    return this._layout;
+  }
+
+  public get style(): boolean {
+    return this._style;
+  }
+
+  public get paint(): boolean {
+    return this._paint;
+  }
+
+  public resolve(): ContainFlags {
+    return this;
+  }
+
+  public equals(value: unknown): value is this {
+    return (
+      value instanceof ContainFlags &&
+      value._size === this._size &&
+      value._inlineSize === this._inlineSize &&
+      value._layout === this._layout &&
+      value._style === this._style &&
+      value._paint === this._paint
+    );
+  }
+
+  public hash(hash: Hash): void {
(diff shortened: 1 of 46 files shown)


rcj-siteimprove:
!pr extract

packages/alfa-style/src/property/contain.ts
```diff
@@ -0,0 +1,120 @@
...
+
+const { either, map } = Parser;
+
+type Specified = Specified.Single | Specified.Multiple;
+type Computed = Specified;
+
+namespace Specified {
+  export type Single = Tuple<
+    [Keyword<"none"> | Keyword<"strict"> | Keyword<"content">]
+  >;
```
Jym77, line 21: **Suggestion:** Does that need to be wrapped in a `Tuple`? Can't it be `Keyword<"none"> | Keyword<"strict"> | Keyword<"content">`?

packages/alfa-style/src/property/contain.ts
```diff
@@ -0,0 +1,120 @@
...
+      Keyword<"style">,
+      Keyword<"paint">,
+    ]
+  >;
+}
+
+const parseSingle: CSSParser<Specified.Single> = map(
+  Keyword.parse("none", "strict", "content"),
+  (keyword) => Tuple.of(keyword),
+);
```
Jym77, line 35:
```suggestion
);

```

packages/alfa-style/src/property/contain.ts
```diff
@@ -0,0 +1,120 @@
...
+    return Err.of("Expected at least one keyword");
+  }
+
+  return Result.of([
+    input,
+    Tuple.of(
+      ...[size, layout, style, paint].filter(
+        (keyword) => keyword !== undefined,
+      ),
+    ) as Specified.Multiple,
```
Jym77, line 100: **Issue:** This actually seems unsafe 🤔 `Tuple` have fixed length, so the type always expects 4 arguments but this may have as little as one after the filtering. Shouldn't this instead default to the default values and thus always create the correct tuple?
`Tuple.of([size ?? sizeDefault, …])`

See, e.g., `Shadow.parse` in `alfa-css`.

packages/alfa-style/src/property/contain.ts
```diff
@@ -0,0 +1,120 @@
...
+);
+const parseMultiple: CSSParser<Specified.Multiple> = (input) => {
+  let size: Keyword<"size"> | Keyword<"inline-size"> | undefined;
+  let layout: Keyword<"layout"> | undefined;
+  let style: Keyword<"style"> | undefined;
+  let paint: Keyword<"paint"> | undefined;
+
+  while (true) {
+    for ([input] of Token.parseWhitespace(input)) {
+    }
```
Jym77, line 44:
```suggestion
    for (const [remainder] of Token.parseWhitespace(input)) {
      input = remainder;
    }
```
I think that needs to update the input 🤔

packages/alfa-style/src/property/contain.ts
```diff
@@ -0,0 +1,120 @@
...
+  >;
+  export type Multiple = Tuple<
+    [
+      Keyword<"size"> | Keyword<"inline-size">,
+      Keyword<"layout">,
+      Keyword<"style">,
+      Keyword<"paint">,
+    ]
+  >;
+}
```
Jym77, line 30: **Suggestion:** Looking at the property definition, it seems that the 4 keywords are more like flags than real values.
This means that the typical questions we'll need to ask about it will be "does this `contain` has `size`"? Storing them as a list of keywords will make that rather cumbersome (`style.computed("contain").includes(Keyword("size"))` is a bit hard to read).

It feels like it would be better to have at least the computed value as 4 boolean, either as a tuple, or as an object, or maybe it makes sense to create a type in `alfa-css` for this property. So we'd internally store 4 booleans but could have the correct stringification to respect the specs.
Then querying it would be `style.comnputed("contain").size` which feels more natural. This should also simplify testing the the `toJSON` would always return the 4 booleans, so order won't matter.

Jym77 (changes requested):

packages/alfa-style/src/property/contain.ts
rcj-siteimprove: It is being updated because of `for([input] ...`, but I agree that's probably a little too hidden/obscure. I'll change it 👍🏼

rcj-siteimprove:
!pr extract

packages/alfa-css/src/value/contain.ts
```diff
@@ -0,0 +1,237 @@
...
+    paint: boolean,
+  ) {
+    return new ContainFlags(size, inlineSize, layout, style, paint);
+  }
+
+  private readonly _size: boolean;
+  private readonly _inlineSize: boolean;
+  private readonly _layout: boolean;
+  private readonly _style: boolean;
+  private readonly _paint: boolean;
```
Jym77, line 46: **Question:** Would it make sense to store that as an `alfa-flag` 🤔 ? (see also how the DOM `Node` names its flags)  We'd still need a wrapper class as it needs to extends `Value`, so maybe that wouldn't help at all. But that could help in providing the query and build helpers.

packages/alfa-css/src/value/contain.ts
```diff
@@ -0,0 +1,237 @@
...
+      style === undefined &&
+      paint === undefined
+    ) {
+      return Err.of("Expected at least one keyword");
+    }
+
+    return Result.of([
+      input,
+      ContainFlags.of(
+        size !== undefined && size.value === "size",
```
Jym77, line 218:
```suggestion
        size?.value === "size",
```

packages/alfa-css/src/value/contain.ts
```diff
@@ -0,0 +1,237 @@
...
+        size !== undefined && size.value === "inline-size",
```
Jym77, line 219:
```suggestion
        size?.value === "inline-size",
```

packages/alfa-css/src/value/contain.ts
```diff
@@ -0,0 +1,237 @@
...
+        layout !== undefined,
+        style !== undefined,
+        paint !== undefined,
```
Jym77, line 222: **Note:** we could store these directly as boolean and use `layout ?? false` here. I'm not sure that really improve things since it means we wouldn't be able to directly do `[input, layout] = result.get()`. Or maybe by wrapping the parser in `const result = map(Keyword.parse("layout"), () => true)(input)`. But that's not really better…

Jym77 (dismissed):
Looks good. A few nitpicky suggestions. Not sure they really improve readability, so feel free to ignore them 😄

packages/alfa-css/src/value/contain.ts
rcj-siteimprove: It could make sense, but I think I'll leave it as it is for now. 😄

packages/alfa-css/src/value/contain.ts
rcj-siteimprove: I'll leave this as is for now 🙂

Jym77 (approved):

PR #1683, opened 2024-09-17
Improve visibility detection of `<option>`

### LINKED ISSUE #1473: Investigate how colors are set and grabbed for `<select>` and `<option>`
It at least seems we try to look for color contrast of `<option>`, including the non-selected ones.

- [x] Follow-up on CST-1378
- [x] Follow-up on CST-2837

Jym77:
Resolves #1473


rcj-siteimprove (approved):
Looks good

PR #1687, opened 2024-09-25
Add support for `scale` property

rcj-siteimprove:
This property will be needed for stacking context calculations.

https://developer.mozilla.org/en-US/docs/Web/CSS/scale


rcj-siteimprove:
@Jym77 Can you help me out with how to get the percentages to work? I based the property implementation on the similar transform function implementation since it's also supposed to work with percentages, I think. I tried using `LengthPercentage`, but couldn't get it to work 🙂

rcj-siteimprove:
Superseded by https://github.com/Siteimprove/alfa/pull/1688

PR #1688, opened 2024-09-26
Support for `scale` and `translate` CSS properties

rcj-siteimprove:
CSS properties `scale` and `translate` have been added. The existing types used when parsing the corresponding transform functions have been reused and additional parsers have been added. In addition the `scale()` transformation function and it's relatives now correctly accept percentages. Two additional functions `scale3d()` and `scaleZ()` have also been added.


rcj-siteimprove:
!pr extract

rcj-siteimprove:
!pr extract

packages/alfa-css/src/value/transform/scale.ts
```diff
@@ -4,46 +4,89 @@ import { Parser } from "@siteimprove/alfa-parser";
...
+    X extends Number.Canonical | Percentage.Canonical,
+    Y extends Number.Canonical | Percentage.Canonical,
+    Z extends Number.Canonical | Percentage.Canonical,
+  >(x: X, y: Y, z?: Z): Scale<X, Y, Z>;
+
+  public static of<
+    X extends Number | Percentage<"percentage">,
+    Y extends Number | Percentage<"percentage">,
+    Z extends Number | Percentage<"percentage">,
+  >(x: X, y: Y, z?: Z): Scale<ToCanonical<X>, ToCanonical<Y>, ToCanonical<Z>>;
```
Jym77, line 52:
```suggestion
  public static of<
    X extends Number | Percentage<"percentage">,
    Y extends Number | Percentage<"percentage">,
  >(x: X, y: Y): Scale<ToCanonical<X>, ToCanonical<Y>, never>;
```
Seems like that overload should only have 2 parameters, since we already have an overload for 3D scale 🤔 
(i.e., with an optional `z`, this already covers the 3-values overload which is not needed)

packages/alfa-css/src/value/transform/scale.ts
```diff
@@ -4,46 +4,89 @@ import { Parser } from "@siteimprove/alfa-parser";
...
+    Z extends Number | Percentage<"percentage">,
+  >(x: X, y: Y, z?: Z) {
+    return new Scale(x.resolve(), y.resolve(), z?.resolve());
   }
 
-  private readonly _x: Number.Canonical;
-  private readonly _y: Number.Canonical;
+  private readonly _x: X;
+  private readonly _y: Y;
+  private readonly _z?: Z;
```
Jym77, line 64: **Suggestion:** Should that be stored as an `Option` instead? Or at least returned so by the getter?

packages/alfa-css/src/value/transform/scale.ts
```diff
@@ -64,10 +107,15 @@ export class Scale
       ...super.toJSON(),
       x: this._x.toJSON(),
       y: this._y.toJSON(),
+      z: this._z !== undefined ? this._z.toJSON() : null,
```
Jym77, line 110:
```suggestion
      z: this.?_z?.toJSON() ?? null,
```

Jym77, line 110: **Note:** We can also get rid of `z` totally by doing a trick like
```
...(this._z !==  undefined ? { z: this._z.toJSON() } : {})
```
I'm never totally sure what is the best solution 😅 The trick makes for more compact JSON (and avoid the need to update a lot of code relying on it, so better backwards compatibility), but is also quite ugly to write and read/understand…

packages/alfa-style/test/property/scale.spec.tsx
```diff
@@ -0,0 +1,142 @@
...
+    source: null,
+  });
+});
+
+test("property is not inherited", (t) => {
+  const element = <div>Hello</div>;
+
+  h.document(
+    [<div class="container">{element}</div>],
+    [h.sheet([h.rule.style(".container", [h.declaration("scale", "2")])])],
```
Jym77, line 29: **(optional)**
```suggestion
    [h.sheet([h.rule.style(".container", { "scale", "2" })])],
```
When the declaration is not needed, we can use a name/value record which is arguably easier to read for humans.

Jym77 (dismissed):
Well done. Looks very smooth in the rest 😄 
A few nitpicky suggestions here and there.

Jym77 (approved):

PR #1689, opened 2024-09-27
Support used values

Jym77:
Part of #955 

While we cannot fully support used values for stuff that depends on layout (e.g., `width: auto`) (nor really need to yet), we can at least return no used value when the property does not apply to the element.

This adds the support for used value, as well as computes it correctly for `overflow` and `text-overflow`. We may have other properties in the same case, but we'll handle that when needed.


rcj-siteimprove (approved):
Looks good 👍🏼

PR #1690, opened 2024-09-30
Document / check list how to add a new style property

Jym77:


rcj-siteimprove (dismissed):
Nice with the new guide! 😃

PR #1692, opened 2024-09-30
Vitest

rcj-siteimprove:
This PR introduces [Vitest ](https://vitest.dev/) to Alfa.

- The command to run all tests is the same as before: `yarn test`.
- To test a specific package, execute `yarn test --project @siteimprove/alfa-xxx`. 

Vitest can run the `.ts(x)` files directly so there is no longer a need to emit JS for the test files, therefore `"outDir": "."` has been replaced by `"noEmit": true` in all `test/tsconfig.json` files.

This PR is meant to lay the foundation for future work in getting code coverage and mutation coverage into Alfa.

The original `alfa-test` package has been renamed to `alfa-test-deprecated` and `alfa-test` is now used to wrap vitest. Consumers that were using `alfa-test` can switch to using `alfa-test-deprecated` or try using the new wrapper which will require them to set up vitest.

### DIFF
--- a/.changeset/orange-points-rule.md
+++ b/.changeset/orange-points-rule.md
@@ -1,0 +1,8 @@
+---
+"@siteimprove/alfa-test-deprecated": minor
+"@siteimprove/alfa-test": minor
+---
+
+**Breaking:** The `alfa-test` package is now called `alfa-test-deprecated` and a new `alfa-test` package was added.
+
+Consumers can either migrate by replacing references to `alfa-test` with `alfa-test-deprecated` or they can attempt to use `alfa-test` which builds on [Vitest](https://vitest.dev).


config/vitest.config.ts
```diff
@@ -0,0 +1,7 @@
+import { defineConfig } from "vitest/config";
+
+export default defineConfig({
+  test: {
+    include: ["**/*.spec.js"],
```
Jym77, line 5:
```suggestion
    include: ["packages/alfa-*/test/**/*.spec.js"],
```
I prefer being less "catch all" when possible 😃

Jym77, line 5: **Question:** Is it possible to run vitest directly on the TS files? (I've heard yes) Not sure what is best (OTOH, running ts-node and vitest on `.ts(x)` would reduce build time, keeping it only for the publication flow).

tsconfig.json
```diff
@@ -11,7 +11,8 @@
     "sourceMap": true,
     "strict": true,
     "types": [],
-    "verbatimModuleSyntax": true
+    "verbatimModuleSyntax": true,
+    "skipLibCheck": true
```
Jym77, line 15: **Nitpick:** I'm always anxious about using `skipLibCheck`. It essentially means some lib has bad typing and is often a symptom of a forgotten update, or a bug to report upstream.

Jym77: Is it due to https://github.com/vitejs/vite/issues/9813 ?

config/vitest.config.ts
rcj-siteimprove: This gives some problems with the `--dir packages/alfa-xxx` command line option. It seems the include path is relative to the directory being passed in the option.

I have looked at the [workspace](https://vitest.dev/guide/workspace.html) feature. It can limits us to the `alfa-*` packages and enables testing individual packages. It will also prefix the package names in the tests which I find also a nice little addition.

The only thing annoying is that it takes the package name from its `package.json` meaning we have to write `yarn test --project (at)siteimprove/alfa-xxx` 😕

rcj-siteimprove: Yes, that's correct. I initially couldn't get it to work with the `.tsx` files, that's why I decided to run the emitted JS, but for some reason now it just works on both `.ts` and `.tsx`. 😅

We could remove the tests the build altogether as I don't think we need them anymore and we would be able to have a single `.tsconfig` in every package again. Or do you think we still need the emitted JS from the test files?

Jym77: Yeah, we might have to do `yarn workspace @siteimprove/alfa-foo test` or something similar indeed. But maybe we can salvage the simpler command (+ shell completion by path, which is the default behaviour and thus works out of the box!) by doing some magic around the `test` script in `package.json` (we already did so in `scripts/test.mjs`…)

Some magic like `test: vitest --project $(sed 's+packages+@siteimprove+)` (plus some more magic around no paramtere 🙈), or just keep a simple `scripts/test.mjs` that wraps vitest with the correct config based on its parameters?


Ultimately, and due to https://github.com/Siteimprove/alfa-integrations/pull/102#issuecomment-2441608723 we might want each individual package to have its own `test` script (and call vitest, probaly with shared config if possible), and the top-level script will be `yarn workspaces --all foreach test` (I also like the package prefixing). Then we need a wrapper to do `yarn workspace @siteimprove/alfa-foo test` nicely (and with directory based completion, unless a nice `yarn workspace` completion utility exists…)

Jym77: Yes, the idea is to remove the build step. It is needed for `src` since that's what we distribute, but we don't care about it for `test`, so we can simplify code structure a bit.
I think knip will also get less confused to not have the `.js` together with the `.ts(x)` 😅 

If we fully switch to ts-node or the like, we can also totally remove the build step in the normal flow. We'd still need it in the publish flow (and so, for the sake of sanity, keep it in the `integrate` workflow), but local runs could be faster if we don't have to rebuild.

Jym77: (might be in another PR, though, no need to make this one explode)

rcj-siteimprove: Just realized there might be some confusing naming overlap here 😅 The workspace feature I looked at is actually a vitest thing (I looks like it tries to do something similar to yarn workspaces). It might also make sense to try and use yarn workspaces, but what I'm looking at currently involves only vitest. Sorry that wasn't very clear.

I'll push some changes soon, then I hope it should be more clear what it's about.

rcj-siteimprove:
!pr extract

tsconfig.json
rcj-siteimprove: I removed the `skipLibCheck` and applied the workaround mentioned in the issue.

rcj-siteimprove:
@Jym77, do we need a changeset when it is (almost) only tooling changes? The only other change I had to do was in `alfa-dom` related to how `h` is exported.

packages/alfa-test-deprecated/src/stack.ts
```diff
@@ -0,0 +1,69 @@
...
+  const frames = error
+    .stack!.split("\n")
+    .slice(1)
+    .map((frame) => frame.trim());
+
+  for (let i = 0, n = frames.length; i < n; i++) {
+    const frame = frames[i];
+
+    // If the frame doesn't contain a location reference, skip the frame.
+    if (!/(\d+:\d+)/.test(frame)) {
```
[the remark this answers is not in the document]
rcj-siteimprove: @Jym77, this must have been pre-existing and somehow got flagged after the rename. Can we ignore it, or do we need to take action?

Jym77 (approved):

PR #1697, opened 2024-10-10
CSS variable resolution

Jym77:
Fix a case where chained CSS variables were not fully resolved, depending on the iteration ordering of the `Map`…


Jym77:
!pr extract

PR #1705, opened 2024-10-28
Rename `tree.Node#serializationId` to `tree.Node#internalId`.

Jym77:
This makes it a bit more general.

### DIFF
--- a/.changeset/fuzzy-toys-turn.md
+++ b/.changeset/fuzzy-toys-turn.md
@@ -6,4 +6,4 @@
 
 **Breaking:** `#serializationId` has been replaced with `#internalId`.
 
-The old `#serializationId` is now deprecated and acts as an alias to `#internalId`.
+The old `#serializationId` is now deprecated and acts as an alias to `#internalId`. It will be removed in a future version.
--- a/.changeset/fuzzy-toys-turn.md
+++ b/.changeset/fuzzy-toys-turn.md
@@ -1,0 +1,9 @@
+---
+"@siteimprove/alfa-rules": minor
+"@siteimprove/alfa-tree": minor
+"@siteimprove/alfa-dom": minor
+---
+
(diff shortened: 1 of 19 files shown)


Jym77:
!pr extract

Jym77:
@jfer-siteimprove @mvy-siteimprove FYI. I believe this will break the Code Checker Page Report that only receives the JSON serialization and thus need to read the correct field name.

mvy-siteimprove:
> @jfer-siteimprove @mvy-siteimprove FYI. I believe this will break the Code Checker Page Report that only receives the JSON serialization and thus need to read the correct field name.

Thanks for the heads up @Jym77. Good that the `serializationId` is still available as an alias. We can safely deploy this, update our usage and then delete the alias.

Jym77:
> > @jfer-siteimprove @mvy-siteimprove FYI. I believe this will break the Code Checker Page Report that only receives the JSON serialization and thus need to read the correct field name.
> 
> Thanks for the heads up @Jym77. Good that the `serializationId` is still available as an alias. We can safely deploy this, update our usage and then delete the alias.

🤔 I am not sure because you only read the JSON, not the actual object.
Or should I keep the `serializationId` alias in the JSON too?

Jym77:
> > > @jfer-siteimprove @mvy-siteimprove FYI. I believe this will break the Code Checker Page Report that only receives the JSON serialization and thus need to read the correct field name.
> > 
> > 
> > Thanks for the heads up @Jym77. Good that the `serializationId` is still available as an alias. We can safely deploy this, update our usage and then delete the alias.
> 
> 🤔 I am not sure because you only read the JSON, not the actual object. Or should I keep the `serializationId` alias in the JSON too?

Should be good now ☺️

rcj-siteimprove (approved):

PR #1707, opened 2024-11-04
Support for `will-change` CSS property

rcj-siteimprove:
CSS property `will-change` has been added.


rcj-siteimprove:
!pr extract

packages/alfa-style/src/property/will-change.ts
```diff
@@ -0,0 +1,51 @@
...
+  "contents",
+];
+
+const animatableFeature: CSSParser<
+  Keyword<"scroll-position"> | Keyword<"contents"> | CustomIdent
+> = either(
+  either(Keyword.parse("scroll-position"), Keyword.parse("contents")),
+  CustomIdent.parse(
+    (customIdent) => !illegalCustomIdents.includes(customIdent.value),
+  ),
```
Jym77, line 34: **Question:** Given that for this property, the `<custom-ident>` must be property names, should we restrict that to property names? It feels this would makes things easier at a later stage when checking if that property is defined. OTOH, we'd need to accept all the existing properties, not just the ones we do support, otherwise we'd risk discarding valid declarations 🤔

packages/alfa-style/src/property/will-change.ts
```diff
@@ -0,0 +1,51 @@
...
+);
```
Jym77, line 35:
```suggestion
> = either(Keyword.parse("scroll-position", "contents"),
  CustomIdent.parse(
    (customIdent) => !illegalCustomIdents.includes(customIdent.value),
  ),
);
```

packages/alfa-style/test/property/will-change.spec.tsx
```diff
@@ -0,0 +1,79 @@
...
+      values: [
+        { type: "custom-ident", value: "transform" },
+        { type: "custom-ident", value: "opacity" },
+      ],
+    },
+    source: h.declaration("will-change", value).toJSON(),
+  });
+});
+
+test("#computed parses one keyword followed by custom identifiers", (t) => {
```
Jym77, line 45: **Chore:** We should have test(s) with more mixing in the list, e.g. a keyword in the middle of several `<custom-ident>`.

Jym77 (changes requested):

packages/alfa-style/src/property/will-change.ts
rcj-siteimprove: Should we add a list of all CSS properties, which will be independent from the ones we support, to check against?

Jym77: I'm not sure of it. It will be a pretty huge list (it must also contain the shorthands). And it will not help us in typing/being sure of the `get` because the unsupported properties will still fail.

That is, if we have the list of supported properties, we can use stuff like type `Property.Name`, or value `Object.keys(Longhands)`, … and ensure that `style.computed("will-change").some(property => style.cascaded(property).isSome())` works to detect whether one of the "will change" property exists on a given element. But since we don't support all properties, we can't guarantee that `property` has type `Property.Name` and thus that the calls will type correctly, and we'll need some extra poking and safeguarding around. So the benefits of effectively restricting them is low.

OTOH, the cost is somewhat high as we need a long list that also needs to be maintained when new properties are added to CSS (or at least used in `will-change`), which is not in our hands (vs auto-magically getting the names from the properties we support).

So, I am rather leaning toward simplicity and not trying to restrict it now. We may revisit that decision later if the benefits get higher.
Also, we should see how browsers react on a `will-change` for a non-existing property. Notably, do they discard the full declaration as invalid; or only ignore the unknown property? Due to forward-compatibility requirements that is often baked in CSS, I wouldn't be surprised of the later (to ensure old browsers don't totally break on newer pages). In that case, we must definitely not restrict ourselves. OTOH, if they do discard the full declaration, then we probably should do so (or we'll parse incorrect declaration and do the wrong thing in these cases) (but may decide not to for the sake of simplicity and revisit once we actually encounter a messed up page…)

packages/alfa-css/src/value/customIdent.ts
```diff
@@ -0,0 +1,81 @@
...
+
+const { map, parseIf } = Parser;
+const { and } = Predicate;
+
+/**
+ * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/custom-ident}
+ *
+ * @public
+ */
+export class CustomIdent
```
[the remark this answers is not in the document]
rcj-siteimprove: One difference is that keywords are not case sensitive whereas custom idents are not.

packages/alfa-style/src/property/will-change.ts
rcj-siteimprove: I did a quick test in Chrome: `will-change: scroll-position, foo` computed to `will-change: scroll-position` (also tried `will-change: foo` which expectedly computed to `will-change: auto`).

rcj-siteimprove:
!pr extract

packages/alfa-css/src/value/keyword.ts
```diff
@@ -16,60 +16,34 @@ const { equals } = Predicate;
...
 
 /**
  * @public
  */
 export namespace Keyword {
   export interface JSON<T extends string = string>
-    extends Value.JSON<"keyword"> {
-    value: T;
-  }
+    extends Ident.JSON<"keyword", T> {}
```
Jym77, line 46: **Suggestion:** This is not really needed (as done in `custom-ident`).

packages/alfa-css/src/value/keyword.ts
```diff
@@ -16,60 +16,34 @@ const { equals } = Predicate;
...
   public equals(value: unknown): value is this {
-    return value instanceof Keyword && value._value === this._value;
-  }
-
-  public hash(hash: Hash): void {
-    hash.writeString(this._value);
+    return value instanceof Keyword && super.equals(value);
   }
 
   public toJSON(): Keyword.JSON<T> {
```
Jym77, line 34:
```suggestion
  public toJSON(): Ident.JSON<"keyword", T> {
```

packages/alfa-css/src/value/keyword.ts
```diff
@@ -16,60 +16,34 @@ const { equals } = Predicate;
...
     return {
       ...super.toJSON(),
-      value: this._value,
     };
```
Jym77, line 37: **Suggestion:** `return super.toJSON()` (no need to spread and rebuild the object).
(which says we can probably just remove the `toJSON` here)

packages/alfa-css/src/value/custom-ident.ts
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public equals(value: unknown): value is this {
+    return value instanceof CustomIdent && super.equals(value);
+  }
+
+  public toJSON(): Ident.JSON<"custom-ident"> {
+    return {
+      ...super.toJSON(),
+    };
```
Jym77, line 36:
```suggestion
    return super.toJSON();
```
No need to spread and rebuild the same object.
🤔 Which actually means this is maybe not even needed

packages/alfa-css/src/value/custom-ident.ts
```diff
@@ -0,0 +1,64 @@
...
+      and(
+        isCustomIdent,
+        (customIdent) =>
+          !illegalCustomIdents.includes(customIdent.value.toLowerCase()),
+        predicate,
+      ),
+      map(Token.parseIdent(), (ident) => CustomIdent.of(ident.value)),
+      () => "Invalid custom-ident",
+    );
+  }
```
Jym77, line 63: **Suggestion:** I would rather do the checks on the string, not on the built object. It feels cleaner on the logic. Also for the caller can then provide a `Predicate<string>` without knowing much about the structure of `CustomIdent`.
```suggestion
  export function parse(
    predicate: Predicate<string> = () => true,
  ): CSSParser<CustomIdent> {
    return map(parseIf(
      and(
        (ident) =>
          !illegalCustomIdents.includes(String.toLowerCase(ident)),
        predicate,
      ),
      () => "Invalid custom-ident",
    ),
    CustomIdent.of);
  }
```

packages/alfa-style/src/property/will-change.ts
```diff
@@ -0,0 +1,48 @@
...
+  "contents",
+];
+
+const animatableFeature: CSSParser<
+  Keyword<"scroll-position"> | Keyword<"contents"> | CustomIdent
+> = either(
+  Keyword.parse("scroll-position", "contents"),
+  CustomIdent.parse(
+    (customIdent) => !illegalCustomIdents.includes(customIdent.value),
+  ),
```
Jym77, line 31:
```suggestion
  CustomIdent.parse(not(illegalCustomIdents.includes)),
```
should work with the change in `CustomIdent`.
(it doesn't read super well, OTOH, so maybe keeping it as `customIdent => !illegalCustomIdents.includes(customIdent)` is still better… Or we add a `String.isInArray` predicate, since we do this quite a lot 🤔 Maybe for later…

packages/alfa-style/src/property/will-change.ts
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/will-change}
+ * @internal
+ */
+export default Longhand.of<Specified, Computed>(
+  Keyword.of("auto"),
+  parse,
+  (value) => value,
+  { inherits: false },
```
Jym77, line 47:
```suggestion
```
no inheritance is the default.

packages/alfa-style/test/property/will-change.spec.tsx
```diff
@@ -0,0 +1,125 @@
+import { Device } from "@siteimprove/alfa-device";
+import { h } from "@siteimprove/alfa-dom";
+import { test } from "@siteimprove/alfa-test";
+
+import { Style } from "../../dist/index.js";
+
+const device = Device.standard();
+
+test("#computed parses keywords", (t) => {
+  for (let kw of ["scroll-position", "contents"] as const) {
```
Jym77, line 10:
```suggestion
  for (const kw of ["scroll-position", "contents"] as const) {
```

packages/alfa-style/test/property/will-change.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+        value: "auto",
+      },
+      source: null,
+    });
+  }
+});
+
+test("#computed does not parse a valid custom ident followed by an invalid custom ident", (t) => {
+  // TODO: For some reason `source` is not `null` for these keywords.
+  for (let kw of [
```
Jym77, line 85:
```suggestion
  for (const kw of [
```

packages/alfa-style/test/property/will-change.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+    t.deepEqual(style.computed("will-change").toJSON(), {
+      value: {
+        type: "keyword",
+        value: "auto",
+      },
+      source: h.declaration("will-change", kw).toJSON(),
+    });
+  }
+
+  for (let kw of [
```
Jym77, line 104:
```suggestion
  for (const kw of [
```

packages/alfa-style/test/property/will-change.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+        type: "keyword",
+        value: "auto",
+      },
+      source: null,
+    });
+  }
+});
+
+test("#computed does not parse a valid custom ident followed by an invalid custom ident", (t) => {
+  // TODO: For some reason `source` is not `null` for these keywords.
```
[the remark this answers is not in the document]
Jym77: This sounds indeed very weird given that the next test does have a `null` source 🤔 

It probably means that the parser did not fail. I'd say the `List.parse` in this case does not fail but returns an empty list of animatable features?
It seems that it would make sense to constrain the list to at least 1 element (which the `List.parse` can easily do iirc).

Jym77 (changes requested):
Looks good 🎉 
Some clean up and polishing.

packages/alfa-css/src/value/ident.ts
```diff
@@ -0,0 +1,67 @@
+import type { Hash } from "@siteimprove/alfa-hash";
```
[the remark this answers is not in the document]
Jym77: Good points 😅 
`textual` would make the most sense, as it mimics the CSS grouping. `url` should indeed be moved there too, it is very ad-hoc that it was put in `image` (just because this is the only super-value using it in Alfa…)

packages/alfa-style/src/property/will-change.ts
rcj-siteimprove: It does look nicely succinct, but I think we would also need to instantiate and store the illegal custom idents and not as strings, i.e. `const illegal = [ CustomIdent.of("will-change"), ...]`. I'll leave this one for later :)

Jym77: No, if `CustomIdent.parse` is updated to take a `Predicate<string>` as per https://github.com/Siteimprove/alfa/pull/1707#discussion_r1839804289, that should work 🤔

rcj-siteimprove: Oh, sorry, I missed that comment 😖 Then it makes sense 👍🏼

Jym77 (approved):
Well done 🎉

PR #1711, opened 2024-11-18
Support for `mask-*` CSS properties

rcj-siteimprove:
Resolves #


rcj-siteimprove:
!pr extract

rcj-siteimprove:
!pr extract

rcj-siteimprove:
!pr extract

rcj-siteimprove:
!pr extract

packages/alfa-style/src/property/mask.ts
```diff
@@ -0,0 +1,153 @@
...
+    const clip = box2 ?? box1;
+
+    return [image, pos, size, repeat, origin, clip, composite, mode] as const;
+  },
+);
+
+const parseList = separatedList(
+  maskLayer,
+  delimited(option(Token.parseWhitespace), Token.parseComma),
+);
```
Jym77, line 102: **Question:** Isn't that a `List.parseCommaSeparated`?

packages/alfa-style/src/property/mask.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * {@link https://developer.mozilla.org/en-US/docs/Web/CSS/border-top}
```
Jym77, line 105: **Chore:** The link to `border-top`  looks weird.

packages/alfa-style/test/property/mask-clip.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+  t.deepEqual(computed(element, "mask-clip"), {
+    value: {
+      type: "list",
+      separator: " ",
+      values: [
+        {
+          type: "keyword",
+          value: "border-box",
+        },
+      ],
```
Jym77, line 18:
```suggestion
      values: [ { type: "keyword", value: "border-box" } ],
```
When prettier is happy with it, I tend to prefer avoiding delimiters-only lines in tests 🙂 (but it's a bit annoying to do the switch everywhere, so feel free to ignore it…)

(alternatively / additionally, it might be worth testing `computed("foo").value.values` just to get rid of the list wrapper (it needs to be kept once to check the separator) and reduce the indentation level and number of empty-ish lines)

packages/alfa-style/test/property/mask-clip.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+    },
+    source: h
+      .declaration("mask-clip", "view-box, fill-box, border-box")
+      .toJSON(),
+  });
+});
+
+test("#computed repeats values when there are more layers than values", (t) => {
```
Jym77, line 112: **Suggestion:** These two feel more like tests for `matchLayers` and should be extracted as unit tests for it rather than repeated in every property.

packages/alfa-style/src/property/helpers/mask-layers.ts
```diff
@@ -0,0 +1,38 @@
...
+ * {@link https://drafts.fxtf.org/css-masking/#layering}.
+ *
+ * @remarks
+ * The computed value depends on the number of layers.
+ * A layer is created for each of the comma separated values for `mask-image`.
+ *
+ * If there are more values than layers, the excess values are discarded.
+ * Otherwise, the values must be repeated until the number of values matches the number of layers.
+ */
+export function matchLayers<V extends Value>(
```
Jym77, line 15: **Suggestion:** Given that the `Style` is only used to compute a number, this could take that number directly, and be moved as `List#extendsToLength` or something similar. That name would match the effect (rather than the intention), and the method would belong to the data structure it changes, improving cohesion.

`List#extendsToLength` could also take another `List` as parameter and get the length itself (instead or in addition to taking a number).

packages/alfa-style/src/property/helpers/mask-layers.ts
```diff
@@ -0,0 +1,38 @@
...
+  value: List<V>,
+  style: Style,
+): List<V> {
+  const numberOfLayers = Math.max(
+    style.computed("mask-image").value.values.length,
```
Jym77, line 20: **Suggestion:** This feels like we should add a `List#size` and use it instead of `List#values.length`.

packages/alfa-style/test/property/mask-position.spec.tsx
```diff
@@ -0,0 +1,227 @@
...
+          },
+        },
+      ],
+    },
+    source: null,
+  });
+});
+
+// TODO: The spec requires the computed value to be two lengths or percentages, not a keyword value.
+// E.g. the keyword `left` should be computes to `0% 50%` in Chrome and Firefox.
```
Jym77, line 46: **Note:** This may actually be a used value 🤔 [CSS spec](https://drafts.fxtf.org/css-masking/#the-mask-position) states that the computed value should be keywords + offsets. Anyway, it's probably OK as is if we don't need more 😄 (this may be something that ultimately needs to be fixed in the `Position` type)

The discrepancy note should be added to the source file, not just to the test.

Jym77 (changes requested):
Looks good 🎉 Some improvements and polishing. Nothing really blocking.

packages/alfa-style/src/property/helpers/mask-layers.ts
```diff
@@ -0,0 +1,38 @@
+import { List, type Value } from "@siteimprove/alfa-css";
+
+import type { Style } from "../../style.js";
+
+/**
+ * {@link https://drafts.fxtf.org/css-masking/#layering}.
```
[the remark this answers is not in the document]
rcj-siteimprove: Yes, I came across that as well, I'll create a follow up issue for that 👍🏼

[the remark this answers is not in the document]
rcj-siteimprove: New issue: https://github.com/Siteimprove/alfa/issues/1730

packages/alfa-style/src/property/mask-image.ts
```diff
@@ -0,0 +1,48 @@
...
+  List.of([MaskReference.initialItem], ", "),
+  List.parseCommaSeparated(MaskReference.parse),
+  (value, style) =>
+    value.map((images) =>
+      images.map((image) =>
+        Selective.of(image)
+          .if(Image.isImage, (image) =>
+            image.partiallyResolve(Resolver.length(style)),
+          )
+          .get(),
```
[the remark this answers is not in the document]
rcj-siteimprove: I'm not sure about this one. In fact it's the same compute function as `background-image` since I thought they should be the same. Should we look into if changes are needed in both places?

packages/alfa-style/src/property/mask.ts
rcj-siteimprove: `List.parseCommaSeparated` only accepts a parser whose result is an `alfa-css` `Value`, but the result of the shorthand parser is not a `Value`. The same applies in [background.ts](https://github.com/Siteimprove/alfa/blob/092e524423ea29e494077c001e88981f05aa19a8/packages/alfa-style/src/property/background.ts#L199-L202). Maybe the `parse[Comma|Space]Separated` functions can be lifted to `alfa-parser` for better reuse, but maybe something for later.

rcj-siteimprove:
!pr extract

rcj-siteimprove:
!pr extract

packages/alfa-style/src/property/mask.ts
Jym77: Ah, good point.
We can't really lift these to `alfa-parser` more than the existing `separatedList`, because the `Token` are an `alfa-css` thing. Best case, we may refactor `delimited(option(Token.parseWhitespace), Token.parseComma)` into a single parser since it's used a lot, but that's a side-improvement for another time.

rcj-siteimprove:
!pr extract

Jym77 (approved):

PR #1713, opened 2024-11-19
Add SIA-R114: `<title>` is descriptive

Jym77:
Resolves II-8675.

The rule is currently kept as experimental until the backend can automate the answer.


rcj-siteimprove (approved):

packages/alfa-rules/test/sia-r114/rule.spec.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+test("evaluate() is inapplicable to documents without title", async (t) => {
+  const document = h.document([
+    <html>
+      <div>Some text in English</div>
+    </html>,
+  ]);
+
+  t.deepEqual(await evaluate(R114, { document }), [inapplicable(R114)]);
+});
```
ricflams, line 133: These two tests are already defined earlier

ricflams (commented):
LGTM, but I found some duplicated code

packages/alfa-rules/test/sia-r114/rule.spec.tsx
Jym77: Fixed in #1715.

PR #1716, opened 2024-11-22
Add filtering of visible descendants in clickable box computation

rcj-siteimprove:
This PR fixes an issue in the clickable box computation where the bounding boxes of all descendants would be used. For invisible descendants their bounding boxes would be an empty rectangle located at (0, 0) and since we are letting the clickable area be the convex hull of the boxes (the smallest containing rectangle), this would give wildly incorrect clickable areas in some cases. This was uncovered after `<select>`s were made visible.

More work will be needed to refine how R111 and R113 use the clickable boxes for sizing and spacing detection.

### DIFF
--- a/.changeset/grumpy-games-fix.md
+++ b/.changeset/grumpy-games-fix.md
@@ -1,0 +1,7 @@
+---
+"@siteimprove/alfa-rules": minor
+---
+
+**Fixed:** The function `getClickableBox` now only considers visible descendants.
+
+The return type of the function was also changed from `Option` to `Result`.
--- a/.changeset/grumpy-games-fix.md
+++ b/.changeset/grumpy-games-fix.md
@@ -1,7 +1,5 @@
 ---
-"@siteimprove/alfa-rules": minor
+"@siteimprove/alfa-rules": patch
 ---
 
-**Fixed:** The function `getClickableBox` now only considers visible descendants.
-
-The return type of the function was also changed from `Option` to `Result`.
+**Fixed:** SIA-R111 and SIA-R113 now ignores invisible descendants of interactive elements when computing the clickable areas.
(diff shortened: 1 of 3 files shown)


.changeset/grumpy-games-fix.md
```diff
@@ -0,0 +1,7 @@
+---
+"@siteimprove/alfa-rules": minor
+---
+
+**Fixed:** The function `getClickableBox` now only considers visible descendants.
+
+The return type of the function was also changed from `Option` to `Result`.
```
Jym77, line 7:
```suggestion
---
"@siteimprove/alfa-rules": patch
---

**Fixed:** SIA-R111 and SIA-R113 now ignores invisible descendants of interactive elements when computing the clickable areas.
```
* Since we do not alter any public API (just an internal one), this is a `patch` bump (even if the `option` => `Result` change is not backward-compatible). If we want to emphasise that change (e.g. for our internal consumers), that would deserve a separate breaking changeset; but here it should be OK since nobody else than Alfa is using this function.
* Since the actual change is not really exposed to users, the changeset should focus on the visible fix, i.e. the improvements in the rules.

Jym77 (dismissed):

packages/alfa-rules/src/common/dom/get-clickable-box.ts
```diff
@@ -2,39 +2,55 @@ import { Cache } from "@siteimprove/alfa-cache";
...
+ * or an error if the element or one of its descendants doesn't have a bounding box
+ * or if the element itself is not visible.
  *
  * @remarks
  * This function assumes that the element can receive pointer events, i.e. is clickable.
  * If called on an element that is not clickable, the function will still return the
  * the area that would be clickable if the element could receive pointer events.
  *
+ * The clickable box is approximated by the smallest rectangle containing
+ * the bounding boxes of the element and all it's visible descendants.
```
rcj-siteimprove, line 24:
```suggestion
 * the bounding boxes of the element and all its visible descendants.
```

Jym77 (approved):

PR #1720, opened 2024-11-25
Add memoization to `Cache`

Jym77:


packages/alfa-aria/src/dom/predicate/is-programmatically-hidden.ts
```diff
@@ -34,39 +34,32 @@ export function isProgrammaticallyHidden(
...
+    (node: Node) =>
+      node.parent(Node.fullTree).some(_hasHiddenAncestors(device, context)),
+  );
 }
+
+const hasHiddenAncestors = Cache.memoize<
+  unknown,
+  [Device, Context],
+  Predicate<Node>
+>(_hasHiddenAncestors);
```
Jym77, line 65: I'm not sure why TS needs some skaffolding here, but not in the simple cases in the tests 🤷

packages/alfa-aria/src/dom/predicate/is-programmatically-hidden.ts
```diff
@@ -34,39 +34,32 @@ export function isProgrammaticallyHidden(
   );
 }
 
-const cache = Cache.empty<Device, Cache<Context, Cache<Node, boolean>>>();
-
-function hasHiddenAncestors(
+function _hasHiddenAncestors(
```
Jym77, line 37: This is just a demo of using the new decorator, not directly related to the changes.

packages/alfa-aria/src/dom/predicate/is-programmatically-hidden.ts
```diff
@@ -34,39 +34,28 @@ export function isProgrammaticallyHidden(
...
+          device,
+          context,
         ),
-      );
+        hasAttribute("aria-hidden", equals("true")),
+      ),
+    ),
+    // Or its parent is programmatically hidden
+    (node: Node) =>
+      node.parent(Node.fullTree).some(_hasHiddenAncestors(device, context)),
```
rcj-siteimprove, line 57: **Question:** Is it on purpose the call is to `_hasHiddenAncestors` and not `hasHiddenAncestors`? I can't tell if it makes a difference, but in the original code, in the recursion the cache would be used, but now it's going to recurse without the cache, I think.

Jym77: Ouch. Good point. 🤔

rcj-siteimprove (approved):
Nice 👏🏼

PR #1721, opened 2024-11-28
Expose ariaAllowedAttributes on ARIA Element

srnjcbsn:
The logic is moved from R18 implementation and made public.

Should we also have a `prohibitedAttributes()` function on the Element type, for symmetry?


.changeset/twenty-boxes-complain.md
```diff
@@ -0,0 +1,8 @@
+---
+"@siteimprove/alfa-rules": minor
+"@siteimprove/alfa-aria": minor
+---
+
+**Added:** Expose "allowedAttributes" on ARIA Element type.
```
Jym77, line 6:
```suggestion
**Added:** Expose `allowedAttributes` on ARIA Element type.
```
This will be rendered as markdown in the changelog.

packages/alfa-aria/src/node/element.ts
```diff
@@ -115,6 +119,65 @@ export class Element extends Node<"element"> {
       ...this._children.map((child) => String.indent(child.toString())),
     ].join("\n");
   }
+
+  private static allowedAttributesForInputType(
+    inputType: InputType
+  ): readonly Attribute.Name[] {
```
Jym77, line 125:
```suggestion
  ): ReadonlyArray<Attribute.Name> {
```
We usually use the `Array<T>` type syntax rather than `T[]` 🤷

packages/alfa-aria/src/node/element.ts
```diff
@@ -115,6 +119,65 @@ export class Element extends Node<"element"> {
...
+      default:
+        return [];
+    }
+  }
+
+  /**
+   * The attributes that are allowed on this element, taking into consideration ARIA in HTML conformance requirements.
+   * See {@link https://w3c.github.io/html-aria/#docconformance}
+   */
+  public allowedAttributes(): readonly Attribute.Name[] {
```
Jym77, line 157:
```suggestion
  public allowedAttributes(): ReadonlyArray<Attribute.Name> {
```

Jym77:
> Should we also have a `prohibitedAttributes()` function on the Element type, for symmetry?

No. YAGNI. We'll move it when we need it…

Jym77:
The Integrate workflows fails because it detects change in the API (the new method) that has not been registered in the docs.

You can extract the new API by using `yarn extract packages/alfa-aria` and committing the changed file. It can also be done automagically from the PR by commenting `!pr extract` (note do not push before that workflow ends, or it will fail to push its results…) I'll do that in the next comment.

Jym77:
!pr extract

srnjcbsn:
> The Integrate workflows fails because it detects change in the API (the new method) that has not been registered in the docs.
> 
> You can extract the new API by using `yarn extract packages/alfa-aria` and committing the changed file. It can also be done automagically from the PR by commenting `!pr extract` (note do not push before that workflow ends, or it will fail to push its results…) I'll do that in the next comment.

I ran it locally and force pushed approximately when you were writing that comment, I think (hadn't noticed you had started reviewing, so I thought I could sneak it in)

Jym77 (approved):

Issue #1733, opened 2024-12-12
Consider making all constructors protected instead of private


Jym77:
Most of our classes constructors are `private`. This effectively prevents the classes to be extended as it prevents a call to `super`.
We should maybe make them `protected` which would let consumers extends the classes if they want to enrich the data.

Jym77:
Done.

PR #1735, opened 2024-12-18
Protected constructors

Jym77:
Part of #1733

* Make constructors that are not part of a Singleton pattern `protected` instead of `private`. This let consumers extend the class if they want.
* Fix a missing singleton pattern.


Jym77:
!pr extract

rcj-siteimprove (approved):

PR #1736, opened 2024-12-18
Make packages type-only

### LINKED ISSUE #1732: Improve handling of type-only packages.
We have several packages that only export types. These still have an `index.ts` file which `export * from "foo.ts"`. We should at the very least replace it by `export type *`, and probably look into ways of only packaging the `.d.ts` files as the rest is useless.

Jym77:
Resolves #1732

The "missing" `main` in `package.json` is what DefiniteylTyped does ([example](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/package.json)), so it should be good enough.


rcj-siteimprove (approved):
👏🏼

PR #1745, opened 2025-01-22
Package skaffolding

Jym77:
Add a script to create an Alfa package/workspace with a default configuration.


packages/alfa-toolchain/src/create-alfa-package.ts
```diff
@@ -0,0 +1,197 @@
...
+git.add([
+  path.join(packageDir, "src", "index.ts"),
+  path.join(packageDir, "CHANGELOG.md"),
+  path.join(packageDir, "config", "api-extractor.json"),
+  path.join(packageDir, "README.md"),
+  path.join(packageDir, "tsconfig.json"),
+  path.join(packageDir, "src", "tsconfig.json"),
+  path.join(packageDir, "test", "tsconfig.json"),
+  path.join(packageDir, "package.json"),
+]);
```
rcj-siteimprove, line 191: **Question: ** It looks like this is the only place simple-git is used. Why not just let the user manually add the files after running the script? Not that it matters much though.

rcj-siteimprove (approved):
Nice addition 💪🏼

PR #1756, opened 2025-02-06
New package `alfa-painting-order`

rcj-siteimprove:
This PR introduces a new package which contains functionality for computing the painting order of HTML elements based on the specification: https://drafts.csswg.org/css2/#painting-order.

The process for computing the painting order is fairly complex and that reflects in the implementation. This is the first step to make something that works in most common use cases. There are possibly room for caching some computations and other optimizations as well as making the algorithm easier to follow and less error-prone.

The current implementation does not consider scrolling frames, but that could be a logical next step if needed.

The painting order can be used for e.g. detecting if elements are fully or partially covered by other elements, which is relevant when determining the clickable area of an element. More on that in future PRs.


packages/alfa-painting-order/src/painting-order.ts
```diff
@@ -0,0 +1,241 @@
...
+      }
+    }
+
+    for (const blockLevel of blockLevels) {
+      if (!defer && createsSC(blockLevel)) {
+        paint(blockLevel, canvas);
+      } else {
+        canvas.push(blockLevel);
+      }
+    }
```
Jym77, line 207: **Suggestion:** It looks like this loop could be refactored by abstracting the array it traverses and the second part of the predicate.
The `positionedOrStackingContexts` can also be split at negative `z-index` and thus the first and last loops also fit in that refactoring.

packages/alfa-painting-order/src/painting-order.ts
```diff
@@ -0,0 +1,241 @@
...
+    canvas: Array<Element>,
+    options: { defer?: boolean } = {
+      defer: false,
+    },
+  ): void {
+    const { defer = false } = options;
+    const positionedOrStackingContexts: Array<Element> = [];
+    const blockLevels: Array<Element> = [];
+    const floats: Array<Element> = [];
+    const inlines: Array<Element> = [];
```
Jym77, line 91: **Suggestion:** It feels like we could use an abstraction ("layer"?) to wrap these. Then `distributeIntoLayers` would produce that abstraction and the next step of pushing things in the `canvas` is a `.toPaintingOrder` (or `.fromLayer`) builder.

packages/alfa-painting-order/src/painting-order.ts
```diff
@@ -0,0 +1,241 @@
...
+      inlines.push(element);
+    }
+
+    function traverse(element: Element) {
+      for (const child of element
+        .children(Node.fullTree)
+        .filter(and(Element.isElement, rendered))) {
+        distributeIntoLayers(child);
+
+        if (or(isPositioned, isFloat, createsSC)(child)) {
```
Jym77, line 168: **Suggestion:** It seems that this test is more or less the same that the one(s) in `distributeIntoLayers` that decide to recurse in the subtree. It feels smelly to have that in two places (and error prone if we change one), especially given that they are not written exactly the same way. Not sure how we can improve that, though.

Jym77 (changes requested):
Looks good. Some code organisation comments.

rcj-siteimprove:
!pr extract

rcj-siteimprove:
!pr extract

packages/alfa-painting-order/src/painting-order.ts
rcj-siteimprove: I tried to get the abstraction right and I think it should be doable, but I'd like to do it however as part of a future improvement PR. I think, in any case, that I am going to improve on this once we start using it.

packages/alfa-painting-order/src/painting-order.ts
rcj-siteimprove: Same as the above comment, I'd like to improve on this and make the code better structured in the near future.

rcj-siteimprove:
@Jym77 I addressed most of the comments, except two that I would like to look into in a follow up PR, but after the other work that depend on this one (using it to compute clickable area).

packages/alfa-painting-order/src/painting-order.ts
```diff
@@ -0,0 +1,286 @@
...
+  public static of(elements: Iterable<Element>): PaintingOrder {
+    return new PaintingOrder(Array.from(elements));
+  }
+
+  private readonly _elements: Array<Element>;
+
+  protected constructor(elements: Array<Element>) {
+    this._elements = elements;
+  }
+
```
Jym77, line 42:
```suggestion

public get elements(): Iterable<Element> {
  return this._elements;
}

```

packages/alfa-painting-order/src/painting-order.ts
```diff
@@ -0,0 +1,286 @@
...
+  public equals(value: this): boolean;
+  public equals(value: unknown): value is this;
+  public equals(value: unknown): boolean {
+    return (
+      value === this ||
+      (PaintingOrder.isPaintingOrder(value) &&
+        Array.equals(value._elements, this._elements))
+    );
+  }
+  public hash(hash: Hash): void {
```
Jym77, line 52:
```suggestion
  }
  
  public hash(hash: Hash): void {
```

packages/alfa-painting-order/src/painting-order.ts
```diff
@@ -0,0 +1,286 @@
...
+
+  protected constructor(elements: Array<Element>) {
+    this._elements = elements;
+  }
+
+  public equals(value: this): boolean;
+  public equals(value: unknown): value is this;
+  public equals(value: unknown): boolean {
+    return (
+      value === this ||
```
Jym77, line 47:
```suggestion
```
We usually don't have the physical equality test in our `.equals` (which does indeed make them less efficient in that case, but that is hopefully a rare case 🤔

packages/alfa-painting-order/src/painting-order.ts
```diff
@@ -0,0 +1,286 @@
...
+   * @public
+   */
+  export const from = Cache.memoize(function (
+    root: Element,
+    device: Device,
+  ): PaintingOrder {
+    return PaintingOrder.of(paint(device, root, Sequence.empty()));
+  });
+
+  const getZLevel = Cache.memoize(function (device: Device, element: Element) {
```
Jym77, line 101: **Suggestion:** This looks cheap to compute (because computed styles are already cached) and probably doesn't need to be cached.

packages/alfa-painting-order/src/painting-order.ts
```diff
@@ -0,0 +1,286 @@
...
+
+  function paint(
+    device: Device,
+    element: Element,
+    canvas: Sequence<Element>,
+    options: { defer?: boolean } = {
+      defer: false,
+    },
+  ): Sequence<Element> {
+    const { defer = false } = options;
```
Jym77, line 149: **Suggestion:** This defines the default for `defer` twice, which can get confusing. E.g. if the second one changes to `const { defer = true } = options`, it requires `options` to be `{}` in order to trigger (`undefined` will get the other default value).

packages/alfa-painting-order/src/painting-order.ts
```diff
@@ -0,0 +1,286 @@
...
+    const [negatives, nonNegatives] = sortAndSplitByZLevel(
+      device,
+      positionedOrStackingContexts,
+    );
+
+    canvas = paintLayer(device, canvas, negatives, element, defer);
+    canvas = paintLayer(device, canvas, blockLevels, element, defer);
+    canvas = paintLayer(device, canvas, floats, element, defer);
+    canvas = paintLayer(device, canvas, inlines, element, defer);
+    canvas = paintLayer(device, canvas, nonNegatives, element, defer);
```
Jym77, line 251: **Note:** It looks like `paintLayer` is only adding to the canvas, not changing more. So it could just return the new elements and this would be a big `canvas.concat`.

Probably for another time, though.

Jym77 (changes requested):

packages/alfa-painting-order/src/painting-order.ts
rcj-siteimprove: That makes sense, I'll keep that in mind for the future.

rcj-siteimprove:
!pr extract

Jym77 (approved):

PR #1773, opened 2025-03-17
Add dedupe-on-install plugin

Jym77:
This should avoid duped packages to even appear, and will notably enable the use of Dependabot in a more streamlined way.


rcj-siteimprove (approved):
Nice 🌮

PR #1775, opened 2025-03-19
SIA-R83: use text layout

Jym77:


packages/alfa-rules/src/sia-r83/rule.ts
```diff
@@ -188,29 +180,37 @@ function overflow(
...
-        test(
-          hasBox(
+        target
+          .getBoundingBox(device)
+          .orElse(() =>
+            target
+              .parent(Node.fullTree)
+              .filter(isElement)
+              .flatMap((parent) => parent.getBoundingBox(device)),
+          )
```
rcj-siteimprove, line 201: **Question:** Do we handle the situation where text content consists of multiple text nodes and therefore also multiple boxes, e.g. `<div>foo <b>bar</b></div>` where each individual text node might have room enough, but the content as a whole doesn't?

rcj-siteimprove (approved):
Looks good. Just one question/concern, might not be relevant.

packages/alfa-rules/src/sia-r83/rule.ts
Jym77: I think so 😅 Because for that to clip it needs to not be allowed to wrap at the spaces 🤔

rcj-siteimprove (approved):

Issue #1210, opened 2022-08-04
Support `content-visibility` and `hidden="until-found"`


Jym77:
- [ ] Support the new property and the new value.
- [ ] Update the UA stylesheet for `hidden="until-found"`
- [ ] Update `isProgrammaticallyHidden`.

Issue #954, opened 2021-10-18
Investigate better handling of pseudo-elements


Jym77:
Pseudo-elements can be used to inject a background colour, dropping them causes false positives in R69 (among other).

```css
.advance-header-inner .content-left::before {
  background: #FC5D0D;
  border-radius: 0 30px 30px 0;
}
```

---
Pseudo-element can also be used to show focus indicator:
```css
.foo:focus::after {
  position: absolute;
  top: -4px;
  right: -8px;
  bottom: -4px;
  left: -8px;
  content: "";
  display: block;
  border: 2px solid #b51133;
  border-radius: 0.25rem;
}
```

---
Pseudo-elements can be used to inject (part of) accessible name: [Accessible name testable statement 552](https://www.w3.org/wiki/AccName_1.1_Testable_Statements#Name_test_case_552) (and several others).

- [ ] Follow up on DEV-12423
- [ ] Follow up on DEV-12462
- [ ] Follow up on CST-2609

PR #1777, opened 2025-03-24
Update user agent sheet for `<details>` to fix visibility of content

### LINKED ISSUE #1776: `isVisible` reports content of closed `<details>`  as visible
The [isVisible](https://github.com/Siteimprove/alfa/blob/ee0c470dab051834d0024ba3868b9c7806a24193/packages/alfa-style/src/node/predicate/is-visible.ts#L29) predicate does not detect content inside a closed `<details>` element as invisible.

Minimal reproducible example:
```
<!doctype html>
<html>
  <head>
    <title>Details Visibility MRE</title>
  </head>
  <body>
    <details>
      <summary>Minimal Reproducible Example</summary>
      <div>I'm invisible</div>
    </details>
  </body>
</html>
```

Comparing with browser dev tools, the `<div>` element has computed styles `visibility: visible` and `display: block` so it's unclear what exactly is the mechanism that makes the element not visible/not rendered.

`<details>` is not added to [user-agent.ts](https://github.com/Siteimprove/alfa/blob/ee0c470dab051834d0024ba3868b9c7806a24193/packages/alfa-cascade/src/user-agent.ts#L4) according to [rendering.html](https://html.spec.whatwg.org/multipage/rendering.html#the-details-and-summary-elements), but it's unclear if adding that will be enough to correctly detect the visibility of the content.

Do we need to add special handling of `<details>` elements in the predicate?

Relates to [II-8780](https://siteimprove-wgs.atlassian.net/browse/II-8780).

[MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details).

[II-8780]: https://siteimprove-wgs.atlassian.net/browse/II-8780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

rcj-siteimprove:
Resolves #1776 

This solution doesn't actually follow the standard, unfortunately, in stead it tries to approximate the observed behavior seen in the browser.

As far as I could find, it's going to be difficult to implement it correctly with what we currently have as it involves a style rule `content-visibility: hidden` of pseudo element `::details-content`.

Can we allow this discrepancy or will it bite us?


packages/alfa-style/test/node/predicate/is-visible.spec.tsx
```diff
@@ -363,3 +363,31 @@ test("isVisible() returns true for single line `<select>`", (t) => {
...
+  const details = (
+    <details>
+      <summary>Hello</summary>
+      {content}
+    </details>
+  );
+
+  h.document([details]);
+
+  t(!isVisible(content));
```
Jym77, line 392: **Suggestion:** We should also test `t(isVisible(summary))` to prevent a regression where we make everything invisible.

Jym77 (changes requested):
This should improve the situation until we have support for the missing bits (#1210 and #954 🙈) 👍

Jym77 (approved):

PR #1784, opened 2025-03-27
SIA-R83: use hyphenation properties

### LINKED ISSUE #1741: SIA-R83 should handle `overflow-wrap` and `hyphens`
We've seen a few false positives on `<div>` or `<li>` that contain a single word. The element's bounding box is the full line, so Alfa doesn't see that the text is actually short and doesn't need wrapping, and since the text is a single word Alfa doesn't find a soft wrap opportunity and flags it incorrectly.

Several of these had `overflow-wrap` set up that means no soft wrap points are needed to wrap, similarly, `hyphens` can create soft-wrap points inside single words (depending on script/`lang`).

We could also try and detect short words (how?) when the element containing them is very wide.

```html
<ul style="overflow: hidden">
  <li>Hello</li>
</ul>
```

- [x] Follow up on CST-3729
- [x] Follow up on CST-3737
- [x] Follow up on CST-3741
- [x] Follow up on CST-3893
- [x] Follow up on CST-3911
- [x] Follow up on CST-4048

Jym77:
Resolves #1741


Jym77:
!pr command

packages/alfa-style/src/text/predicate/has-soft-wrap-opportunity.ts
```diff
@@ -0,0 +1,53 @@
...
+import { hasComputedStyle } from "../../element/predicate/has-computed-style.js";
+
+import { hasHyphenationOpportunity } from "./has-hyphenation-opportunity.js";
+
+const { or } = Predicate;
+
+/**
+ * Test whether a Text node has soft wrap opportunities.
+ *
+ * @PrivateRemarks
```
rcj-siteimprove, line 16: **Nitpick:**
```suggestion
 * @privateRemarks
```

rcj-siteimprove (approved):
Looks good, just found one nitpick.

rcj-siteimprove (approved):

PR #1787, opened 2025-03-31
Setup dependabot config

Jym77:
Switch to a working 🤞 dependabot config for all updates.
We should monitor this for some times, then if it works fine:
- [ ] Delete renovate.json, de-activate Renovate (and forking-renovate).
- [ ] Make similar changes to other Alfa repos.
- [ ] Delete `config/alfa-renovate.json` that is shared with the rest.

Note that other repos may have to use a `registries` key to find packages in the Github registry.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,19 +1,59 @@
 version: 2
 updates:
   - package-ecosystem: "npm"
-    directory: "/"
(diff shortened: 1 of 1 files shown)


rcj-siteimprove (approved):
Looks good 👍🏼

PR #1789, opened 2025-04-02
Fix serialization of imported stylesheets

rcj-siteimprove:
The `rules` property of `ImportRule` does not use the private field `_rules` which is left unset. Instead it calls to `_sheet.rules`. The parent class however serializes the `_rules` field which is then empty in the serialized output. To fix this, we override the serialization in `ImportRule` to use the public `rules` property to serialize `_sheet.rules`.


rcj-siteimprove:
!pr extract

mvy-siteimprove (approved):

PR #1845, opened 2025-07-24
SIA-R117: Image accessible name is descriptive

rcj-siteimprove:
New rule corresponding to the ACT rule [Image accessible name is descriptive](https://www.w3.org/WAI/standards-guidelines/act/rules/qt1vmo/).
 
Unrelated to the new rule, I updated yarn upon suggestion.


packages/alfa-rules/src/sia-r117/rule.ts
```diff
@@ -0,0 +1,87 @@
...
+       * In the ACT rule, an element is **not** applicable if it has an ancestor in the flat tree that is named from
+       * author, or it is an img element where the current request's state is not completely available (e.g. if the image
+       * doesn't exist).
+       *
+       * Regarding the first condition, we could recurse the DOM and stop descending as
+       * soon as we encounter an element named from author. I'm not yet sure how to determine if an element is named
+       * from author, but I think we can use `Name.source`.
+       *
+       * I don't think the second condition is possible to check in Alfa, and I'm not sure if we want
+       * to ask a question about it.
```
srnjcbsn, line 46: > I'm not sure if we want to ask a question about it.

The second condition from the ACT makes sense, but I don't understand this part. Are you suggesting not asking the question if we can't determine whether the image is available? I suppose this will be the case for all images with an `src` that points to a network resource.

packages/alfa-rules/src/common/act/question.ts
```diff
@@ -263,5 +263,10 @@ export namespace Question {
       type: "boolean",
       message: "Does this heading describe the following content?",
     },
+    // R117 (experimental)
+    "is-image-accessible-name-descriptive": {
+      type: "boolean",
+      message: "Does the accessible name of this element describe the image?",
```
srnjcbsn, line 269: I don't know where these question messages show up so this is probably super nitpicky, but would it make sense to reword this into something along the lines of the [ACT's description](https://www.w3.org/WAI/standards-guidelines/act/rules/qt1vmo) ("This rule checks that the accessible names of images serve an equivalent purpose to the image.")? 

Eg:
> Does the accessible name of this element serve an equivalent purpose to the image?

Or, based on [G94](https://www.w3.org/WAI/WCAG21/Techniques/general/G94): 
> Does the accessible name of this element present the same information as the image?

Both are way more clunky than the original message though.

rcj-siteimprove: I think it could make sense. I do kind of like the "present the same information" suggestion, so I think I will go with that.

rcj-siteimprove:
!pr extract

packages/alfa-rules/src/sia-r117/rule.ts
rcj-siteimprove: I'm not suggesting not asking the "is accessible name descriptive?" question asked by the expectation, this a about a different (hypothetical) question asked by the applicability. For some rules I think we ask questions in the applicability if Alfa is not able to determine automatically if the rule even applies to an element. This could be such a case because Alfa is not able to know if the image was not loaded e.g. if the URL is wrong. So we could in theory ask a question like "does the URL point to a valid image?" or "does the image exist?" and if the answer would be no, then the rule wouldn't apply to the given element.

My first thought was that we don't want to do such a thing and we'd rather just have the rule (incorrectly) apply to broken images, but come to think of it, I guess we do know if an image is broken or we can assume that it is if we couldn't download it so we could actually ask about it in the applicability 🤔 

Anyway, don't know if the above made sense, so we can also have a talk about if it didn't.

srnjcbsn (approved):

packages/alfa-rules/src/sia-r117/rule.ts
```diff
@@ -35,24 +35,49 @@ export default Rule.Atomic.of<Page, Element, Question.Metadata>({
...
+        let targets: Array<Element> = [];
+
+        const applicable = and(
+          or(hasName("img"), hasName("canvas"), hasName("svg")),
+          isVisible(device),
+          hasAccessibleName(device),
         );
+
+        const namedFromAuthor = hasAccessibleName(device, (name) =>
+          name.source.some(({ type }) => type === "label"),
```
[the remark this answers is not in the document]
rcj-siteimprove: No, good point, I have added `"reference"` to the check.

srnjcbsn (approved):

PR #1900, opened 2025-10-16
Update packages dependency graphs

Jym77:
* Add dependency graphs generation to release cycle.
* Re-generate all individual package graphs.


rcj-siteimprove (approved):
Looks good 👍🏼

PR #2070, opened 2026-04-20
SIA-R101

rcj-siteimprove:


rcj-siteimprove:
!pr extract

packages/alfa-rules/src/common/act/question.ts
```diff
@@ -248,6 +248,11 @@ export namespace Question {
       type: "boolean",
       message: `Is this element visible when it's focused?`,
     },
+    // R101 (experimental)
+    "has-repeated-content-before-main": {
+      type: "boolean",
+      message: "Does the page have repeated content before its main content?",
```
Jym77, line 258: **Note:** I think it is better to make this a `node` or even `node[]` question, asking for all the repeated contents. The reason being that SIA-R97 will have to ask something for each block of repeated content and it might be better to keep the same question.
OTOH, a block of repeated content might be more than one node (subtree) 🤔 which would require a new type of question…

Let's keep it like that for now, and revisit when we write R97.

packages/alfa-rules/src/sia-r101/rule.ts
```diff
@@ -0,0 +1,89 @@
...
+  tags: [Scope.Page, Stability.Experimental],
+  evaluate({ device, document }) {
+    return {
+      applicability() {
+        return withDocumentElement(document);
+      },
+
+      expectations(target) {
+        return {
+          1: Question.of("has-repeated-content-before-main", target)
```
Jym77: **Suggestion:** I think the question would be easier to handle for end users if the subject is the main content rather than the full page (i.e. "Is there repeated content before `<div id="main">`?")

That requires getting the main content first, which will need a new question. I believe this goes in the right direction as it will likely be needed anyway by the other rules (e.g. checking for headings at the start of the main content).

Counter-intuitively, the `main` landmark doesn't need to be unique,  so the question might internally need to be of type `node[]` even if we end up presenting it as a `node` question in the platform (or maybe we can work around that directly in the rule, having a `node` question that is mapped to `node[]` so that `answerIf` can feed it with the list of `main` landmarks?)
Also, even if "the main content" can be several nodes (not just a subtree), I think we are good having "the start of main content" node instead.

packages/alfa-rules/src/sia-r101/rule.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    return !document
+      .descendants(Node.flatTree)
+      .takeUntil((node) => node.equals(main.get()))
+      .some(
+        and(
+          isIncludedInTheAccessibilityTree(device),
+          isContent(Node.flatTree),
+          not(and(Text.isText, (text) => text.data.trim() === "")),
```
Jym77:
```suggestion
          not(and(Text.isText, Text.is(String.isWhitespace)),
```

Jym77 (changes requested):
I think we should ask for the main content explicitly as this answer will be needed for the other rules anyway.

packages/alfa-rules/src/sia-r101/rule.ts
rcj-siteimprove: Having the main content as subject makes sense, but not sure I understand the "start of main content" part at the end. Are you suggesting that we only ask the question "has-repeated-content-before-main" for the first main landmark, or something else? I suppose we could ask it for every main landmark.

Jym77: So, in an ideal world, we could:
1. ask for (unique) main content;
2. ask for any (multiple) repeated contents;
3. Do the check ourselves if any (2) is before (1).

This runs into various problems… 

The first one is that neither "(multiple) repeated content" nor "(unique) main content" is always a single node (subtree), but more realistically a slice of the DOM tree (e.g. two sibling `<div>`, …) This is not really problematic for "(unique) main content" because we do not really care where it stops, only where it starts (to see whether there is repeated content, or a heading, or a skip link, …) So, asking for "start of main content" is effectively as good as asking for "all nodes that are part of main content". For repeated content, this gets a bit more tricky, notably for the "repeated content collapsible" rule, where we need to know the full repeated content, so we might as well ask for the end question directly (in this case ask "is there repeated content before [node identified as start of main]?" rather than "where is (start of?) repeated content?") Not 100% sure that "start of repeated content won't do the job, but let's keep our minds open on that once we have all the rules.

The second problem is that main content is not necessarily unique. The reasonable approach here is that if there is (at least) one "main content" that is good, then the rule should pass (this is what we do for R87 and limits false positives). Locally for this rule, this means that we'll only look at the first main content. But that won't be true for the other rules. And since we want to share the detection of main content (especially share the questions that are expensive to answer), we should already have the support for several "main content".
It is quite likely that several "main contents" will be a rare occurrence and that when the question is answered externally we'll only get one, but at least internally (for the `answerIf` part that will grab all elements with a role of `main`), we need support for several. I *think* it should be possible to have a `node` question that goes through mapping to a `node[]` one that can use `answerIf`, but not 100% sure (especially, not sure we can go all the way to the `answerIf` part). If not, we'll do the wrapping on the interviewer side (Dory).

rcj-siteimprove:
!pr extract

packages/alfa-rules/src/sia-r101/rule.ts
```diff
@@ -0,0 +1,105 @@
...
+                  Question.of("has-repeated-content-before-main", main, target)
+                    .answerIf(
+                      hasNoAccessibleContentBefore(device, document),
+                      false,
+                    )
+                    .map((hasRepeatedContent) =>
+                      expectation(
+                        !hasRepeatedContent,
+                        () => Outcomes.HasNoRepeatedContentBeforeMain,
+                        () => Outcomes.HasRepeatedContentBeforeMain,
```
Jym77:
```suggestion
                        hasRepeatedContent,
                        () => Outcomes.HasRepeatedContentBeforeMain,
                        () => Outcomes.HasNoRepeatedContentBeforeMain,
```
We can flip the logic around, `expectation` is just an `if … then … else` wrapper.

packages/alfa-rules/test/sia-r101/rule.spec.tsx
```diff
@@ -0,0 +1,378 @@
...
+});
+
+test("evaluate() automatically passes when the only content before main is whitespace-only text", async (t) => {
+  const documentWithWhitespaceTextBeforeMain = h.document([
+    <html lang="en">
+      <head>
+        <title>The Three Kingdoms, Chapter 1</title>
+      </head>
+      <body>
+        {h.text(" ")}
```
Jym77, line 42: **Question:** Does that also works if the text is wrapped in an element? (`<span> </span>`)
It feels like it should, but that is also probably a bit too tricky to detect for now (we'll handle that later if needed).
Maybe we should document the behaviour with an example?

Jym77 (dismissed):
Looks pretty good. Some minor improvements suggestions (non blocking).

packages/alfa-rules/test/sia-r101/rule.spec.tsx
rcj-siteimprove: Yes, (`<span> </span>`) works because the span element is not in the accessibility tree. I've added a test for that.

Jym77 (approved):

PR #2129, opened 2026-07-02
A11Y-11673: Ignore non-clickable composite containers in target size

2biazdk:
## Problem

Target size checks could report focusable composite widget containers, such as menus, when their child items were the actual pointer targets. This made exposed non-target regions like menu dividers appear as undersized targets.

## Solution

Exclude container-style composite widget roles from pointer target applicability while keeping their child widgets, such as menu items, covered by the rule.

## Related Jira

https://siteimprove-wgs.atlassian.net/browse/A11Y-11673

### DIFF
--- a/.changeset/target-size-composite-containers.md
+++ b/.changeset/target-size-composite-containers.md
@@ -2,4 +2,4 @@
 "@siteimprove/alfa-rules": patch
 ---
 
-**Fixed:** Target size checks no longer report menu containers as pointer targets when their child menu items are the actual targets.
+**Fixed:** Target size checks (SIA-R111, SIA-R113) no longer report non-clickable composite widget containers as pointer targets when their child widgets are the actual targets.
--- a/.changeset/target-size-composite-containers.md
+++ b/.changeset/target-size-composite-containers.md
@@ -1,0 +1,5 @@
+---
+"@siteimprove/alfa-rules": patch
+---
+
+**Fixed:** Target size checks no longer report composite widget containers, such as menus, as pointer targets when their child widgets are the actual targets.
--- a/.changeset/target-size-composite-containers.md
+++ b/.changeset/target-size-composite-containers.md
@@ -2,4 +2,4 @@
 "@siteimprove/alfa-rules": patch
 ---
 
-**Fixed:** Target size checks no longer report composite widget containers, such as menus, as pointer targets when their child widgets are the actual targets.
+**Fixed:** Target size checks no longer report menu containers as pointer targets when their child menu items are the actual targets.
--- a/packages/alfa-rules/src/common/applicability/targets-of-pointer-events.ts
+++ b/packages/alfa-rules/src/common/applicability/targets-of-pointer-events.ts
@@ -103,12 +103,8 @@ export function isTarget(device: Device): Predicate<Element> {
     isFocusable(device),
     isVisible(device),
     not(isScrolledBehind(device)),
-    hasRole(device, hasTargetRole),
+    hasRole(device, isPointerTargetWidget),
   );
-}
-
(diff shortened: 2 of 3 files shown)


2biazdk:
@Jym77 Let me know what you think about this fix.
… Is it misunderstood?

packages/alfa-rules/src/common/applicability/targets-of-pointer-events.ts
```diff
@@ -103,10 +103,20 @@ export function isTarget(device: Device): Predicate<Element> {
...
 }
 
+function hasTargetRole(role: Role): boolean {
+  return role.isWidget() && !isMenuContainer(role);
+}
+
+function isMenuContainer(role: Role): boolean {
+  // Menus organize child menu items; the child items provide the pointer
+  // targets, not the container itself.
+  return role.hasName("menu") || role.hasName("menubar");
```
Jym77: **Question:** ARIA list [much more "composite" widget roles](https://w3c.github.io/aria/#widget_roles) ([combobox](https://w3c.github.io/aria/#combobox), [grid](https://w3c.github.io/aria/#grid), [listbox](https://w3c.github.io/aria/#listbox), [menu](https://w3c.github.io/aria/#menu), [menubar](https://w3c.github.io/aria/#menubar), [radiogroup](https://w3c.github.io/aria/#radiogroup), [tablist](https://w3c.github.io/aria/#tablist), [tree](https://w3c.github.io/aria/#tree), [treegrid](https://w3c.github.io/aria/#treegrid)). I am not sure how they are actually used (which ones are normally clickable and which aren't). Are `menu` and `menubar` the only ones we want to skip?

packages/alfa-rules/src/common/applicability/targets-of-pointer-events.ts
```diff
@@ -103,10 +103,20 @@ export function isTarget(device: Device): Predicate<Element> {
...
+}
```
Jym77:
```suggestion
const isMenuContainer = Role.hasName("menu", "menubar");
```
should create the correct predicate directly.

packages/alfa-rules/src/common/applicability/targets-of-pointer-events.ts
```diff
@@ -103,10 +103,20 @@ export function isTarget(device: Device): Predicate<Element> {
...
     isVisible(device),
     not(isScrolledBehind(device)),
-    hasRole(device, (role) => role.isWidget()),
+    hasRole(device, hasTargetRole),
   );
 }
 
+function hasTargetRole(role: Role): boolean {
+  return role.isWidget() && !isMenuContainer(role);
+}
```
Jym77:
```suggestion
const isNonContainerWidget = and(role => role.isWidget(), not(isMenuContainer);
```
* Use a more explicit name.
* Predicate combinator is a bit wonky here since the first one is not directly a predicate 🙈, but I think that is still closer to the style we usually have for these.

.changeset/target-size-composite-containers.md
```diff
@@ -0,0 +1,5 @@
+---
+"@siteimprove/alfa-rules": patch
+---
+
+**Fixed:** Target size checks no longer report menu containers as pointer targets when their child menu items are the actual targets.
```
Jym77:
```suggestion
**Fixed:** Target size checks (SIA-R111, SIA-R113) no longer report menu containers as pointer targets when their child menu items are the actual targets.
```

Jym77 (changes requested):
Question about the list of "non-interactive containers", plus some polishing.

Jym77 (approved):

PR #2133, opened 2026-07-06
Typedoc: separate extract and document operations

Jym77:
* Add a "review" theme for Typedoc's markdown extraction, stripping pretty much everything except detailed signature.
  * It includes a patched copy of `typedoc-plugin-markdown`, including the changes proposed in  https://github.com/typedoc2md/typedoc-plugin-markdown/issues/877, https://github.com/typedoc2md/typedoc-plugin-markdown/issues/878, and https://github.com/typedoc2md/typedoc-plugin-markdown/issues/879. This should be cleaned up (maybe re-adapted) when/if the patches are accepted upstream. This does make the PR needlessly big due to all the files in the `vendor` directory…
* Setup "release-ready" documentation generation that uses a given tag to generate the source links; update the documentation and release workflow to create the correct documentation.


Jym77:
!pr extract

PR #2134, opened 2026-07-08
[FEAT-212] Add sign functions.

pooya-si:
## Description 

Add sign functions

## Jira Ticket

[FEAT-212](https://siteimprove-wgs.atlassian.net/browse/FEAT-212)

[FEAT-212]: https://siteimprove-wgs.atlassian.net/browse/FEAT-212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.changeset/wise-signs-abound.md
```diff
@@ -0,0 +1,7 @@
+---
+"@siteimprove/alfa-css": minor
+---
+
+**Added:** The `abs()` and `sign()` CSS math functions are now supported in calculations.
+
+`abs(A)` resolves to the absolute value of its argument, keeping its type. `sign(A)` resolves to `-1`, `0`, or `1` as a `<number>`, regardless of the type of its argument. As with the other math functions, arguments that are percentages or relative lengths are only resolved once a resolver is provided. See [CSS Values and Units 4, § Sign-Related Functions](https://drafts.csswg.org/css-values-4/#sign-funcs).
```
Jym77, line 7: **Note:** So far, we've tend to rarely write detailed changesets, especially when it is mostly "it does what the spec says". OTOH, this probably was a bad habit and given that these details only go in the package,s changelog, this is certainly better with a bit more details.

packages/alfa-css/src/calculation/math-expression/function/abs.ts
```diff
@@ -0,0 +1,101 @@
...
+
+/**
+ * {@link https://drafts.csswg.org/css-values/#sign-funcs}
+ *
+ * @public
+ */
+export class Abs extends Function<"abs", [Expression]> {
+  public static of(expression: Expression): Abs {
+    // `abs()` keeps the type of its argument.
+    return new Abs([expression], expression.kind);
```
Jym77: **Nitpick:** I think I prefer if the constructor does the wrapping into an array (rather than the factory). But I can't really give any rational reason why… So it's just a matter of taste and you can do what feels better for you 😃

packages/alfa-css/src/calculation/math-expression/function/abs.ts
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  protected constructor(args: [Expression], kind: Kind) {
+    super("abs", args, kind);
+  }
+
+  protected get _arg(): Expression {
```
Jym77:
```suggestion
  public get value(): Expression {
```
1. getters are meant for external use, so they should be `public` and without `_` in their name.
2. We very rarely use abbreviated names in Alfa's APIs.
3. I'm not sure what is the best name to use here ("argument", "parameter", "value", …) given it has no obvious semantics (vs, e.g., the `clamp` parameters).  I've picked `value` because it is used in many of our other classes, OTOH usually with a slightly different meaning (the class is some sort of "wrapper" for the value).

packages/alfa-css/src/calculation/math-expression/function/abs.ts
```diff
@@ -0,0 +1,101 @@
...
+
+/** @public */
+export namespace Abs {
+  export function isAbs(value: unknown): value is Abs {
+    return value instanceof Abs;
+  }
+
+  export const parse = (parseSum: CSSParser<Expression>) =>
+    map(
+      CSSFunction.parse("abs", (input) => parseSum(input)),
```
Jym77:
```suggestion
      CSSFunction.parse("abs", parseSum),
```
I think we can eta-compress this call (which avoids creating a new function). This sometimes beak because of the way JS handles `this` binding in anonymous functions, so double check that the tests don't break 😺

packages/alfa-css/src/calculation/math-expression/function/sign.ts
```diff
@@ -0,0 +1,94 @@
...
+/**
+ * {@link https://drafts.csswg.org/css-values/#sign-funcs}
+ *
+ * @public
+ */
+export class Sign extends Function<"sign", [Expression]> {
+  public static of(expression: Expression): Sign {
+    // `sign()` always resolves to a `<number>`, regardless of the type of its
+    // argument.
+    return new Sign([expression], Kind.of());
```
Jym77: **Nitpick:** same as for `Abs.of`, I feel the constructor should do the wrapping in an array, but this is just personal taste.

Jym77: **Suggestion:** Given that the kind is fixed, we don't need to pass it to the constructor, and can let it do the "static" kind building that is needed for the `super` call.

packages/alfa-css/src/calculation/math-expression/function/sign.ts
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  protected constructor(args: [Expression], kind: Kind) {
+    super("sign", args, kind);
+  }
+
+  protected get _arg(): Expression {
```
Jym77:
```suggestion
  protected get value(): Expression {
```
Same comment.

packages/alfa-css/src/calculation/math-expression/function/sign.ts
```diff
@@ -0,0 +1,94 @@
...
+
+/** @public */
+export namespace Sign {
+  export function isSign(value: unknown): value is Sign {
+    return value instanceof Sign;
+  }
+
+  export const parse = (parseSum: CSSParser<Expression>) =>
+    map(
+      CSSFunction.parse("sign", (input) => parseSum(input)),
```
Jym77:
```suggestion
      CSSFunction.parse("sign", parseSum),
```

packages/alfa-css/test/calculation/function/abs.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+  ] as const) {
+    const calculation = parse(`abs(${arg})`);
+
+    t(calculation.isNumber());
+
+    t.deepEqual(calculation.toJSON(), {
+      type: "math expression",
+      expression: {
+        type: "value",
+        value: { type: "number", value: value },
```
Jym77:
```suggestion
        value: { type: "number", value },
```
When the variable has the same name as the field, there is no need to repeat it.

packages/alfa-css/test/calculation/function/abs.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+  ] as const) {
+    const actual = serialize(`abs(${arg})`);
+
+    t.deepEqual(actual, {
+      type: "math expression",
+      expression: {
+        type: "value",
+        value: {
+          type: type,
+          value: value,
```
Jym77:
```suggestion
          type,
          value,
```
(which may be enough to convince prettier to put the full object in one line…)

packages/alfa-css/test/calculation/function/abs.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+  t.deepEqual(reduced.toJSON(), {
+    type: "math expression",
+    expression: {
+      type: "value",
+      value: { value: 14, type: "length", unit: "px" },
+    },
+  });
+});
+
+test(".reduce() re-derives the kept wrapper's kind from the reduced argument", (t) => {
```
Jym77, line 150: **Praise:** Cool test 😎

packages/alfa-css/test/calculation/function/sign.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+  ] as const) {
+    const calculation = parse(`sign(${arg})`);
+
+    t(calculation.isNumber());
+
+    t.deepEqual(calculation.toJSON(), {
+      type: "math expression",
+      expression: {
+        type: "value",
+        value: { type: "number", value: value },
```
Jym77:
```suggestion
        value: { type: "number", value },
```

packages/alfa-css/test/calculation/function/sign.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    ["-90deg", -1],
+    ["0deg", 0],
+  ] as const) {
+    const actual = serialize(`sign(${arg})`);
+
+    t.deepEqual(actual, {
+      type: "math expression",
+      expression: {
+        type: "value",
+        value: { type: "number", value: value },
```
Jym77:
```suggestion
        value: { type: "number", value },
```

Jym77 (changes requested):
Great work. Several minor comments/polishing.

Jym77 (approved):

PR #2135, opened 2026-07-10
[FEAT-239] [FEAT-227] [FEAT-234] Support `lh`, `rlh` and `rch` units.

pooya-si:
## Description 

Support `lh`, `rlh` and `rch` units.

## Jira Tickets

[FEAT-239](https://siteimprove-wgs.atlassian.net/browse/FEAT-239)
[FEAT-227](https://siteimprove-wgs.atlassian.net/browse/FEAT-227)
[FEAT-234](https://siteimprove-wgs.atlassian.net/browse/FEAT-234)

[FEAT-239]: https://siteimprove-wgs.atlassian.net/browse/FEAT-239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FEAT-227]: https://siteimprove-wgs.atlassian.net/browse/FEAT-227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[FEAT-234]: https://siteimprove-wgs.atlassian.net/browse/FEAT-234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Jym77 (changes requested):
Ah, spotting a problem I haven't foreseen 😅  and some TechDebt we should fix instead of turning it into a broken window… Sorry, this turns into a "going back to the old drawing board" thing…
* The problem here is that `line-height: 2lh` is a valid CSS declaration where `lh` should here be resolved against the **parent** element (to avoid circular reference). This is specifically called out in the text after the [rlh definition](https://drafts.csswg.org/css-values/#rlh). If I understand properly, this should also solve the circularity problem and the need for thunks.
* The TechDebt is the spreading of the 1.2 "magical" value. `line-height` was a property we added very early, and since then we've added a model of ["used" value](https://www.w3.org/TR/css-cascade/#used) (which, in our case, is still mixed with [actual values](https://www.w3.org/TR/css-cascade/#actual), but that's another topic). We should rewrite the `line-height` support to have that and save ourselves the spreading of a magic constant in many places in the code.

---
Line-height used value:
* [ ] For an example of a property with a simple-ish used value, see [outline-color](https://github.com/Siteimprove/alfa/blob/4b6edc148fe257569aeabe6d4c72f2a5d8da43a8/packages/alfa-style/src/property/outline-color.ts#L26), or search for `use:` in `alfa-style` package (many are returning options, but `line-height` shouldn't need to).
* [ ] The used value for `line-height` should resolve both the numbers and the `normal` keyword, essentially doing the `Selective` part you do in the resolver here. This follows the text for the `normal` and `<number>` cases in [the definition of line-height](https://drafts.csswg.org/css2/#propdef-line-height).
* [ ] Search for other places where `line-height` is used and whether we have some handling of `normal` here, replace with the used value (i.e. `style.computed("line-height")` should become `style.used`). One such place is [is-wide-enough](https://github.com/Siteimprove/alfa/blob/4b6edc148fe257569aeabe6d4c72f2a5d8da43a8/packages/alfa-rules/src/common/expectation/is-wide-enough.ts) (where the comment on line 46 incorrectly refers to `font-size` when it should be `line-height` 🙈) which is a bit complex due to possibly handling 2 other properties, but I think that replacing the `style.computed` call at line 18 should do the trick (and clean a bit the `switch` which hopefully will be caught by type checking). 

---
For picking the parent style.
* [ ] We already do that in the `font-*` properties (as per the same text in CSS specs), e.g. [font-size](https://github.com/Siteimprove/alfa/blob/main/packages/alfa-style/src/property/font-size.ts#L106-L111), by calling the resolver on the parent style (`style.parent` is effectively the style of the parent element as there is an underlying styles tree mimicking the DOM tree). (that property is complex, but the key part is here)
* [ ] This will get a bit more tricky for `line-height` as we only need to call the parent style for the `(r)lh` properties. I _think_ we will have to add one new parameter to `Resolver.Length` (and two to `Resolver.LengthPercentage`) to optionally pass styles for `(r)lh` units (defaulting to the same as the existing ones). I'm not a huge fan of this solution (in terms of code complexity for a special case, which is going to be used only once currently), but I do not see an cleaner way for now.
* [ ] I think this will also allow to remove the thunks and pass values directly. The circularity should be broken by going to the parent at each call. There is still a word of caution along the root (where the empty style has itself as a parent), but because it uses the default UA stylesheet that does not have `lh`, I think it will work out (the same way it works out for the `font-size` property where we do not seem to have special anti-circularity handling 🤔).
* [ ] We definitely need test of `line-height` being expressed in `(r)lh` to make sure it effectively works 😄

packages/alfa-css/src/value/numeric/length.ts
```diff
@@ -194,13 +194,22 @@ export namespace Length {
...
   /**
-   * Build a (fixed) length resolver, using basis for the relative units
+   * Build a (fixed) length resolver, using basis for the relative units.
+   *
+   * @remarks
+   * `lhBase` (and `rlhBase`) depend on the computed `line-height`, which
+   * itself may contain `lh` (resp. `rlh`) units; callers resolve these
+   * against the parent's (resp. initial) line-height, avoiding a circular
+   * dependency. They default to `1.2 * emBase` (resp. `1.2 * remBase`),
+   * assuming a `normal` line-height.
```
Jym77:
```suggestion
```
This remark is no longer necessary here since the latest version of the resolver doesn't care where `(r)lhbase` come from.

packages/alfa-css/src/value/numeric/length.ts
```diff
@@ -194,13 +194,22 @@ export namespace Length {
...
    */
   export function resolver(
     emBase: Canonical,
     remBase: Canonical,
     vwBase: Canonical,
     vhBase: Canonical,
+    lhBase: Canonical,
+    rlhBase: Canonical,
```
Jym77, line 205: **Note** (for futures selves, non-blocking): This starts being a lot of parameters. It's still OK for now, but if we need to add more (to support other units), it might be better to change that to a single object (of type `{ emBase: Canonical, … }`) since that will effectively name the parameters at call site.

.changeset/tall-lines-resolve.md
```diff
@@ -0,0 +1,6 @@
+---
```
Jym77, line 1: **Chore:** We need another (breaking) changeset for the API change of `resolver` in `alfa-css`. (since we're still in major 0, breaking changes still only get minor bumps).

packages/alfa-css/test/value/collection/list.spec.ts
Jym77, line 45: **Note** (for future selves, non-blocking): at some point, we may want to have a single shared resolver for all those tests, or a resolver factory with optional overload of the default values when the test calls it. The fact that we have to update it in so many tests is a strong sign for that…

packages/alfa-style/src/resolver.ts
```diff
@@ -35,6 +35,7 @@ export namespace Resolver {
    */
   function lengthResolver(
     style: Style,
+    options: Options = {},
```
Jym77: **Issue:** Let's align with the way we handle font size, by passing a style for `line-height` computation too.

```suggestion
    lineHeightStyle: Style = style
```
We can then move here the computation that are currently in `line-height.ts`, and only send the parent style from there. Other calls won't need to change as `font-size` also resolves `(r)lh` against its parent style.

---
Note that the `style` is currently used for two purposes: finding the device (resolving `vh` and `vw` units), and resolving the font size. This is OK as the device should be the same for an element and its parent. OTOH, it does make some assumptions that maybe we want to break down… Maybe we should go all way to have a `style`, `fontSizeStyle`, and `lineHeightStyle` parameters, with each defaulting to the previous, and only `font-size` and `line-height` overwriting them.
Another possibility could be to have a single style, and flags for "use parent for font size / line height", which would at least internalize the coupling.

I'm not fully sure these two refactors would make things easier to understand and maintain, though… So maybe just a possibility to keep in mind.

packages/alfa-style/src/resolver.ts
```diff
@@ -43,7 +44,29 @@ export namespace Resolver {
     const fontSize = style.computed("font-size").value;
     const rootFontSize = style.root().computed("font-size").value;
```
Jym77, line 50:
```suggestion
const fontSize = style.computed("font-size").value;
const rootFontSize = style.root().computed("font-size").value;
const lineHeight = lineHeightStyle.used("line-height").value;
const rootLineHeight = lineHeightStyle.root().used("line-height").value;
```

packages/alfa-style/src/resolver.ts
```diff
@@ -43,7 +44,29 @@ export namespace Resolver {
...
+      width,
+      height,
+      lineHeightBase,
+      rootLineHeightBase,
+    );
+  }
+
+  export interface Options {
+    lineHeightBase?: Length.Canonical;
+    rootLineHeightBase?: Length.Canonical;
```
Jym77:
```suggestion
    return Length.resolver(
      fontSize,
      rootFontSize,
      width,
      height,
      lineHeight,
      rootLineHeight
    );
  }
```

packages/alfa-style/src/resolver.ts
```diff
@@ -53,8 +76,9 @@ export namespace Resolver {
   export function lengthPercentage(
     base: Length.Canonical,
     style: Style,
+    options?: Options,
```
Jym77: **Issue:** Same change here.
```suggestion
    lineHeightStyle: Style = style,
```

packages/alfa-style/src/resolver.ts
```diff
@@ -53,8 +76,9 @@ export namespace Resolver {
...
   ): LengthPercentage.Resolver {
-    return { percentageBase: base, length: lengthResolver(style) };
+    return { percentageBase: base, length: lengthResolver(style, options) };
```
Jym77:
```suggestion
    return { percentageBase: base, length: lengthResolver(style, lineHeightStyle) };
```

packages/alfa-style/test/property/line-height.spec.tsx
```diff
@@ -36,6 +36,104 @@ test("#computed() resolves `line-height: calc(1 + 2)`", (t) => {
...
+  t.deepEqual(style.computed("line-height").toJSON(), {
+    value: {
+      type: "length",
+      value: 32,
+      unit: "px",
+    },
+    source: h.declaration("line-height", "2em").toJSON(),
+  });
+});
+
```
Jym77:
```suggestion
```
This test seems to be included in the next one (expect for the result of tre computation, which should also be part of the next one).

packages/alfa-style/test/property/line-height.spec.tsx
```diff
@@ -36,6 +36,104 @@ test("#computed() resolves `line-height: calc(1 + 2)`", (t) => {
...
+  for (const value of [
+    "normal",
+    "1.5",
+    "24px",
+    "150%",
+    "2em",
+    "calc(1em + 2px)",
+    "2lh",
+    "1rlh",
+  ]) {
```
Jym77: **Chore:** Add the expected used value, so we can check the computation.

```suggestion
  for (const [value, used] of [
    ["normal", 19.2],
    ["1.5", 24],
    ["24px", 24],
    ["150%", 24],
    ["2em", 32],
    ["calc(1em + 2px)", 18],
    ["2lh", 38.4],
    ["1rlh", 19.2],
  ]) {
```

packages/alfa-style/test/property/line-height.spec.tsx
```diff
@@ -36,6 +36,104 @@ test("#computed() resolves `line-height: calc(1 + 2)`", (t) => {
...
+    const element = <div style={{ lineHeight: value, fontSize: "16px" }} />;
```
Jym77, line 50: **Note:** we may want to inject a font size different from the default, to make sure it has been taken into account properly. OTOH, there might already be a test for that in this file 🤔

packages/alfa-style/test/property/line-height.spec.tsx
```diff
@@ -36,6 +36,104 @@ test("#computed() resolves `line-height: calc(1 + 2)`", (t) => {
...
+
+    const style = Style.from(element, device);
+
+    t(style.computed("line-height").value !== undefined);
```
Jym77:
```suggestion
    t.deepEqual(style.used("line-height").value.toJSON(), {
      type: "length",
      value: used,
      unit: "px",
    });
```

packages/alfa-style/test/property/line-height.spec.tsx
```diff
@@ -54,3 +152,73 @@ test("#computed() resolves percentages from the element's `font-size`", (t) => {
...
+    <div style={{ lineHeight: "calc(1lh + 50%)", fontSize: "10px" }}>
+      Hello
+    </div>
+  );
+
+  <div style={{ lineHeight: "20px" }}>{target}</div>;
+
+  const style = Style.from(target, device);
+
+  // 1lh = 20px (parent's line-height) + 50% of 10px (own font-size) = 25px.
```
Jym77, line 159: **Praise:** Awesome test and explanation 👍

packages/alfa-style/test/property/line-height.spec.tsx
```diff
@@ -54,3 +152,73 @@ test("#computed() resolves percentages from the element's `font-size`", (t) => {
...
+    },
+    source: h.declaration("line-height", "calc(1rlh + 50%)").toJSON(),
+  });
+});
+
+test("#computed() resolves chained `lh` and `rlh` line-heights down a tree", (t) => {
+  const target = <div style={{ lineHeight: "calc(0.5lh + 0.5rlh)" }} />;
+  const b = <div style={{ lineHeight: "calc(1lh + 1rlh)" }}>{target}</div>;
+  const a = <div style={{ lineHeight: "2lh" }}>{b}</div>;
+  const root = <div style={{ lineHeight: "20px" }}>{a}</div>;
```
Jym77, line 199: **Praise:** Yes, this one is greatly needed too 🎉

Jym77 (changes requested):
Good work! We still need to simplify a bit the `alfa-style` resolver by passing a `lineHeight` style and delegates the handling there (the same way it is done for the `style` being used to find the font size).

Jym77 (approved):

PR #2145, opened 2026-07-16
Fix role computation of elements based on names

### LINKED ISSUE #298: Elements whose implicit role depends on the presence of an accessible name are not correctly handled

### LINKED ISSUE #1757: `<img>` with empty `alt` but a name should have a role of `img`
As per recent change in HTML AAM / ARIA in HTML, images with an empty `alt` but another naming mechanism get a role of `img`, not `presentation`.

* This works both for `aria-label(ledby)` and `title`, the other ways of naming images.
* TODO: check when these are empty whether UA still default to them.

Jym77:
Resolves #298 
Resolves #1757 

* `<aside>` and `<section>` without a name are correctly getting the generic role as default.
* `` with a name are correctly getting the image role as default.

Browsers do currently differ on the corner cases (empty or whitespace `aria-label` or `title`, `aria-labelledby` referring to invalid or empty elements). We mostly follow what Chrome does as it is the most frequently used browser. Hopefully these behaviours will be clarified in the specifications or the WPT and browsers will align; at which point we should follow suite.

`<form>` without a name are special. They still need to be treated as forms by Assistive Technologies (for interaction purpose, …) but are not treated as landmarks. The old `hasIncorrectRoleWithoutName` predicate has been renamed `isNotLandmarkWithoutName` and only handles the `<form>` case now.


Jym77:
!pr extract

rcj-siteimprove (approved):
Looks good 😃

PR #2146, opened 2026-07-17
Fix clipped by masking frequent shapes

### LINKED ISSUE #1548: SIA-R83 (Text is clipped) doesn't detect invisibility due to clip-path
The CSS properties clip-path and/or -webkit-clip-path seems to be cause a false positive for this rule.   With those CSS properties one clip it completely so that it's invisible, even while its bounding rect's width/height are much greater than zero.  If the text isn't visible then it doesn't meet the applicability for the rule so it shouldn't be checked.  But it is checked, and it fails. 

The HTML that I saw on a customer's page which exhibited this looked roughly like this:
```
<div style="clip-path: polygon(0px 0px,0px 0px,0px 0px); -webkit-clip-path: polygon(0px 0px,0px 0px,0px 0px); overflow: hidden; height: 10px; width: 10px;">lorem ipsum dolor</div>
```
But that HTML on its own doesn't reproduce the problem, so there's something I don't understand here.   @Jym77 I won't post the customer's page here but if you would like me to file a jira too or do anything else to be more specific on this, let me know.

Jym77:
Resolves #1548 

Elements that are clipped (via `clip-path`) by an "obviously empty" shape are invisible.

This also de-deprecate the `Rectangle` shape for whitespace separated syntax, as it is part of the [`<basic-shape>` CSS type](https://drafts.csswg.org/css-shapes-1/#funcdef-basic-shape-rect); the comma-separated syntax, only used in the [deprecated `clip` property](https://drafts.csswg.org/css-masking/#clip-property) is still present via a deprecated parser.

Thanks to @wittjeff for investigating commonly used empty shape.


rcj-siteimprove (approved):
Looks good 😃

PR #2149, opened 2026-07-21
[A11Y-11106] Implement SIA-R118.

pooya-si:
## Description 

Implement SIA-R118.

## Jira Ticket 

A11Y-11106

## Related Documentation  

https://www.w3.org/TR/WCAG2/#images-of-text
https://github.com/Siteimprove/sanshikan/blob/main/rules/sia-r118.md


packages/alfa-rules/src/common/act/question.ts
```diff
@@ -278,5 +278,29 @@ export namespace Question {
       message:
         "Does the accessible name of this element present the same information as the image?",
     },
+    // R118 (experimental)
+    "is-image-text-human-language": {
```
Jym77:
```suggestion
    "does-image-contain-text": {
```
While the fact that it is "human language" is indeed important, the focus of the question is about containing text, so the URI should reflect that.

packages/alfa-rules/src/common/act/question.ts
```diff
@@ -278,5 +278,29 @@ export namespace Question {
...
+      type: "boolean",
+      message:
+        "Does the image contain visible text that expresses something in a human language?",
```
Jym77, line 285:
```suggestion
        "Does this image contain visible text that expresses something in a human language?",
```
The question will be linked to a specific image (the question's subject), so we can use demonstrative pronoun (even if these questions' text are not really used anywhere, still better to keep them nice…)

packages/alfa-rules/src/common/act/question.ts
```diff
@@ -278,5 +278,29 @@ export namespace Question {
...
+    },
+    "is-image-text-decorative": {
+      type: "boolean",
+      message: "Is the image with text purely decorative?",
```
Jym77, line 289:
```suggestion
      message: "Is this image with text purely decorative?",
```

packages/alfa-rules/src/common/act/question.ts
```diff
@@ -278,5 +278,29 @@ export namespace Question {
...
+    },
+    "is-image-text-incidental": {
+      type: "boolean",
+      message: "Is the text an insignificant part of the image?",
```
Jym77, line 293:
```suggestion
      message: "Is the text an insignificant part of this image?",
```

packages/alfa-rules/src/sia-r118/rule.ts
```diff
@@ -0,0 +1,128 @@
...
+            expectation(
+              redundant,
+              () => Outcomes.IsRedundant,
+              () => Outcomes.IsImageOfText,
+            ),
+          );
+
+        const isEssential = () =>
+          Question.of("is-image-text-essential", target).map((essential) =>
+            expectation<Question.Metadata, Element, Element, -1>(
```
Jym77, line 55: **Question:** Does the build break without specifying the depths (`-1`, …)?

If possible, having as much of the depths automatically inferred is much better as it is more stable when we need to make changes on the rules. Given the structure of this questions chain, I suspect we only need to specify one depth, maybe even none of them. It is definitely worth trying to remove them one by one until something breaks 😃

packages/alfa-rules/src/sia-r118/rule.ts
```diff
@@ -0,0 +1,128 @@
...
+              () => Outcomes.IsDecorative,
+              isIncidental,
+            ),
+          );
+
+        return {
+          1: Question.of("is-image-text-human-language", target).map(
+            (humanLanguage) =>
+              expectation<Question.Metadata, Element, Element, 2>(
+                humanLanguage,
```
Jym77, line 84:
```suggestion
            (hasText) =>
              expectation<Question.Metadata, Element, Element, 2>(
                hasText,
```
Here also, the important bit is "having text", more than the "expressing something in human language" part.

packages/alfa-rules/src/common/act/question.ts
```diff
@@ -278,5 +278,29 @@ export namespace Question {
...
+    },
+    "is-image-text-essential": {
+      type: "boolean",
+      message:
+        "Is the presentation of the text essential (for example, a logo)?",
```
Jym77:
```suggestion
        "Is the presentation of the text in this image essential (for example, a logo)?",
```

Jym77 (changes requested):
Looks great. 

There is a discussion around how we will concretely answer the questions (connection with an AI, …) and how we can reduce the number of requests/tokens used. This may be ignored for now, knowing we can change the rule later on. Two points to consider:
* In the current implementation, the questions are asked sequentially. This means that Dory does not know what is the next question until we get the answer to the current one, and thus each question has to be asked separately. We may still have some special code in Nemo (who is dispatching the questions to further systems) to detect that "does-image-contain-text" should actually ask about all the exceptions in one go and then unpack the answer and store it as several answers. This, however, forces a bit of the logic to be duplicated inside Nemo. Another possibility could be to ask all questions together (which may need a new return type of several boolean, or encoding them in an integer, …); so that if we later decide to add or remove one of them, we do not need any change in Nemo. (or we build a system in the Question class to group several questions into a single one, similar to the classic `Promise.all` wrapper, this may be the most future-proof solution, but also the most complex one).
* The "logo" exception is a case for which we can certainly make a good heuristic (if the image's name is `logo` or `*_logo_*`, or a few other cases) (this would fit the (yet unbuilt) "tier 1" [way to answer questions](https://siteimprove-wgs.atlassian.net/wiki/spaces/A11Y/pages/7551680703/Ways+to+answer+questions+asked+by+Alfa.), which I am convinced we'll need to build just to save AI tokens by not asking repeatedly whether `logo.jpg` has essential text presentation… In the current "sequential" setup, that question is asked second. We could build the heuristic to answer the "does-image-contain-text" question instead, but that is definitely a mismatch (or the heuristic also say "yes" to that question if it detects a logo, which is also a bit confusing). No very neat way here, maybe the "group all questions" approach is the good one.

Anyway, these are probably more things to keep in mind for later and to revisit when we add the rule to the platform.

Jym77 (approved):

rcj-siteimprove (dismissed):

PR #2152, opened 2026-07-22
Refactor `alfa-act` rule types

rcj-siteimprove:


rcj-siteimprove:
!pr extract

packages/alfa-act/src/rule/atomic.ts
```diff
@@ -0,0 +1,182 @@
...
+    tags: Array<Tag>,
+    evaluate: Atomic.Evaluate<I, T, Q, S>,
+  ) {
+    super(uri, requirements, tags, (input, oracle, outcomes, performance) =>
+      evaluateRule(
+        this,
+        oracle,
+        outcomes,
+        performance,
+        async (instrument) => {
```
Jym77, line 61:
```suggestion
        // For atomic rules, we collect the final outcomes by first running the Applicability,
        // and then running the expectations on each target.
        async (instrument) => {
```
I was a bit confused on the flow going on here between the files. It is the correct flow, though 😄

packages/alfa-act/src/rule/composite.ts
```diff
@@ -0,0 +1,187 @@
...
+    composes: Array<Rule<I, T, Q, S>>,
+    evaluate: Composite.Evaluate<I, T, Q, S>,
+  ) {
+    super(uri, requirements, tags, (input, oracle, outcomes, performance) =>
+      evaluateRule(
+        this,
+        oracle,
+        outcomes,
+        performance,
+        async (instrument) => {
```
Jym77, line 70:
```suggestion
        // For composite rules, we collect the final outcomes by first evalutaing every
        // atomic rule in it, and then running the composition logic.
        async (instrument) => {
```

rcj-siteimprove:
!pr extract

packages/alfa-act/src/rule/evaluation.ts
```diff
@@ -0,0 +1,198 @@
...
+              item.target,
+              item.expectations,
+              rule,
+              oracle,
+              item.oracleUsed,
+            ),
+          ),
+        );
+
+      return resolvePhase ? instrument.phase(resolvePhase, run) : run();
```
Jym77: **Suggestion:** Since `Instrument#phase` already has logic for "just call `run`" (when there is no performance), I think we should also delegate there the test for "no phase name", this will make this call here easier: we always just call `instrument.phase`.

Jym77 (dismissed):
Yes, this is the split and abstractions we need to simplify that.
I'm not really convinced by the name "instrument" nor the associated methods, but that is a detail and can be changed later. I'm also not sure whether the Resolvable/Collected abstractions need to be lifted further (to the usual class+namespace rather than just an interface), but that can also wait.

Definitely a big step in the good direction.

Jym77 (dismissed):

rcj-siteimprove:
!pr extract

Jym77 (approved):

PR #2168, opened 2026-09-02
Intern Context.hover/active/focus/visit per element

lukasbob:
## Summary

`Context.focus(element)` (and the other single-state static factories — `.hover()`, `.active()`, `.visit()`) constructed a brand-new `Context` instance on every call, even for the same element. `alfa-cascade`/`alfa-style`'s internal caches are `WeakMap`-keyed on `Context` **object identity**, and those caches are scoped to the `Document`/device (they persist for the life of the page object, not just one rule's evaluation) — so repeated queries for the same element under the same single-state context missed the cache every time, even across two full evaluations of the same already-loaded page.

This interns the four factories (`Cache<Element, Context>` per state, returning the same instance for the same element). Pure caching fix — no change to which selectors match or what `Context.getState`/`.hasState`/etc. return for any given element.

## Correction: this has ~no measurable production impact

The PR originally claimed a large real-world performance win based on `alfa-benchmark`'s comparison numbers (`sia-r65` -99.67% etc.). **That framing was wrong.** Traced the actual mechanism with instrumented call counters on `Context.focus()` and `Cascade.get()`'s non-empty-context path:

- `sia-r65` run **once**, cold, against a fresh page: 100% of `Cascade.get()` calls miss (50,278/50,278) — **identical with or without this fix**, ~135-140s either way.
- `sia-r65` run a **second** time against the *same already-loaded page* (which is exactly what `alfa-benchmark`'s `run.ts` does: one discarded warmup pass, then the timed pass the benchmark numbers come from): with the fix, the second pass collapses to ~0.5s. Without the fix, the second pass is just as slow as the first.

The fix only pays off when the *same* `Document`/device is evaluated more than once in the same process — which is exactly the benchmark harness's warmup+timed methodology, but is **not** how production auditing works: every rule is evaluated exactly once per page, so there's no second pass to benefit from. This fix is confirmed to have effectively zero real-world performance impact, despite the dramatic benchmark numbers.

It's kept anyway as a correctness-neutral cache-identity fix (harmless, verified safe, technically sound), not as a performance fix. The original `sia-r65`/`sia-r62` production performance problem that motivated this investigation remains unsolved.

## Test plan

- [x] `yarn build packages/alfa-selector` (+ downstream: `alfa-cascade`, `alfa-style`, `alfa-rules`, `alfa-act`)
- [x] `yarn test packages/alfa-selector packages/alfa-cascade packages/alfa-style packages/alfa-rules packages/alfa-act` — 2146/2146 passing
- [x] Manual before/after outcome-count check on `sia-r65` against two local fixtures — identical pass/fail/cantTell counts with and without the fix
- [x] Instrumented call-count check isolating a single fresh evaluation from a repeat evaluation of the same page (see correction above)
- [x] `yarn changeset status` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)


pooya-si:
@lukasbob thanks a lot for this work, Mads! 
I have a question around `Measured impact` and the `After` improvements. How was this measured? Was the second run performed after the first run? Or were they performed independently? I am asking because for instance in `packages/alfa-rules/src/sia-r65/rule.ts`, I see `withFocus` object is passed down to all other functions and it's the only reference to the `Context` so I don't fully understand how it makes such big difference.
Using the same context across several different rules definitely has bigger impact but on a single rule I am not sure how - I'd appreciate more details :)

lukasbob:
> @lukasbob thanks a lot for this work, Mads! I have a question around `Measured impact` and the `After` improvements. How was this measured? Was the second run performed after the first run? Or were they performed independently? I am asking because for instance in `packages/alfa-rules/src/sia-r65/rule.ts`, I see `withFocus` object is passed down to all other functions and it's the only reference to the `Context` so I don't fully understand how it makes such big difference. Using the same context across several different rules definitely has bigger impact but on a single rule I am not sure how - I'd appreciate more details :)

You are very right to question this. I made a mistake here, by running two passes over the rules, warming the cache in the first run. This is not how the checks are run in our pipeline, so the real impact is likely negligible. Apologies!

pooya-si:
> > @lukasbob thanks a lot for this work, Mads! I have a question around `Measured impact` and the `After` improvements. How was this measured? Was the second run performed after the first run? Or were they performed independently? I am asking because for instance in `packages/alfa-rules/src/sia-r65/rule.ts`, I see `withFocus` object is passed down to all other functions and it's the only reference to the `Context` so I don't fully understand how it makes such big difference. Using the same context across several different rules definitely has bigger impact but on a single rule I am not sure how - I'd appreciate more details :)
> 
> You are very right to question this. I made a mistake here, by running two passes over the rules, warming the cache in the first run. This is not how the checks are run in our pipeline, so the real impact is likely negligible. Apologies!

Thanks a lot for the details, and please don't apologize for this great work. I learned about the performance bottlenecks here and I really appreciate it :-) 
I am trying find a way to test this optimization

